PR #2, opened 2016-12-08
Specialization

szeiger:
An attempt at adding specialized collections to the design. My goal was to have some collections be specialized while others are not, without impacting the API design too much (like having to pass a `ClassTag` to every combinator or requiring concrete types in order not to lose specialization).

/cc @odersky @Ichoran

### DIFF
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -2,8 +2,9 @@ package strawman.collection
 
 import scala.annotation.unchecked.uncheckedVariance
 import scala.reflect.ClassTag
-import scala.{Int, Boolean, Array, Any, Unit, StringContext}
+import scala.{Int, Long, Double, Boolean, Array, Any, Unit, StringContext}
 import java.lang.String
+import java.util.PrimitiveIterator
 
 import strawman.collection.mutable.{ArrayBuffer, StringBuilder}
 
@@ -11,6 +12,35 @@ import strawman.collection.mutable.{ArrayBuffer, StringBuilder}
 trait Iterable[+A] extends IterableOnce[A] with IterableLike[A, Iterable] {
   /** The collection itself */
   protected def coll: this.type = this
+
+  /** The ClassTag of the elements of this collection. Can be more or less precise than A but not inompatible with it.
+    * For example, an Iterable[Int] can have a ClassTag[Any] if a more precise type was not known at construction
+    * time, and an Iterable[Any] can have a ClassTag[Int] (but only if all elements are really of type Int). */
+  def elementClassTag: ClassTag[_ >: (A @uncheckedVariance)]
+
+  /** Get a specialized iterator. The default implementation simply boxes and unboxes as necessary. It should
+    * be overridden in subclasses with primitive data representations. */
+  def specializedIterator(implicit spec: Specialized[A @uncheckedVariance]): spec.Iterator = {
+    val it = iterator
+    spec.asInstanceOf[Specialized[_]] match {
+      case Specialized.SpecializedInt =>
+        new PrimitiveIterator.OfInt {
+          override def nextInt(): Int = it.next().asInstanceOf[Int]
+          override def hasNext: Boolean = it.hasNext
+        }
+      case Specialized.SpecializedLong =>
+        new PrimitiveIterator.OfLong {
+          override def nextLong(): Long = it.next().asInstanceOf[Long]
+          override def hasNext: Boolean = it.hasNext
+        }
+      case Specialized.SpecializedDouble =>
+        new PrimitiveIterator.OfDouble {
+          override def nextDouble(): Double = it.next().asInstanceOf[Double]
+          override def hasNext: Boolean = it.hasNext
+        }
+      case _ => it
+    }
+  }.asInstanceOf[spec.Iterator]
 }
 
 /** Base trait for Iterable operations
--- a/src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
+++ b/src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
@@ -1,35 +1,40 @@
 package strawman.collection.mutable
 
-import scala.{Array, Int, Boolean, Unit, AnyRef}
-import scala.Predef.intWrapper
-import strawman.collection.{IndexedView, Iterable, IterableFactory, IterableOnce, Seq, SeqLike}
+import scala.{AnyRef, Array, Boolean, Double, Int, Long, Unit, specialized}
+import scala.Predef.{classOf, intWrapper}
+import scala.annotation.unchecked.uncheckedVariance
+import scala.reflect.ClassTag
+import strawman.collection.{ArrayView, IndexedView, Iterable, IterableFactory, IterableOnce, Seq, SeqLike, Specialized}
 
 /** Concrete collection type: ArrayBuffer */
-class ArrayBuffer[A] private (initElems: Array[AnyRef], initLength: Int)
+class ArrayBuffer[@specialized(Int, Long, Double) A] private[collection] (initElems: Array[A], initLength: Int)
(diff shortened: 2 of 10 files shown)


julienrf:
This looks good to me.

szeiger:
Rebased on master. Does anyone else have comments on this?

Ichoran:
To be honest I'm not thrilled with it.  It looks like it will work and is pretty compact, which are two big pluses.  But it has potential for runtime overhead, doesn't look that easy to extend, and doesn't have a clear path for getting efficient higher-order operations.  I guess it comes down to what you were planning on having `Iterator` do.  Java 8's approach with PrimitiveIterator is clunky as anything.

I guess the fundamental issue that I don't quite see the answer to from this PR is whether specialized collections are fundamentally different types that are wrapped to look like a non-specialized type, or whether they use Scala's specialization machinery.  Right now the trickiest stuff where that matters (map, collect, flatMap, etc.) isn't there.  That's really where the specialization budget gets blown, and is why we can't just drop @specialized on the whole hierarchy.

odersky:
Maybe @DarkDimius could comment on automatic specialization prospects?

szeiger:
It's a hybrid approach that uses `@specialized` for collection types that are really specialized without requiring it everywhere. The fundamental stream processing methods like `map` and `flatMap` require specialization-aware implementations based on `elementClassTag` and recognizing specialized function types. I'll add a commit that imlements them.

The machinery for that in `SpecializationUtil` is much uglier than I would like and could be inefficient for smaller collections. AFAIK there is no fast and easy runtime reflection for Scala's specialization, which would be needed to improve `map`.

Just adding `@specialized` everywhere would be both too little and too much. It duplicates code where it is unnecessary but does not help with operations like `map`.

I'm currently using Java's specialized Iterators as a placeholder for a better solution. Initially I didn't want to touch our `Iterator` trait but it may be best to use Scala specialization for it.

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -2,15 +2,45 @@ package strawman.collection
...
+  def elementClassTag: ClassTag[_ >: (A @uncheckedVariance)]
+
+  /** Get a specialized iterator. The default implementation simply boxes and unboxes as necessary. It should
+    * be overridden in subclasses with primitive data representations. */
+  def specializedIterator(implicit spec: Specialized[A @uncheckedVariance]): spec.Iterator = {
+    val it = iterator
+    spec.asInstanceOf[Specialized[_]] match {
+      case Specialized.SpecializedInt =>
+        new PrimitiveIterator.OfInt {
+          override def nextInt(): Int = it.next().asInstanceOf[Int]
```
DarkDimius: I don't think this is worth it, if it was boxed it's better to keep it boxed as otherwise you most likely will introduce slowdown by reboxing.

szeiger: It has to unbox here because it returns an Int iterator which operates on primitive values. If the consumer of the iterator expects to box the results, it should use `iterator` instead of `specializedIterator`.

DarkDimius:
@szeiger, the more I work on specialization, the more I believe that it shouldn't be user-controlled. 
The reason is that a single re-boxing eliminates most of the benefits of specialization, while keeping the disadvantages(code explosion). The second reboxing will introduce a performance regression.
This is very visible on operations that are complex and have multiple opportunities to re-box, such as `flatMap` on normal collections or `partition` on iterators.

At the same time, there's no valid `fits-them-all` way to specify what specializations to use, someone may have their entire codebase written in Floats because he doesn't need precesion of doubles, while other wants to operate on `Bytes` to preserve space. 

Based on my experience the only viable approaches I see are transitive specialization and numbers-encoded-in-pointers(smi's and similar ones). As we can't go second route on JVM, we are left with a choice of what mechanism should transitive specialization be based on, extremes being .Net-like runtime reification or C++-like template expansion.

The original strawman prototypes have been successfully specialized using analysis in Linker, I'd assume it makes sense to continue checking that it can be specialized from time to time.
We should also keep our attention on specialization that Oracle is working on in project Valhalla. So far, we are, unfortunately, very incompatible with it and we'll need to re-ingeneer substantial parts of both compilers  when/if Valhalla ships.

szeiger:
I agree in principle and I'd certainly prefer automatic transitive specialization over the mess we have to deal with at the moment. Are there any plans to bring this to Scala 2 or are we stuck with the status quo for the foreseeable future (Scala 2.13 / 2.14)?

In the latter case we either need to live with `@specialized` plus some additional, manual hacks, or ignore the problem.

Ichoran:
There's also the issue that there are still a non-negligible number of bugs in specialization, and for this to be meaningful we'd have to go through and test to make sure everything is specialized.  Not impossible, but not entirely trivial either.  (We'd have to look at bytecode.)

I would really benefit from specialization--it is hard to overstate how tired I am of writing while loops on arrays--but it's really important that it work robustly or it's just a distraction from the more effective method of writing while loops on arrays.

Anyway, I think as a proof of concept that it can be done at all, this is a success.  I'm less certain that it's a good idea.

szeiger:
I keep hearing about these bugs in specialization but in my use of it in Slick I didn't encounter anything that I would call a bug. There are limitations where specialization requires some hand-holding (runtime type checks plus casts to specialized types) but in the end it worked very well in Slick for composing allocation-free ResultSet readers.

Ichoran:
Well, I've personally hit https://issues.scala-lang.org/browse/SI-5490 and https://issues.scala-lang.org/browse/SI-6804 and https://issues.scala-lang.org/browse/SI-7659 and https://issues.scala-lang.org/browse/SI-4790 and https://issues.scala-lang.org/browse/SI-8247 and some others, but I think that's enough for now.

szeiger:
I've added a commit that implements `map` and `flatMap`. It's rather ugly but from what I've seen in the bytecode it is specialized correctly. I'm waiting for the benchmarking code to be merge before running any benchmarks.

src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
```diff
@@ -57,17 +62,145 @@ class ArrayBuffer[A] private (initElems: Array[AnyRef], initLength: Int)
...
+          a(i) = ff(elems(start + i))
+          i += 1
+        }
+        a
+      case _ =>
+        val ff = f.asInstanceOf[A => AnyRef]
+        val a = new Array[AnyRef](length)
+        while(i < length) {
+          // Note: Function1 is not specialized for AnyRef, so even if A is a primitive type
+          // we cannot call ff without boxing the argument.
```
Ichoran: This is true, but note that we _could_ pick out the other things that `Function1` is specialized for even if we can't store them that way, in order to save the boxing on the way in.  Also, note that formally `A => AnyRef` is incorrect, so if it *was* specialized on AnyRef, the cast would throw an error.  Any harm in leaving it as `A => Any`?

Ichoran:
That code is awfully ugly.  Is there any way to abstract it into a SpecializedArrayBackedLike class so we don't have to write that mess for *every* leaf class we want specialized?

src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
szeiger: > we *could* pick out the other things that `Function1` is specialized for even if we can't store them that way, in order to save the boxing on the way in

Right, we could do this manually here. The bigger question is how far we want to go with primitive types. I only implemented specialization for Int, Long and Double, the same way that Java streams do. If we specialized for all types supported by Function1 Scala specialization would take care of this case.

> Any harm in leaving it as `A => Any`?

Then we need a cast in `a(i) = ff(elems(start + i))`. This would probably be the safer choice.

szeiger:
Yes, I expect that some parts can be generalized, in particular building from a possibly specialized iterator.

SethTisue:

PR #5, opened 2016-12-19
Make it clearer that the `iterator` method will return a fresh Iterator at each call

julienrf:
Also removes the cycle between `Iterator` and `IterableOnce`.

### DIFF
--- a/src/test/scala/strawman/collection/test/Test.scala
+++ b/src/test/scala/strawman/collection/test/Test.scala
@@ -121,9 +121,9 @@ class StrawmanTest {
     val ys7: String = xs7
     val xs8 = xs.drop(2)
     val ys8: String = xs8
-    val xs9 = xs.map(_ + 1)  // !!! need a language change to make this work without the : Char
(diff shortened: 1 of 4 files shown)


src/test/scala/strawman/collection/test/Test.scala
```diff
@@ -121,9 +121,9 @@ class StrawmanTest {
     val ys7: String = xs7
     val xs8 = xs.drop(2)
     val ys8: String = xs8
-    val xs9 = xs.map(_ + 1)  // !!! need a language change to make this work without the : Char
+    val xs9 = xs.map(_ + 1)
```
DarkDimius: why was the comment removed?

julienrf: Because the type ascription has already gone.

DarkDimius: Ah, i see. Scalac has already introduced same overload resolution rule as dotty did. Thanks!

odersky:
LGTM 👍

PR #8, opened 2016-12-26
Equality between collections

julienrf:
- Override equals in the immutable hierarchy to compare the collections’ elements.

This introduces a change compared to the current collections. Indeed, the current collections always compare their elements, even mutable collections:

~~~ scala
scala> List(1, 2, 3) == Vector(1, 2, 3)
res0: Boolean = true

scala> List(1, 2, 3) == collection.mutable.ListBuffer(1, 2, 3)
res1: Boolean = true
~~~

I think it would be better to compare the elements only in the case of immutable collections.

To achieve this change, I introduced an immutable hierarchy (a base `immutable.Iterable` trait, etc.) that is not shared with mutable collections. The main drawback is that now we have `collection.Iterable` and `collection.immutable.Iterable` (and same for `Seq`), which might be confusing. I’d like to find distinct names.

### DIFF
--- a/src/main/scala/strawman/collection/immutable/Iterable.scala
+++ b/src/main/scala/strawman/collection/immutable/Iterable.scala
@@ -1,0 +1,40 @@
+package strawman.collection.immutable
+
+import scala.{Any, Boolean, Int}
+import scala.util.hashing.MurmurHash3
+
+/**
+  * Base class of immutable collections. (Must *not* be inherited by mutable collections)
+  */
+trait Iterable[+A]
+  extends strawman.collection.Iterable[A]
+  with strawman.collection.IterableLike[A, Iterable] {
+
+  override def equals(o: Any): Boolean = {
+    o match {
+      case iterable: Iterable[A] => sameElements(iterable)
+      case _ => false
+    }
+  }
+
+  override def hashCode(): Int =
+    IterableUtils.orderedHash(this, IterableUtils.iterableSeed)
+
+}
+
+// Temporary: TODO move to MurmurHash3.scala
+object IterableUtils {
+
+  val iterableSeed: Int = "Iterable".##
+
+  final def orderedHash(xs: Iterable[_], seed: Int): Int = {
+    var n = 0
+    var h = seed
+    xs foreach { x =>
+      h = MurmurHash3.mix(h, x.##)
+      n += 1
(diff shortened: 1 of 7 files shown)


src/main/scala/strawman/collection/immutable/Iterable.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import scala.{Any, Boolean, Int}
+import scala.util.hashing.MurmurHash3
+
+/**
+  * Base class of immutable collections. (Must *not* be inherited by mutable collections)
+  */
+trait Iterable[+A]
+  extends strawman.collection.Iterable[A]
+  with strawman.collection.IterableLike[A, Iterable] {
```
julienrf: Note that the `immutable.Iterable` trait inherits from the “mutability-agnostic” trait `Iterable`. I don’t think this is a problem since this trait only provides read operations.

Ichoran:
Can you explain why you think this is better?  This seems on the surface to be a step backwards for ease of use.  Scala takes pains to make sure (4L: Any) == (4.0: Any), for instance.  What does this change gain us / what problem does it solve?

odersky:
I also think we should not do this. Arguments can go both ways whether it's better to do element-wise comparison for all collections or reference-based equality for mutable ones. In the end I prefer the uniformity of the "elementwise everywhere" rule, in particular since there are collection types which can have mutable and immutable instances. 

But in any case I believe this is a point where backwards compatibility rules everything else. Changing equality of collections will lead to change in behavior without changing the types. This I believe we must avoid wherever we can, since these are the hardest porting problems to solve.

lihaoyi:
FWIW, in Python mutable lists are compared element-wise and nobody has a problem with it. If anyone wants reference equality, it's available using `is` (`eq` in Scala) just like it is for every other value in your program.

In Javascript mutable lists are compared by reference equality and there's a [400-vote stackoverflow question on how to do it "properly"](http://stackoverflow.com/questions/7837456/how-to-compare-arrays-in-javascript), with multiple libraries (lodash, underscore) providing element-by-element equality.

Similarly, Java's mutable arrays' reference-equality generally causes headaches and AFAICT people don't like it (http://stackoverflow.com/questions/8777257/equals-vs-arrays-equals-in-java)

PR #10, opened 2016-12-28
Split code to several files.

julienrf:
- Introduce `strawman.collection.mutable` and `strawman.collection.immutable` packages.
- `Seq` is still in `strawman.collection` package because it is implemented by both mutable (e.g. `ArrayBuffer`) and immutable (e.g. `List`) concrete collections.


odersky:
LGTM. I think this is a very useful first step towards having things in separate files.

PR #11, opened 2017-01-03
Add Range

julienrf:
`Range` is interesting because it is monomorphic.

This work is based on the `immutable-array` branch.

This implementation differs from Scala 2.12’s implementation because it uses a single type `Range` to represent both inclusive and non-inclusive ranges. I think that the only consequence is that it is not anymore possible to use a “type pattern” to match inclusive ranges (in 2.12 we can write `case r: Range.Inclusive`). Am I missing something else? I found that having a unique `Range` type was simpler but I will be happy to use the same design as in 2.12 for the sake of compatibility, if needed.

Also, which concrete collection should we use for operations that return an `IndexedSeq[Int]` (e.g. `filter`)? Here, I chose to use `immutable.Array[Int]`, but maybe `immutable.Vector[Int]` will provide the best performance / size trade-off.

### DIFF
--- a/src/main/scala/strawman/collection/immutable/Range.scala
+++ b/src/main/scala/strawman/collection/immutable/Range.scala
@@ -64,7 +64,7 @@ final class Range(
       || (start == end && !isInclusive)
     )
 
-  private val numRangeElements: Int = {
+  val length: Int = {
     if (step == 0) throw new IllegalArgumentException("step cannot be 0.")
     else if (isEmpty) 0
     else {
@@ -100,8 +100,8 @@ final class Range(
   override def tail: Range = {
     if (isEmpty)
       Nil.tail
-
-    drop(1)
+    if (length == 1) newEmptyRange(end)
+    else new Range(start + step, end, step, isInclusive)
   }
 
   protected def copy(start: Int = start, end: Int = end, step: Int = step, isInclusive: Boolean = isInclusive): Range =
@@ -115,8 +115,7 @@ final class Range(
   def by(step: Int): Range = copy(start, end, step)
 
   // Override for performance
-  override def size: Int = numRangeElements
-  def length: Int = numRangeElements
+  override def size: Int = length
 
   // Check cannot be evaluated eagerly because we have a pattern where
   // ranges are constructed like: "x to y by z" The "x to y" piece
@@ -124,7 +123,7 @@ final class Range(
   // which means it will not fail fast for those cases where failing was
   // correct.
   private def validateMaxLength() {
-    if (numRangeElements < 0)
+    if (length < 0)
       fail()
   }
   private def description = "%d %s %d by %s".format(start, if (isInclusive) "to" else "until", end, step)
@@ -132,7 +131,7 @@ final class Range(
 
   def apply(idx: Int): Int = {
     validateMaxLength()
-    if (idx < 0 || idx >= numRangeElements) throw new IndexOutOfBoundsException(idx.toString)
+    if (idx < 0 || idx >= length) throw new IndexOutOfBoundsException(idx.toString)
     else start + (step * idx)
   }
 
@@ -162,7 +161,7 @@ final class Range(
     */
   override def take(n: Int): Range =
     if (n <= 0 || isEmpty) newEmptyRange(start)
-    else if (n >= numRangeElements && numRangeElements >= 0) this
+    else if (n >= length && length >= 0) this
     else {
       // May have more than Int.MaxValue elements in range (numRangeElements < 0)
       // but the logic is the same either way: take the first n
@@ -176,7 +175,7 @@ final class Range(
     */
   override def drop(n: Int): Range =
     if (n <= 0 || isEmpty) this
-    else if (n >= numRangeElements && numRangeElements >= 0) newEmptyRange(end)
+    else if (n >= length && length >= 0) newEmptyRange(end)
     else {
       // May have more than Int.MaxValue elements (numRangeElements < 0)
       // but the logic is the same either way: go forwards n steps, keep the rest
--- a/src/main/scala/strawman/collection/immutable/Range.scala
+++ b/src/main/scala/strawman/collection/immutable/Range.scala
@@ -51,16 +51,7 @@ final class Range(
     with SeqMonoTransformsFromIterable[Int, IndexedSeq[Int]]
     with Serializable { range =>
 
-  def iterator(): Iterator[Int] =
-    new Iterator[Int] {
-      private var current: Seq[Int] = range
-      def next() = {
-        val x = current.head
-        current = current.tail
-        x
-      }
-      def hasNext = !current.isEmpty
-    }
+  def iterator(): Iterator[Int] = new RangeIterator(start, step, lastElement, isEmpty)
 
   private def gap           = end.toLong - start.toLong
   private def isExact       = gap % step == 0
@@ -277,4 +268,25 @@ object Range {
 
 
 
+}
+
+/**
+  * @param lastElement The last element included in the Range
+  * @param initiallyEmpty Whether the Range was initially empty or not
+  */
+private class RangeIterator(
+  start: Int,
+  step: Int,
+  lastElement: Int,
+  initiallyEmpty: Boolean
+) extends Iterator[Int] {
+  private var _hasNext: Boolean = !initiallyEmpty
+  private var _next: Int = start
+  def hasNext: Boolean = _hasNext
+  def next(): Int = {
+    val value = _next
+    _hasNext = value != lastElement
+    _next = value + step
+    value
+  }
 }
--- a/src/main/scala/strawman/collection/immutable/Range.scala
+++ b/src/main/scala/strawman/collection/immutable/Range.scala
@@ -1,0 +1,280 @@
+package strawman.collection.immutable
+
+import strawman.collection.mutable.Iterator
+import strawman.collection.{IndexedSeq, Iterable, IterableFactory, IterableOnce, IterablePolyTransformsFromIterable, Seq, SeqLike, SeqMonoTransformsFromIterable}
+
+import scala.{Boolean, IllegalArgumentException, IndexOutOfBoundsException, Int, inline, Long, StringContext, Serializable, SerialVersionUID, Unit}
+import scala.Predef.{???, augmentString}
+import java.lang.String
+
+/** The `Range` class represents integer values in range
+  *  ''[start;end)'' with non-zero step value `step`.
+  *  It's a special case of an indexed sequence.
+  *  For example:
+  *
+  *  {{{
+  *     val r1 = 0 until 10
+  *     val r2 = r1.start until r1.end by r1.step + 1
+  *     println(r2.length) // = 5
+  *  }}}
+  *
+  *  Ranges that contain more than `Int.MaxValue` elements can be created, but
+  *  these overfull ranges have only limited capabilities. Any method that
+  *  could require a collection of over `Int.MaxValue` length to be created, or
+  *  could be asked to index beyond `Int.MaxValue` elements will throw an
+  *  exception. Overfull ranges can safely be reduced in size by changing
+  *  the step size (e.g. `by 3`) or taking/dropping elements. `contains`,
+  *  `equals`, and access to the ends of the range (`head`, `last`, `tail`,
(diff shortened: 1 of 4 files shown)


src/main/scala/strawman/collection/immutable/Range.scala
```diff
@@ -0,0 +1,280 @@
...
+    with IterablePolyTransformsFromIterable[Int, IndexedSeq]
+    with SeqMonoTransformsFromIterable[Int, IndexedSeq[Int]]
+    with Serializable { range =>
+
+  def iterator(): Iterator[Int] =
+    new Iterator[Int] {
+      private var current: Seq[Int] = range
+      def next() = {
+        val x = current.head
+        current = current.tail
```
sjrd: We can do *much* better than `current.tail`. If `current` was simply an `Int`, it's `current += step`. This is important because create a `Range` has a non-trivial cost. Iterating over a `Range` should not have to allocate $n$ instances of `Range`!

sjrd: Building on top of the algorithm of `foreach` (which has been refined many times in the history of `Range`):
```scala
def iterator(): Iterator[Int] = new Iterator[Int] {
  private var _hasNext: Boolean = !isEmpty
  private var _next: Int = start
  def hasNext: Boolean = _hasNext
  def next(): Int = {
    val value = _next
    _hasNext = value != lastElement
    _next = value + step
    value
  }
}
```

Ichoran: Agreed.  The iterator should get the relevant values from `Range` and simply `+=` through them by `step`, stopping when it hits `lastElement`.  Termination will need to be stored in a boolean and/or hackily stored as a `step` size of 0, so the internal fields will be `value`, `step`, `lastValue`, and maybe `completed`.

src/main/scala/strawman/collection/immutable/Range.scala
```diff
@@ -0,0 +1,280 @@
...
+    *
+    *  $doesNotUseBuilders
+    *
+    *  @return  a new range consisting of all the elements of this range except the first one.
+    */
+  override def tail: Range = {
+    if (isEmpty)
+      Nil.tail
+
+    drop(1)
```
Ichoran: Going through `drop(1)` does a huge amount of unnecessary computation.  We already know it's not empty.  We need only `if (numRangeElements == 1) newEmptyRange(end) else new Range(start + step, end, step, isInclusive)`

src/main/scala/strawman/collection/immutable/Range.scala
```diff
@@ -0,0 +1,280 @@
...
+  /** Create a new range with the `start` and `end` values of this range and
+    *  a new `step`.
+    *
+    *  @return a new range with a different step
+    */
+  def by(step: Int): Range = copy(start, end, step)
+
+  // Override for performance
+  override def size: Int = numRangeElements
+  def length: Int = numRangeElements
```
Ichoran: Any reason not to get rid of `numRangeElements` and just use `size` or `length`?

src/main/scala/strawman/collection/immutable/Range.scala
Ichoran: @sjrd - I'd tend to internalize `_step` and `_lastElement` so that the original Range could be GCed (and maybe wouldn't even be created if escape analysis succeeded), but that algorithm looks good.

sjrd: > I'd tend to internalize _step and _lastElement

Good point.

julienrf: Thanks for your remarks, I added a commit providing a better `iterator` implementation.

src/main/scala/strawman/collection/immutable/Range.scala
julienrf: Not to my knowledge. I removed `numRangeElements`. Thanks!

szeiger:
> `Range` is interesting because it is monomorphic

There's also the constrained polymorphic `NumericRange`. Can we unify them without a loss of performance for `Range`? It would be nice to unify them at least from the user's point of view, even if they are backed by different implementations for efficiency reasons.

julienrf:
You mean that from the users point of view we would have `type Range = NumericRange[Int]`?

julienrf:
@szeiger I think we can easily have a special implementation of `NumericRange[Int]`, but then it’s not easy to have zero overhead when _constructing_ a range. I see the following two solutions:

- Looking at the supplied `Integral[N]` instance:

~~~ scala
object NumericRange {
  def apply[N](start: N, end: N, step: N)(implicit num: Integral[N]): NumericRange[N] =
    if (num == Integral.Int) /* specialized Int range */
    else /* generic numeric range */
}
~~~

The drawback is that we still would have to box and unbox the bounds once.

- Overloaded constructors:

~~~ scala
object NumericRange {
  def apply(start: Int, end: Int, step: Int): NumericRange[Int] = …
  def apply[N](start: N, end: N, step: N)(implicit num: Integral[N]): NumericRange[N] = …
}
~~~

The goal is to avoid boxing the range parameters at creation, but then if we write `NumericRange('a', 'z', 'b' - 'a')` the parameters are converted to `Int` and you get a `NumericRange[Int]` instead of a `NumericRange[Char]` (I’m not sure that’s the way overloading should work, BTW). Note that if you explicitly instantiate `N` to `Char` you get the expected `NumericRange[Char]`.

For information, I implemented `NumericRange` in the strawman and noticed a 10x slowdown between the generic `NumericRange[Int]` and `Range`.

Last side remark about naming. `Range` has the advantage of being short. I would not be super happy to replace all my `Range` occurrences with `NumericRange[Int]`.

szeiger:
I agree in principle that we want to avoid the boxing overhead but unless we support specialized collections it's probably inconsequential in this case because all values produced by the `Range` still have to be boxed anyway.

The overloaded version should work if we also overload for `Byte`, `Short` and `Char` (i.e. all primitive numeric types smaller than `Int`) to catch the widening conversions.

Ichoran:
Range is extremely important for fast indexing.  One of the oldest and most highly complained-about drawbacks of Scala was that
```scala
for (i <- 0 until 1000) foo(i)
```
was slower than the while-loop equivalent.  Now it's often the same.  Making this slow again is very consequential.  Range absolutely cannot be boxed.

That said, it's fine if Range is a NumericRange[Int].  We just need to be careful about how we allow various numeric ranges to be constructed so that we can carefully build an optimized Range by hand and get fully optimized performance at least when the exact type is known (and not store unused AnyRef copies of vals as specialization normally does).

Building ranges efficiently is just a matter of tedium: use method 2 that @julienrf proposed, and don't overload just for `Int` but for anything that auto-converts to `Int`.

szeiger:
I hadn't thought about how `Range.foreach` is specialized before, in fact I assumed it's not. For anyone else wondering, this is the definition in `Range`

```
  @inline final override def foreach[@specialized(Unit) U](f: Int => U) {
```

which leads to an overloaded method:

```
public final void foreach$mVc$sp(scala.Function1<java.lang.Object, scala.runtime.BoxedUnit>);
```

The body of that method can call the `Int => Unit` overloaded `apply` of `Function1`:

```
85: invokeinterface #399,  2          // InterfaceMethod scala/Function1.apply$mcVI$sp:(I)V
```

There is a second non-specialized copy of `foreach` which has to box every value (because only *combinations* of all type parameters can be specialized; there is no way to call an `Int => _` without boxing).

A downside of the current implementation is that it relies on static dispatch.  If you upcast from `Range` to a supertype you always get the boxed version. It should be possible to perform that check at runtime instead (but the extra check could negatively impact performance of the static dispatch case and there is no way to have separate implementations for both). A related problem is that `foreach` accepts functions with arbitrary return types to avoid type-inference issues but using anything other than `Unit` is a lot slower.

Ichoran:
@szeiger - I agree that these are drawbacks of the current design; do you have a suggestion for how to overcome some of them without losing the highly desirable "for loop is as fast as a while loop" property?  Shall we test whether the runtime check can work without penalty in a static context?  I guess that's pretty important to know.  I guess I'll do a quick check now.

Ichoran:
I'm getting pretty encouraging results.  Looks like a _single_ test doesn't clobber performance in the simplest case. Code I used is here, along with results: https://gist.github.com/Ichoran/9a2344a5567832423b63ac3989813d4b

So it looks, preliminarily, as though either approach is okay.  But we definitely need _some_ approach.

sjrd:
It would be good to check that property on Scala.js too. In the current collections library, a `Range.foreach` actually *becomes* a while loop in the JS code, and without any `Long` operation for common cases to boot (e.g., when `start = 0`).

I'll try to play with this eventually, but for the next 1.5 weeks I don't have any time to allocate to this.

szeiger:
An alternative would be to make `foreach` be `@specialized(Unit)` in `Iterable` so that specialized subclasses would automatically get appropriately specialized implementations.

Ichoran:
Yes, but then what about `map`?  What about `filter`? If we go that route we have to go pretty far on that route to avoid surprising performance behavior.

szeiger:
Well, that's true for both approaches. Currently no method other than `foreach` is specialized.

I think it would also be worth trying out what happens when we simply specialize *everything* (up to `Iterable` and `Iterator`). How bad is the code size overhead? How much performance improvements do we get without a lot of manual work?

Ichoran:
@szeiger - Agreed, that'd probably be instructive.  Do you want to try it out or shall I?  (I guess we're the two most obvious people to do it.)

szeiger:
@Ichoran I think we should wait until we have a more complete library so we can better judge the code size and performance impact. If it turns out not to be a viable option we'll have to fall back to the current approach.

julienrf:
@Ichoran Which JVM do you use? With OpenJDK 1.8 I get the following result:

~~~
While loop
Benchmark comparison (in 2.489 s)
Significantly different (p ~= 1.649e-12)
  Time ratio:    1.05184   95% CI 1.03882 - 1.06487   (n=45)
    First     464.3 ns   95% CI 459.2 ns - 469.5 ns
    Second    488.4 ns   95% CI 485.8 ns - 491.0 ns
Benchmark comparison (in 721.6 ms)
Significantly different (p ~= 3.799e-08)
  Time ratio:    0.95029   95% CI 0.93607 - 0.96450   (n=20)
    First     473.4 ns   95% CI 468.5 ns - 478.2 ns
    Second    449.8 ns   95% CI 444.9 ns - 454.7 ns

Ranged
Benchmark comparison (in 745.9 ms)
Significantly different (p ~= 3.782e-12)
  Time ratio:    0.93301   95% CI 0.91993 - 0.94608   (n=20)
    First     477.1 ns   95% CI 472.6 ns - 481.7 ns
    Second    445.1 ns   95% CI 440.6 ns - 449.7 ns
Benchmark comparison (in 732.2 ms)
Not significantly different (p ~= 0.1628)
  Time ratio:    0.98669   95% CI 0.96800 - 1.00537   (n=20)
    First     473.1 ns   95% CI 466.8 ns - 479.4 ns
    Second    466.8 ns   95% CI 460.5 ns - 473.1 ns

Runged
Benchmark comparison (in 858.9 ms)
Significantly different (p ~= 0)
  Time ratio:    4.61509   95% CI 4.50469 - 4.72548   (n=20)
    First     443.3 ns   95% CI 435.3 ns - 451.3 ns
    Second    2.046 us   95% CI 2.014 us - 2.078 us
Benchmark comparison (in 800.7 ms)
Significantly different (p ~= 0)
  Time ratio:    4.93238   95% CI 4.72537 - 5.13940   (n=20)
    First     448.4 ns   95% CI 433.7 ns - 463.0 ns
    Second    2.211 us   95% CI 2.153 us - 2.270 us
~~~

Ichoran:
@julienrf - I have
```
openjdk version "1.8.0_121"
OpenJDK Runtime Environment (build 1.8.0_121-8u121-b13-0ubuntu1.16.04.2-b13)
```
I probably ran it on 2.11, though.  I don't recall.  Did you use 2.12?  Maybe the new lambda encoding is to blame (again)?

julienrf:
The problem comes from the Scala version indeed. With 2.11 I get the same results as you. The regression happens when I use 2.12. @retronym Can we assume that this regression will be fixed for 2.13?

julienrf:
Superseeded by #29.

PR #14, opened 2017-01-13
[WIP] First version to add mutable collections

odersky:
This is currently WIP.

### DIFF
--- a/src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
+++ b/src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
@@ -1,28 +1,40 @@
 package strawman.collection.mutable
 
-import scala.{Array, Int, Boolean, Unit, AnyRef}
+import java.lang.IndexOutOfBoundsException
+import scala.{Array, Int, Long, Boolean, Unit, AnyRef}
+import strawman.collection.{IterableFactory, Iterable, IterableOnce, SeqLike, IndexedView}
 import scala.Predef.intWrapper
-import strawman.collection.{IndexedView, Iterable, IterableFactory, IterableOnce, Seq, SeqLike}
 
 /** Concrete collection type: ArrayBuffer */
 class ArrayBuffer[A] private (initElems: Array[AnyRef], initLength: Int)
-  extends Seq[A]
+  extends IndexedOptimizedGrowableSeq[A]
     with SeqLike[A, ArrayBuffer]
     with Buildable[A, ArrayBuffer[A]]
     with Builder[A, ArrayBuffer[A]] {
 
   def this() = this(new Array[AnyRef](16), 0)
 
-  private var elems: Array[AnyRef] = initElems
-  private var start = 0
+  private var array: Array[AnyRef] = initElems
   private var end = initLength
 
-  def apply(n: Int) = elems(start + n).asInstanceOf[A]
+  /** Ensure that the internal array has at least `n` cells. */
+  private def ensureSize(n: Int): Unit =
+    array = RefArrayUtils.ensureSize(array, end, n)
 
-  def length = end - start
+  /** Reduce length to `n`, nulling out all dropped elements */
+  private def reduceToSize(n: Int): Unit = {
+    RefArrayUtils.nullElems(array, n, end)
+    end = n
+  }
+
+  def apply(n: Int) = array(n).asInstanceOf[A]
+
+  def update(n: Int, elem: A): Unit = array(n) = elem.asInstanceOf[AnyRef]
+
+  def length = end
   override def knownSize = length
 
-  override def view = new ArrayBufferView(elems, start, end)
+  override def view = new ArrayBufferView(array, end)
 
   def iterator() = view.iterator()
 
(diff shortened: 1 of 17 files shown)


src/main/scala/strawman/collection/mutable/Seq.scala
```diff
@@ -0,0 +1,76 @@
+package strawman
+package collection
+package mutable
+
+trait Seq[A] extends strawman.collection.Seq[A] with Growable[A] {
```
Ichoran: This means that `Array` cannot be wrapped as a `Seq`.  Is this intentional?

src/main/scala/strawman/collection/mutable/Seq.scala
```diff
@@ -0,0 +1,76 @@
...
+  // +=, ++=, clear inherited from Growable
+  def +=:(elem: A): this.type = { insert(0, elem); this }
+  def +=:(elem1: A, elem2: A, elems: A*): this.type = elem1 +=: elem2 +=: elems.toStrawman ++=: this
+  def ++=:(elems: IterableOnce[A]): this.type = { insertAll(0, elems); this }
+
+  def dropInPlace(n: Int): this.type = { remove(0, n); this }
+  def dropRightInPlace(n: Int): this.type = { remove(length - n, n); this }
+  def takeInPlace(n: Int): this.type = { remove(n, length); this }
+  def takeRightInPlace(n: Int): this.type = { remove(0, length - n); this }
+  def sliceInPlace(start: Int, end: Int): this.type = takeInPlace(end).dropInPlace(start)
```
Ichoran: As a strawman this is fine, but obviously it will need to be overridden for collections that can't take/drop efficiently (e.g. because they don't have a last element index which makes `takeInPlace` O(1)).

Ichoran:
This looks reasonable.  I think we might need more fine-grained traits to distinguish between fixed-length collections (where the contents can be altered) and growable things, but that doesn't undermine the strategy.

src/main/scala/strawman/collection/mutable/Seq.scala
odersky: My assumption was that most mutable sequences have indexed access. The only exception I know of is LIstBuffer. And, yes, `sliceInPlace` should be overridden for this one.

src/main/scala/strawman/collection/mutable/Seq.scala
odersky: No, not intentional. I am not sure yet what the right scehem should be. It seems natural that mutable sequences should not just have `update`, but also `insert`, `remove`, `+=`. But then what to do about `WrappedArray`? One possibility is to also have these things for `WrappedArray` and accept that e.g. `+=` has linear cost. That's what current collections seem to do. Another possibility would be to back `WrappedArray` by a `ResizableArray`.

src/main/scala/strawman/collection/mutable/Seq.scala
Ichoran: `UnrolledBuffer` is another one that we have, and a `SortedSeq` backed by a tree would be another.  It might fit better in `IndexedOptimizedSeq`, since there presumably the exceptions would be vanishingly rare (and most everything else would be an `IndexedOptimizedSeq` anyway).

src/main/scala/strawman/collection/mutable/Seq.scala
Ichoran: The problem with having `+=` and other Growable operations produce a new collection on `WrappedArray` is that your `inPlace` operations no longer operate on the collection you wrapped, and it's not at all obvious that this is what happened, since you might go back to the original array and expect to find the changes there.  One could also envision a mutable proxied Seq backed by whatever was useful (not just `Array`)--the equivalent of a lens for a set of mutable values.  We needn't necessarily support these sorts of things, but I think the cost would be relatively low, just like having a distinction between `Seq` and `IndexedOptimizedSeq` is (hopefully) low.

odersky: Yes, good point. Current collections ignore this issue. Do people have proposals how to solve this? It seems there's no nice option.

src/main/scala/strawman/collection/mutable/Iterator.scala
```diff
@@ -4,9 +4,10 @@ import scala.{Boolean, Int, Unit, Nothing, NoSuchElementException}
 import strawman.collection.{IndexedView, IterableOnce}
 
 /** A core Iterator class */
-trait Iterator[+A] { self =>
+trait Iterator[+A] extends IterableOnce[A] { self =>
```
julienrf: Is this used somewhere? I think it is better to avoid cycles in the hierarchy. Here, the problem is that if someone wants to understand the implementation of the collections it is hard to find a starting point. It is hard to find a first abstraction layer on top of which other abstraction layers are built. If we have a look at any collection and follow its ancestors, we always end up on `IterableOnce`. This one has an `iterator(): Iterator[A]` member, if we have a look at the `Iterator` type, we see that it also exends `IterableOnce`, etc.

BTW, I’m not sure this is related but the new design leads to stack overflowing the scala compiler extremely easily, making it super unproductive to diagnose type errors… This might be due to the cycles in the class hierarchy (there is also one with `View`).

odersky: It's used implicitly in that lots of operations use `IterableOnce` parameters (analogous to `TraversableOnce in the current collections), and the idea is that both an Iterator and an Iterable are valid arguments for this.

src/main/scala/strawman/collection/mutable/Seq.scala
Ichoran: Is there something wrong with the "obvious" solution of having the `Growable` trait implemented on only a subset of collections?  So we'd have `GrowableSeq` that extends both `Growable` and `Seq`, or alternatively `UpdatingSeq` which only allows update, and have `Seq` be `UpdatingSeq with Growable`?

odersky: This is what's wrong:

    scala> collection.mutable.Seq(1, 2, 3)
    res1: scala.collection.mutable.Seq[Int] = ArrayBuffer(1, 2, 3)

So far, even the default `Seq` naturally supports `+=`. I don't think we can take that away. So the less drastic change would probably be to have a `FixedLengthSeq` superclass of `Seq` and have `WrappedArray` inherit from that instead of `Seq`.

Ichoran: That is true, but on the other hand

```scala
scala> val s = collection.mutable.Seq(1)
s: scala.collection.mutable.Seq[Int] = ArrayBuffer(1)

scala> s += 2
<console>:12: error: value += is not a member of scala.collection.mutable.Seq[Int]
       s += 2
         ^
```

so people who are going by the API and not by pattern-matching can't rely on it even now.  And that `Seq` might very well be a `WrappedArray` that doesn't have the operation now.

So I think there's a decent argument for leaving it alone, too.  But having `+=` as a core functionality (and one that was usually lurking there one pattern match away, even if you couldn't count on it) is appealing too, so `FixedLengthSeq` (or maybe `SizedSeq` so it's not too wordy?) is also reasonable.

odersky: Ah, cool. I completely misremembered what `mutable.Seq` was so far. I thought it had `+=` as well as `update`. So, yes, indeed, we can have `GrowableSeq` underneath `Seq`.

src/main/scala/strawman/collection/mutable/Iterator.scala
julienrf: I’m still afraid that this could lead to very surprising behaviors. Consider for instance the following program:

~~~ scala
def f(xs: TraversableOnce[Int]): Unit = {
  println(xs.iterator().next())
  println(xs.iterator().next())
}
~~~

We basically create a new iterator and print its first element, twice in a row. If I run this program I would expect to see the same value printed twice. However, if I pass an `Iterator[Int]` as a parameter to `f`, it will print the first element and then the second element. That’s because `iterator()`’s implementation in `Iterator` returns `this` instead of a fresh iterator.

odersky: But that's why it says `IterableOnce`! It means you can take one `iterator()`, but not several ones. Note that `TraversableOnce` was not part of 2.8 collections. It got added later by @paulp, because we experienced a lot of pain regarding the divergence of `Iterator` and `Iterable`.

src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
```diff
@@ -78,16 +121,46 @@ object ArrayBuffer extends IterableFactory[ArrayBuffer] {
...
+
+  /** Remove elements of this array at indices after `sz`.
+   */
+  def nullElems(array: Array[AnyRef], start: Int, end: Int): Unit = {
+    var i = start
+    while (i < end) {
+      array(i) = null
+      i += 1
+    }
+  }
```
Ichoran: Except for very short arrays, this would be faster as `java.util.Arrays.fill(array, start, end, null)`

Ichoran:
Looks good to me!

src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
```diff
@@ -31,45 +43,76 @@ class ArrayBuffer[A] private (initElems: Array[AnyRef], initLength: Int)
 
   protected[this] def newBuilder = new ArrayBuffer[A]
 
+  def clear() =
```
julienrf: Minor: you should add an explicit return type

src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
```diff
@@ -31,45 +43,76 @@ class ArrayBuffer[A] private (initElems: Array[AnyRef], initLength: Int)
...
-  /** Overridden to use array copying for efficiency where possible. */
-  override def ++[B >: A](xs: IterableOnce[B]): ArrayBuffer[B] = xs match {
-    case xs: ArrayBuffer[B] =>
-      val elems = new Array[AnyRef](length + xs.length)
-      Array.copy(this.elems, this.start, elems, 0, this.length)
-      Array.copy(xs.elems, xs.start, elems, this.length, xs.length)
-      new ArrayBuffer(elems, elems.length)
-    case _ => super.++(xs)
+  def insertAll(idx: Int, elems: IterableOnce[A]): Unit = {
+    if (idx < 0 || idx > end) throw new IndexOutOfBoundsException
```
julienrf: What about introducing a private method `ensureWithinBounds(i: Int): Unit`?

src/main/scala/strawman/collection/mutable/Growable.scala
```diff
@@ -0,0 +1,69 @@
...
+import scala.annotation.tailrec
+import strawman.collection.{toOldSeq, toNewSeq}
+
+/** This trait forms part of collections that can be augmented
+ *  using a `+=` operator and that can be cleared of all elements using
+ *  a `clear` method.
+ *
+ *  @author   Martin Odersky
+ *  @version 2.8
+ *  @since   2.8
```
julienrf: 2.13?

src/main/scala/strawman/collection/mutable/Builder.scala
```diff
@@ -1,27 +1,27 @@
...
-trait Builder[-A, +To] { self =>
+trait Builder[-A, +To] extends Growable[A] { self =>
 
   /** Append an element */
   def +=(x: A): this.type
 
+  /** Clears the contents of this builder.
+   *  After execution of this method the builder will contain no elements.
+   */
+  def clear(): Unit
```
julienrf: Note that we already inherit `clear` from `Growable`. I’m curious if there is a reason to re-define it here (maybe just to update the scaladoc?).

odersky: Probably. I copy-pasted that one from existing sources.

src/main/scala/strawman/collection/mutable/Growable.scala
odersky: Yes, copy-paste fall-out.

src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
odersky: Yes, maybe. But it's not universal. We sometimes need two values to compare.

odersky:
I'll work on this and #18 this Sunday to address reviewers comments. If we add maps and sets it would be good to do this on top of the fleshed out structure of #18.

src/main/scala/strawman/collection/mutable/Builder.scala
odersky: Note that the same holds for `+=`. It's also added in `Builder` for documentation.

odersky:
Superseded by #18

PR #18, opened 2017-01-17
Add base traits of collection hierarchy

odersky:
- adds {mutable,immutable}.Iterable
 - adds immutable.Seq
 - fleshes out inheritance of ...Like traits so that all collection
   base traits transform to themselves.

The design is similar to current collections, in that `map` and other transformers are supported on all collections. I wanted to create that as a baseline first. We can then see what it would take to remove some of this functionality for strict mutable collections and general collections.

It's based on the "add mutable collections" PR.

### DIFF
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -54,7 +54,7 @@ trait IterableOps[+A] extends Any {
   protected def coll: Iterable[A]
   private def iterator() = coll.iterator()
 
-  /** Apply `f` to each element for tis side effects
+  /** Apply `f` to each element for its side effects
    *  Note: [U] parameter needed to help scalac's type inference.
    */
   def foreach[U](f: A => U): Unit = iterator().foreach(f)
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -1,4 +1,5 @@
-package strawman.collection
+package strawman
+package collection
 
 import scala.annotation.unchecked.uncheckedVariance
 import scala.reflect.ClassTag
@@ -54,8 +55,10 @@ trait IterableOps[+A] extends Any {
   protected def coll: Iterable[A]
   private def iterator() = coll.iterator()
 
-  /** Apply `f` to each element for tis side effects */
-  def foreach(f: A => Unit): Unit = iterator().foreach(f)
+  /** Apply `f` to each element for tis side effects
+   *  Note: [U] parameter needed to help scalac's type inference. 
+   */
+  def foreach[U](f: A => U): Unit = iterator().foreach(f)
 
   /** Fold left */
   def foldLeft[B](z: B)(op: (B, A) => B): B = iterator().foldLeft(z)(op)
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -4,8 +4,7 @@ package collection
 import scala.annotation.unchecked.uncheckedVariance
 import scala.reflect.ClassTag
 import scala.{Int, Boolean, Array, Any, Unit, StringContext}
-import java.lang.String
-
+import java.lang.{String, UnsupportedOperationException}
 import strawman.collection.mutable.{ArrayBuffer, StringBuilder}
 
 /** Base trait for generic collections */
@@ -56,7 +55,7 @@ trait IterableOps[+A] extends Any {
   private def iterator() = coll.iterator()
 
   /** Apply `f` to each element for tis side effects
-   *  Note: [U] parameter needed to help scalac's type inference. 
+   *  Note: [U] parameter needed to help scalac's type inference.
    */
   def foreach[U](f: A => U): Unit = iterator().foreach(f)
 
@@ -169,7 +168,9 @@ trait IterableMonoTransforms[+A, +Repr] extends Any {
   def drop(n: Int): Repr = fromIterableWithSameElemType(View.Drop(coll, n))
 
   /** The rest of the collection without its first element. */
-  def tail: Repr = drop(1)
+  def tail: Repr =
+    if (coll.isEmpty) drop(1)
+    else throw new UnsupportedOperationException
 }
 
 /** Transforms over iterables that can return collections of different element types.
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -168,9 +168,10 @@ trait IterableMonoTransforms[+A, +Repr] extends Any {
   def drop(n: Int): Repr = fromIterableWithSameElemType(View.Drop(coll, n))
 
   /** The rest of the collection without its first element. */
-  def tail: Repr =
-    if (coll.isEmpty) drop(1)
-    else throw new UnsupportedOperationException
+  def tail: Repr = {
+    if (coll.isEmpty) throw new UnsupportedOperationException
+    drop(1)
+  }
 }
 
 /** Transforms over iterables that can return collections of different element types.
--- a/src/main/scala/strawman/collection/mutable/Growable.scala
+++ b/src/main/scala/strawman/collection/mutable/Growable.scala
@@ -1,11 +1,3 @@
-/*                     __                                               *\
-**     ________ ___   / /  ___     Scala API                            **
-**    / __/ __// _ | / /  / _ |    (c) 2003-2013, LAMP/EPFL             **
-**  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
-** /____/\___/_/ |_/____/_/ | |                                         **
-**                          |/                                          **
-\*                                                                      */
-
 package strawman.collection.mutable
 
 import strawman.collection.IterableOnce
@@ -16,14 +8,6 @@ import strawman.collection.{toOldSeq, toNewSeq}
 /** This trait forms part of collections that can be augmented
  *  using a `+=` operator and that can be cleared of all elements using
  *  a `clear` method.
- *
- *  @author   Martin Odersky
- *  @version 2.8
- *  @since   2.8
- *  @define coll growable collection
- *  @define Coll `Growable`
- *  @define add  add
- *  @define Add  add
  */
 trait Growable[-A] {
 
@@ -59,6 +43,8 @@ trait Growable[-A] {
       case xs: scala.collection.LinearSeq[_] => loop(xs.asInstanceOf[scala.collection.LinearSeq[A]])
       case xs => xs.iterator() foreach += // Deviation: IterableOnce does not define `foreach`.
     }
+    // @ichoran writes: Right now, this actually isn't any faster than using an iterator
+    // for List. Maybe we should just simplify the code by deferring to iterator foreach?
     this
   }
 
--- a/src/main/scala/strawman/collection/mutable/Growable.scala
+++ b/src/main/scala/strawman/collection/mutable/Growable.scala
@@ -1,0 +1,68 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___     Scala API                            **
+**    / __/ __// _ | / /  / _ |    (c) 2003-2013, LAMP/EPFL             **
+**  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
+** /____/\___/_/ |_/____/_/ | |                                         **
+**                          |/                                          **
+\*                                                                      */
+
+package strawman
+package collection
+package mutable
+
+import scala.annotation.tailrec
+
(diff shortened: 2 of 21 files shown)


src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -169,7 +168,9 @@ trait IterableMonoTransforms[+A, +Repr] extends Any {
   def drop(n: Int): Repr = fromIterableWithSameElemType(View.Drop(coll, n))
 
   /** The rest of the collection without its first element. */
-  def tail: Repr = drop(1)
+  def tail: Repr =
+    if (coll.isEmpty) drop(1)
```
julienrf: It should be `coll.nonEmpty` here .

odersky: Oops, yes. It's too hot here (37 celsius at 9am this morning!) Can't think clearly.

src/main/scala/strawman/collection/mutable/Growable.scala
```diff
@@ -0,0 +1,69 @@
...
+import scala.annotation.tailrec
+import strawman.collection.{toOldSeq, toNewSeq}
+
+/** This trait forms part of collections that can be augmented
+ *  using a `+=` operator and that can be cleared of all elements using
+ *  a `clear` method.
+ *
+ *  @author   Martin Odersky
+ *  @version 2.8
+ *  @since   2.8
```
Ichoran: Version number

src/main/scala/strawman/collection/mutable/Growable.scala
```diff
@@ -0,0 +1,69 @@
...
+  def ++=(xs: IterableOnce[A]): this.type = {
+    @tailrec def loop(xs: scala.collection.LinearSeq[A]): Unit = {
+      if (xs.nonEmpty) {
+        this += xs.head
+        loop(xs.tail)
+      }
+    }
+    xs match {
+      case xs: scala.collection.LinearSeq[_] => loop(xs.asInstanceOf[scala.collection.LinearSeq[A]])
+      case xs => xs.iterator() foreach += // Deviation: IterableOnce does not define `foreach`.
```
Ichoran: Are we sure that performance with `foreach` will equal manual traversal?  If no, we at least should leave a note to check performance later.

Ichoran: Answered my own question: in simple benchmarks they're the same.

src/main/scala/strawman/collection/mutable/Growable.scala
```diff
@@ -0,0 +1,69 @@
...
+   */
+  def ++=(xs: IterableOnce[A]): this.type = {
+    @tailrec def loop(xs: scala.collection.LinearSeq[A]): Unit = {
+      if (xs.nonEmpty) {
+        this += xs.head
+        loop(xs.tail)
+      }
+    }
+    xs match {
+      case xs: scala.collection.LinearSeq[_] => loop(xs.asInstanceOf[scala.collection.LinearSeq[A]])
```
Ichoran: Right now, this actually isn't any faster than using an iterator for List.  Maybe we should just simplify the code by deferring to iterator foreach?

src/main/scala/strawman/collection/mutable/ListBuffer.scala
```diff
@@ -33,29 +41,163 @@ class ListBuffer[A]
...
+  private def removeAfter(prev: Predecessor[A], n: Int) = {
+    @tailrec def ahead(p: List[A], n: Int): List[A] =
+      if (n == 0) p else ahead(p.tail, n - 1)
+    setNext(prev, ahead(getNext(prev), n))
+    len -= n
+  }
+
+  def mapInPlace(f: A => A): this.type = {
+    ensureUnaliased()
+    val buf = new ListBuffer[A]
```
Ichoran: This doesn't look like a mapInPlace to me.

src/main/scala/strawman/collection/mutable/ListBuffer.scala
```diff
@@ -33,29 +41,163 @@ class ListBuffer[A]
...
+      }
+      p.asInstanceOf[Predecessor[A]]
+    }
+
+  private def getNext(p: Predecessor[A]): List[A] =
+    if (p == null) first else p.next
+
+  private def setNext(p: Predecessor[A], nx: List[A]): Unit =
+    if (p == null) first = nx else p.next = nx
+
```
Ichoran: Maybe a `swapInPlace` would be good here to abstract over the `update` and `mapInPlace`-type operations?

src/main/scala/strawman/collection/mutable/ListBuffer.scala
odersky: If we want to avoid re-allocation of nodes, we'd have to make `head` of List mutable as well. But that has far-reaching consequences. Not sure it's worth it.

src/main/scala/strawman/collection/mutable/ListBuffer.scala
```diff
@@ -33,29 +41,163 @@ class ListBuffer[A]
...
+      prev = cur.asInstanceOf[Predecessor[A]]
+      cur = follow
+    }
+    this
+  }
+
+  def patchInPlace(from: Int, patch: collection.Seq[A], replaced: Int): this.type = {
+    ensureUnaliased()
+    val p = locate(from)
+    removeAfter(p, replaced `min` length - from)
```
Ichoran: Precedence is not super-obvious here, especially with point-free backquoted `min`.

src/main/scala/strawman/collection/mutable/Seq.scala
```diff
@@ -0,0 +1,91 @@
...
+    var i = 0
+    while (i < size) { this(i) = f(this(i)); i += 1 }
+    this
+  }
+}
+
+trait IndexedOptimizedGrowableSeq[A] extends IndexedOptimizedSeq[A] with GrowableSeq[A] {
+  def flatMapInPlace(f: A => IterableOnce[A]): this.type = {
+    var i = 0
+    val newElemss = new Array[IterableOnce[A]](size)
```
Ichoran: Isn't it better to just copy the current elements so that the iterators can be GCed along the way?  And presumably there's already/will be an efficient `toArray` method which would simplify the implementation also.

src/main/scala/strawman/collection/mutable/Seq.scala
```diff
@@ -0,0 +1,91 @@
...
+    while (i < size) { newElemss(i) = f(this(i)); i += 1 }
+    clear()
+    i = 0
+    while (i < size) { ++=(newElemss(i)); i += 1 }
+    this
+  }
+  def filterInPlace(p: A => Boolean): this.type = {
+    var i = 0
+    var j = 0
+    while (i < size) {
```
Ichoran: Not sure whether it's worth doing a leading run that just skips up to the first place the condition fails:
```scala
while (i < size && p(apply(i)) i += 1
var j = i
i += 1
```
and then go into `while (i < size)`

src/main/scala/strawman/collection/mutable/Seq.scala
```diff
@@ -0,0 +1,91 @@
...
+  def insertAll(idx: Int, elems: IterableOnce[A]): Unit
+  def remove(idx: Int): A
+  def remove(from: Int, n: Int): Unit
+  def flatMapInPlace(f: A => IterableOnce[A]): this.type
+  def filterInPlace(p: A => Boolean): this.type
+  def patchInPlace(from: Int, patch: collection.Seq[A], replaced: Int): this.type
+
+  // +=, ++=, clear inherited from Growable
+  def +=:(elem: A): this.type = { insert(0, elem); this }
+  def +=:(elem1: A, elem2: A, elems: A*): this.type = elem1 +=: elem2 +=: elems.toStrawman ++=: this
```
Ichoran: I like symmetry, but I'm not so sure that this isn't confusing.  We deprecated this style of left-assignment-in-place somewhere around 2.10, I think because it felt confusing to people and in practice wasn't really used; it's not in the current collections library any more.  I'd recommend leaving all of the `+=:` variants out.

Ichoran:
I made a bunch of fiddly little comments, but overall it looks like a good design.  Not sure if we want to be more granular and add `Shrinkable` as well as `Growable`.

Ichoran:
Mouse error, sorry about the spurious close.

src/main/scala/strawman/collection/mutable/ListBuffer.scala
Ichoran: I guess since we're reallocating everything anyway it doesn't matter whether it's really "in place" or not.  Fair enough--I retract the comment.  It's as inPlace as makes sense when you remove all the places!

julienrf: @Ichoran it is “in place” because at the end the reference returned by the operation is `this` (whereas the usual `map` method returns a new collection).

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -54,8 +54,10 @@ trait IterableOps[+A] extends Any {
   protected def coll: Iterable[A]
   private def iterator() = coll.iterator()
 
-  /** Apply `f` to each element for tis side effects */
-  def foreach(f: A => Unit): Unit = iterator().foreach(f)
+  /** Apply `f` to each element for tis side effects
```
smarter: typo: tis -> its

src/main/scala/strawman/collection/mutable/ListBuffer.scala
szeiger: @odersky What consequences are you referring to? We already run the risk of unpublished values in concurrency race conditions for `tail`. Having the same for `head` wouldn't make things worse.

src/main/scala/strawman/collection/mutable/Seq.scala
```diff
@@ -0,0 +1,91 @@
...
+  def padToInPlace(len: Int, elem: A): this.type = {
+    while (length < len) +=(elem)
+    this
+  }
+}
+
+trait IndexedOptimizedSeq[A] extends Seq[A] {
+  def mapInPlace(f: A => A): this.type = {
+    var i = 0
+    while (i < size) { this(i) = f(this(i)); i += 1 }
```
szeiger: Should this kind of loop first copy `size` to a local var? I always do it out of habit but I have to admit I haven't benchmarked it on any modern JVM.

szeiger:
We could call it `Resizable` instead of `Growable` (and a possible `Shrinkable`) in addition. It indicates that the internal memory of a collection can be reallocated to change the size. In addition to growing we could add an explicit `trim` method (assuming we don't want to auto-shrink on certain operations).

src/main/scala/strawman/collection/mutable/ListBuffer.scala
Ichoran: @julienrf - Yes, but my point was meant to be that having something "in place" throw out everything and recreate it from scratch kind of goes against the idea of "in place".  But I had forgotten that this was not any old linked list but specifically a `ListBuffer`, which means that the data is immutable so if we want to change it we have to reallocate the `::`.  In a different mutable list implementation (e.g. `DoublyLinkedList`), we'd want to change the data without reallocating all the nodes.

src/main/scala/strawman/collection/mutable/Growable.scala
odersky: That's copy pasted code from the current implementation of Growable. I wanted to be conservative ad not touch it. But let's make a note about the alternative.

src/main/scala/strawman/collection/mutable/Seq.scala
odersky: I was worried about the cost of repeated insertions. That's why I used clear() and `++=`. But there's probably a better way to do it. I made a note.

src/main/scala/strawman/collection/mutable/Seq.scala
odersky: Yes, better to assume the worst here.

odersky:
> We could call it Resizable instead of Growable (and a possible Shrinkable) in addition. It indicates that the internal memory of a collection can be reallocated to change the size. In addition to growing we could add an explicit trim method (assuming we don't want to auto-shrink on certain operations).

`trim` is implicitly provided by `dropInPlace`, `dropRightInPlace`

odersky:
@szeiger Good point about adding `trim` methods. If we add them, then indeed `Resizable` will be a better term. But note that `Growable` is inherited from `Builder` which needs none of these niceties. So maybe `Resizable` should be a subtrait of `Growable`?

odersky:
Since we seem to have consensus about this one barring some details I'll merge this now. I tried to left things to do and to discuss in comments.

szeiger:
Good point. Makes sense to have both.

PR #22, opened 2017-01-24
make JOL work

retronym:
- Fork tests using JOL to give it a regular classpath to introspect, rather than the exotic one that SBT in-process running ends up with.
  - Avoid apparent scalac bug with `DelayedInit`, that manifest as:

```
java.lang.IncompatibleClassChangeError: strawman.collection.MemoryFootprint and strawman.collection.MemoryFootprint$delayedInit$body disagree on InnerClasses attribute
```

### DIFF
--- a/benchmarks/memory/src/main/scala/strawman/collection/MemoryFootprint.scala
+++ b/benchmarks/memory/src/main/scala/strawman/collection/MemoryFootprint.scala
@@ -2,13 +2,13 @@ package strawman.collection
 
 import strawman.collection.immutable.{LazyList, List}
 
-import scala.{Any, AnyRef, App, Int, Long}
-import scala.Predef.{ArrowAssoc, println}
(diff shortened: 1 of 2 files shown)


retronym:
Review by @julienrf

julienrf:
Thanks a lot, that’s better!

Now it still fails with an `OutOfMemoryError` for collections with more than 30k elements. I tried with `-Xmx2048M` with no improvement. (Note that with the code that’s in the `benchmark` branch that runs fine)

retronym:
Just took a quick look, it appears that the `GraphLayout` facility in JMH isn't designed for large object graphs. For each instance it finds, it constructs a string with the path from a root object. So of linked structures, with paths like `".next.next.next"`, this representation is quadratic with respect to the length of the list. 

Sorry about that, I hadn't run into that before.

PR #23, opened 2017-01-27
Sketch Set and SortedSet

julienrf:
`SortedSet[A]` can not inherit from `Iterable[A]` because it would have to implement a `fromIterable[B](it: Iterable[B]): SortedSet[B]` , which is not possible because we miss an `Ordering[B]`.

Therefore, the LUB of `SortedSet[A]` and `List[A]` is not `Iterable[A]` but `IterableOnce[A]`. Note that we could pull up a lot of method definitions which are identical in both `SortedSet` and `List` in an intermediate trait, which would sit in between `IterableOnce` and `Iterable`, and which would basically contain all the operations provided by `IterableOps` and `IterableMonoTransforms`. I applied this idea to the `Set` hierarchy only but we could generalize it to `Iterable`.

Before going further, maybe someone has a better idea?

Also, here are some questions about our current design:
- Should we keep both `Buildable` and `FromIterable`? It seems that they overlap. I do prefer the more general `Buildable` thing over `FromIterable` ;
- Should we pull up `mapInPlace` and friends in a `GrowableLike` trait? (currently they are defined in both `mutable.Set` and `mutable.Seq` but they are unrelated inheritance-wise ;
- Should we keep both `XxxOps` and `XxxLike`? I’d be in favor of putting everything in `XxxLike` traits and discard `XxxOps`.

### DIFF
--- a/src/main/scala/strawman/collection/mutable/Set.scala
+++ b/src/main/scala/strawman/collection/mutable/Set.scala
@@ -14,7 +14,7 @@ trait Set[A]
   def +=(elem: A): this.type
   def -=(elem: A): this.type
 
-  def get(elem: A): Option[A] = if (contains(elem)) Some(elem) else None
+  def get(elem: A): Option[A]
 
   def insert(elem: A): Boolean =
     !contains(elem) && { +=(elem); true }
--- a/src/main/scala/strawman/collection/mutable/Set.scala
+++ b/src/main/scala/strawman/collection/mutable/Set.scala
@@ -5,13 +5,16 @@ import strawman.collection.IterableOnce
 import scala.{Int, Boolean, Unit, Option, Some, None}
 import scala.Predef.???
 
-trait Set[A] extends collection.Set[A] with Growable[A] {
+/** Base trait for mutable sets */
+trait Set[A]
+  extends collection.Set[A]
+    with collection.SetLike[A, Set]
+    with Growable[A] {
 
   def +=(elem: A): this.type
   def -=(elem: A): this.type
 
-  def contains(elem: A): Boolean
-  def get(elem: A): Option[A]
+  def get(elem: A): Option[A] = if (contains(elem)) Some(elem) else None
 
   def insert(elem: A): Boolean =
     !contains(elem) && { +=(elem); true }
@@ -57,6 +60,7 @@ trait Set[A] extends collection.Set[A] with Growable[A] {
       -=(elem)
   }
 }
+
 object Set {
-  def apply[A](xs: A*): Set[A] = ???
+  def apply[A](xs: A*): Set[A] = HashSet(xs: _*)
 }
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -7,11 +7,27 @@ import scala.{Int, Boolean, Array, Any, Unit, StringContext}
 import java.lang.{String, UnsupportedOperationException}
 import strawman.collection.mutable.{ArrayBuffer, StringBuilder}
 
-/** Base trait for generic collections */
-trait Iterable[+A] extends IterableOnce[A] with IterableLike[A, Iterable] {
+/** Base trait for generic collections that only support operations
+  * returning collections with the same type of elements `A`.
+  */
+trait EndoIterable[+A]
+  extends IterableOnce[A]
+    with EndoIterableLike[A, EndoIterable] {
+
   /** The collection itself */
   protected def coll: this.type = this
+
 }
+
+/** Base trait for endomorphic collection operations */
+trait EndoIterableLike[+A, +C[X] <: EndoIterable[X]]
+  extends EndoIterableOps[A]
+    with IterableMonoTransforms[A, C[A @uncheckedVariance]] // sound bcs of VarianceNote
+
+/** Base trait for generic collections */
+trait Iterable[+A]
+  extends EndoIterable[A]
+    with IterableLike[A, Iterable]
 
 /** Base trait for Iterable operations
   *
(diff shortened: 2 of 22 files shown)


src/main/scala/strawman/collection/mutable/Set.scala
```diff
@@ -5,13 +5,16 @@ import strawman.collection.IterableOnce
...
+  extends collection.Set[A]
+    with collection.SetLike[A, Set]
+    with Growable[A] {
 
   def +=(elem: A): this.type
   def -=(elem: A): this.type
 
-  def contains(elem: A): Boolean
-  def get(elem: A): Option[A]
+  def get(elem: A): Option[A] = if (contains(elem)) Some(elem) else None
```
julienrf: BTW, @odersky I was not sure what was the purpose of this method.

szeiger: I assume it's supposed to get the canonical representation of `elem` from the set, so this implementation is wrong.

szeiger:
> `SortedSet[A]` can not inherit from `Iterable[A]` because it would have to implement a `fromIterable[B](it: Iterable[B]): SortedSet[B]` , which is not possible because we miss an `Ordering[B]`.

This looks like a major problem with the new design. If we supported an implicit evidence (which could be parameterized) for `fromIterable` we'd also have to add that to all common transformation methods like `map` or `flatMap`, similar to `CanBuildFrom`.

julienrf:
Exactly.

In the existing collections this `Ordering` instance is chained to `CanBuildFrom`.

src/main/scala/strawman/collection/mutable/Set.scala
odersky: @szeiger I agree, implementation is wrong. I wanted it to be the same as `get` on a Java set.

odersky:
This seems to be he first case where `CanBuildFrom` is really needed. It will be hard to do this without. Fortunately, I believe `SortedSet` is not that common. Can we quickly get so me data on how common it is?

odersky:
No, wait, why not model it like `Array` or `String`?

I.e. 
    
    trait Set[A] extends Iterable[A] with SetLike[Set]
    trait SortedSet[A] extends Set[A] with SetLike[Set]

That is the `map`, `flatMap` etc that `SortedSet` inherits from `Set` return again `Set`, not `SortedSet`. 

We then add overloaded methods directly to `SortedSet`:

    def map(f: A => B)(implicit ord: Ordering[B]): SortedSet[B]

I have verified that this works in this simple test case:

    trait Set[A] extends Iterable[A] {
      override def map[B](f: A => B): Set[B] = ???
    }

    trait SortedSet[A] extends Set[A] {
      def map[B](f: A => B)(implicit ev: Ordering[B]): SortedSet[B] = ???
    }

    object Test {
      val ss: SortedSet[Int] = ???
      val s2 = ss.map(_ + 1)
      val s3: SortedSet[Int] = s2
    }

szeiger:
I don't think it's that uncommon. We have basically the same situation (or at least one that can be modeled the same way) in other cases like `BitSet`: You need an extra evidence for the element type to build a collection.

This got me thinking that we could solve it the same way I already had in mind for `BitSet`: Do not build another constrained collection with the inherited methods. If both `SortedSet` and `BitSet` only build a `Set`, we are safe, and we can inherit from all the desirable supertypes. We could have overloaded methods that build another constrained collection, or maybe require the use of views with a special builder at the end.

szeiger:
Looks like I was a minute too late :)

odersky:
@szeiger Yes, same idea. I agree it should be applicable to `BitSet` as well.

Ichoran:
Incidentally, from Olafur's usage data, I count 1200 instances of "SortedSet" or "TreeSet" in the file as compared to 2162 for Queue and 130 for DoubleLinkedList.

But I think we have a potential way forwards, so it's not too important if these numbers are right.  (Grepping javap output is more meaningful for method calls than for non-leaf classes.)

julienrf:
> I have verified that this works in this simple test case, which shows that overloading resolution works as required

I must confess that I’m a bit skeptical about this solution for the following reasons:
- the current [overloading resolution](https://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#overloading-resolution) mechanism is already quite complicated and I’m not even sure it describes our case ;
- it might be confusing for users to have two `map` methods that are used in the exact same way (because one has an implicit parameter which is not visible at use site) but have different return types: how can they choose to use one overload over the other?

What do people think?

odersky:
> the current overloading resolution mechanism is already quite complicated and I’m not even sure it describes our case ;

That's why we have a strawman - let's try it out!

> it might be confusing for users to have two map methods that are used in the exact same way (because one has an implicit parameter which is not visible at use site) but have different return types: how can they choose to use one overload over the other?

They can force the hand of overloading resolution by  adding an explicit return type. IMO that's probably simpler for most people than understand and fiddling with CanBuildFrom.

julienrf:
OK, I’ll implement a new PR with the design you proposed, to see how it feels!

julienrf:
Done in #24.

julienrf:
I pushed a few more commits, introducing `EndoIterable` as a least upper bound between `TreeSet` and `List`. `EndoIterable` is a subclass of `IterableOnce` and a superclass of `Iterable`. It supports all methods of `Iterable` excepted the ones that transform the collection into another collection containing different types of elements (e.g. `map`, `flatMap`, etc.). I also introduced `HashMap`, to see how it fits with this design and the result is mitigated: I still have to duplicate the `map` method definitions because the one inherited by `Iterable` is based on a type constructor of kind `* -> *`, which does not match the kind of `Map`. I tried to implement `TreeMap` but got troubles because `SortedLike` expects a type parameter with a kind that is incompatible with `Map`.

Based on the following comments https://github.com/scala/collection-strawman/pull/24#discussion_r99586474 (and I think @Ichoran also argued about the advantage of having a mechanism similar to `CanBuildFrom`, but I can not find the corresponding comments), I will experiment, in a new branch, a mechanism similar to `CanBuildFrom` to resolve the following issues:
- transformation methods taking an implicit evidence (e.g. `ClassTag` or `Ordering`) ;
- heterogeneous kinds (`Map` vs `Seq`) sharing the same abstractions.

julienrf:
I’m closing this PR since we settled on the design shown in #24.

PR #24, opened 2017-02-06
Overloading based implementation of Set, TreeSet, Map and TreeMap

julienrf:
I sketched the integration of `Set`, `TreeSet`, `Map` and `TreeMap` collections. They were challenging because `Map` takes two type parameters where previous collections were only taking one type parameter, and also because `TreeSet` and `TreeMap` take an additional implicit `Ordering` parameter.

This is an alternative to #23 that makes `Set` extend `Iterable` and `SortedSet` have two overloads of `map` (one has an additional implicit `Ordering` parameter).

### DIFF
--- a/src/main/scala/strawman/collection/Set.scala
+++ b/src/main/scala/strawman/collection/Set.scala
@@ -39,7 +39,8 @@ trait SetLike[A, +C[X] <: Set[X]]
 }
 
 /** Monomorphic transformation operations */
-trait SetMonoTransforms[A, +Repr] {
+trait SetMonoTransforms[A, +Repr]
+  extends IterableMonoTransforms[A, Repr] {
 
   def & (that: Set[A]): Repr
 
--- a/src/main/scala/strawman/collection/Set.scala
+++ b/src/main/scala/strawman/collection/Set.scala
@@ -4,15 +4,23 @@ package collection
 import scala.{Any, Boolean, Equals, Int}
 import scala.util.hashing.MurmurHash3
 
+
+/** Base trait for set collections */
 trait Set[A]
   extends Iterable[A]
     with SetLike[A, Set]
 
+/** Base trait for set operations */
 trait SetLike[A, +C[X] <: Set[X]]
-  extends SetOps[A]
+  extends IterableLike[A, C]
+    with SetMonoTransforms[A, C[A]]
     with Equals {
 
   protected def coll: C[A]
+
+  def contains(elem: A): Boolean
+
+  def subsetOf(that: Set[A]): Boolean
 
   def canEqual(that: Any) = true
 
@@ -20,9 +28,9 @@ trait SetLike[A, +C[X] <: Set[X]]
     that match {
       case set: Set[A] =>
         (this eq set) ||
-        (set canEqual this) &&
-        (coll.size == set.size) &&
-        (this subsetOf set)
+          (set canEqual this) &&
+            (coll.size == set.size) &&
+            (this subsetOf set)
       case _ => false
     }
 
@@ -30,11 +38,12 @@ trait SetLike[A, +C[X] <: Set[X]]
 
 }
 
-trait SetOps[A] extends Any {
+/** Monomorphic transformation operations */
+trait SetMonoTransforms[A, +Repr] {
 
-  protected def coll: Set[A]
+  def & (that: Set[A]): Repr
 
-  def subsetOf(that: Set[A]): Boolean
+  def ++ (that: Set[A]): Repr
 
 }
 
--- a/src/main/scala/strawman/collection/Sorted.scala
+++ b/src/main/scala/strawman/collection/Sorted.scala
@@ -1,0 +1,37 @@
+package strawman.collection
+
+import strawman.collection.mutable.Builder
+
+import scala.Ordering
+
+/** Base trait for sorted collections */
+trait Sorted[A] extends SortedLike[A, Sorted]
+
+trait SortedLike[A, +C[X] <: Sorted[X]]
+  extends SortedPolyTransforms[A, C] {
+
+  def ordering: Ordering[A]
+
+  def range(from: A, until: A): C[A]
+
+}
+
+/** Polymorphic transformation methods on sorted collections */
+trait SortedPolyTransforms[A, +C[X] <: Sorted[X]] {
+
+  def map[B](f: A => B)(implicit ordering: Ordering[B]): C[B]
+
+}
+
+/**
+  * Factories for collections whose elements require an ordering
+  */
+trait OrderingGuidedFactories[C[_]] {
+
+  def builder[A](implicit ordering: Ordering[A]): Builder[A, C[A]]
+
+  def empty[A : Ordering]: C[A] = builder[A].result
+
+  def apply[A : Ordering](as: A*): C[A] = (builder[A] ++= as.toStrawman).result
+
+}
--- a/src/main/scala/strawman/collection/Sorted.scala
+++ b/src/main/scala/strawman/collection/Sorted.scala
@@ -5,14 +5,13 @@ import strawman.collection.mutable.Builder
 import scala.Ordering
 
 /** Base trait for sorted collections */
-trait Sorted[A] extends SortedLike[A, Sorted]
+trait Sorted[A] extends SortedLike[A, Sorted[A]]
 
-trait SortedLike[A, +C[X] <: Sorted[X]]
-  extends SortedPolyTransforms[A, C] {
+trait SortedLike[A, +Repr] {
 
   def ordering: Ordering[A]
 
-  def range(from: A, until: A): C[A]
+  def range(from: A, until: A): Repr
 
 }
 
--- a/src/main/scala/strawman/collection/Sorted.scala
+++ b/src/main/scala/strawman/collection/Sorted.scala
@@ -17,7 +17,8 @@ trait SortedLike[A, +C[X] <: Sorted[X]]
 }
 
 /** Polymorphic transformation methods on sorted collections */
-trait SortedPolyTransforms[A, +C[X] <: Sorted[X]] {
+trait SortedPolyTransforms[A, +C[X] <: Sorted[X]]
+  extends IterablePolyTransforms[A, Iterable] {
 
   def map[B](f: A => B)(implicit ordering: Ordering[B]): C[B]
 
--- a/src/main/scala/strawman/collection/immutable/SortedSet.scala
+++ b/src/main/scala/strawman/collection/immutable/SortedSet.scala
@@ -1,0 +1,21 @@
+package strawman.collection.immutable
+
+import strawman.collection.{SetMonoTransforms, Sorted, SortedLike}
+
+/** Base trait for sorted sets */
+trait SortedSet[A]
+  extends Set[A]
+    with Sorted[A]
+    with SortedSetLike[A, SortedSet] // Inherited SortedSet operations return a `SortedSet`
+
+trait SortedSetLike[A, +C[X] <: SortedSet[X]]
+  extends SortedLike[A, C]
+    with SetLike[A, Set] // Inherited Set operations return a `Set`
+    with SetOps[A, C[A]] // Override the return type of Set ops to return C[A]
+    with SetMonoTransforms[A, C[A]] {
+
+  // We have to write again the definition of methods inherited from SortedPolyTransforms
+  // so that they take precedence over the one inherited from IterablePolyTransforms
+  def map[B](f: A => B)(implicit ordering: scala.Ordering[B]): C[B]
+
+}
--- a/src/main/scala/strawman/collection/immutable/SortedSet.scala
+++ b/src/main/scala/strawman/collection/immutable/SortedSet.scala
@@ -1,6 +1,6 @@
 package strawman.collection.immutable
 
-import strawman.collection.{SetMonoTransforms, Sorted, SortedLike, SortedPolyTransforms}
+import strawman.collection.{Sorted, SortedLike, SortedPolyTransforms}
 
 /** Base trait for sorted sets */
 trait SortedSet[A]
@@ -12,5 +12,4 @@ trait SortedSetLike[A, +C[X] <: SortedSet[X]]
   extends SortedLike[A, C[A]]
     with SortedPolyTransforms[A, C]
     with SetLike[A, Set] // Inherited Set operations return a `Set`
-    with SetOps[A, C[A]] // Override the return type of Set ops to return C[A]
-    with SetMonoTransforms[A, C[A]]
+    with SetMonoTransforms[A, C[A]] // Override the return type of Set ops to return C[A]
--- a/src/main/scala/strawman/collection/immutable/SortedSet.scala
+++ b/src/main/scala/strawman/collection/immutable/SortedSet.scala
@@ -12,10 +12,4 @@ trait SortedSetLike[A, +C[X] <: SortedSet[X]]
   extends SortedLike[A, C]
     with SetLike[A, Set] // Inherited Set operations return a `Set`
     with SetOps[A, C[A]] // Override the return type of Set ops to return C[A]
-    with SetMonoTransforms[A, C[A]] {
-
-  // We have to write again the definition of methods inherited from SortedPolyTransforms
-  // so that they take precedence over the one inherited from IterablePolyTransforms
-  def map[B](f: A => B)(implicit ordering: scala.Ordering[B]): C[B]
-
-}
+    with SetMonoTransforms[A, C[A]]
--- a/src/main/scala/strawman/collection/immutable/SortedSet.scala
+++ b/src/main/scala/strawman/collection/immutable/SortedSet.scala
@@ -1,6 +1,6 @@
 package strawman.collection.immutable
 
-import strawman.collection.{SetMonoTransforms, Sorted, SortedLike}
+import strawman.collection.{SetMonoTransforms, Sorted, SortedLike, SortedPolyTransforms}
 
 /** Base trait for sorted sets */
 trait SortedSet[A]
@@ -9,7 +9,8 @@ trait SortedSet[A]
     with SortedSetLike[A, SortedSet] // Inherited SortedSet operations return a `SortedSet`
 
 trait SortedSetLike[A, +C[X] <: SortedSet[X]]
-  extends SortedLike[A, C]
+  extends SortedLike[A, C[A]]
+    with SortedPolyTransforms[A, C]
     with SetLike[A, Set] // Inherited Set operations return a `Set`
     with SetOps[A, C[A]] // Override the return type of Set ops to return C[A]
     with SetMonoTransforms[A, C[A]]
--- a/src/test/scala/strawman/collection/test/Test.scala
+++ b/src/test/scala/strawman/collection/test/Test.scala
@@ -130,7 +130,7 @@ class StrawmanTest {
     val ys9: Seq[Int] = xs9
     val xs9a = xs.map(_.toUpper)
     val ys9a: String = xs9a
-    val xs10 = xs.flatMap((x: Char) => s"$x,$x")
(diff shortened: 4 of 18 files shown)


src/main/scala/strawman/collection/Sorted.scala
```diff
@@ -0,0 +1,37 @@
...
+  extends SortedPolyTransforms[A, C] {
+
+  def ordering: Ordering[A]
+
+  def range(from: A, until: A): C[A]
+
+}
+
+/** Polymorphic transformation methods on sorted collections */
+trait SortedPolyTransforms[A, +C[X] <: Sorted[X]] {
```
julienrf: This trait is analogous to `IterablePolyTransforms` but polymorphic transformation methods takes an additional implicit `Ordering` parameter.

src/main/scala/strawman/collection/Sorted.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  def map[B](f: A => B)(implicit ordering: Ordering[B]): C[B]
+
+}
+
+/**
+  * Factories for collections whose elements require an ordering
+  */
+trait OrderingGuidedFactories[C[_]] {
```
julienrf: Analogous to `IterableFactories`, but with an implicit `Ordering`.

src/main/scala/strawman/collection/immutable/SortedSet.scala
```diff
@@ -0,0 +1,21 @@
...
+  extends SortedLike[A, C]
+    with SetLike[A, Set] // Inherited Set operations return a `Set`
+    with SetOps[A, C[A]] // Override the return type of Set ops to return C[A]
+    with SetMonoTransforms[A, C[A]] {
+
+  // We have to write again the definition of methods inherited from SortedPolyTransforms
+  // so that they take precedence over the one inherited from IterablePolyTransforms
+  def map[B](f: A => B)(implicit ordering: scala.Ordering[B]): C[B]
+
+}
```
julienrf: One problem is that `Sorted` and `Set` are independent so there is no disambiguation mechanism between their overloaded methods. That’s why I had to re-define the methods of `Sorted` (to make them take precedence over those of `Set`).

I also tried an approach based on intermediate `LowPriority` linear traits, without success:

~~~ scala
trait SortedSetLikeLowPriority1[A, +C[X] <: SortedSet[X]]
  extends SetLike[A, Set]
    with SetOps[A, C[A]]
    with SetMonoTransforms[A, C[A]]

trait SortedSetLikeLowPriority2[A, +C[X] <: SortedSet[X]]
  extends SortedSetLikeLowPriority1[A, C]
    with SortedLike[A, C]

trait SortedSetLike[A, +C[X] <: SortedSet[X]]
  extends SortedSetLikeLowPriority2[A, C]
~~~

src/test/scala/strawman/collection/test/Test.scala
```diff
@@ -288,6 +288,12 @@ class StrawmanTest {
     println(xs17.to(List))
   }
 
+  def sortedSets(xs: strawman.collection.immutable.SortedSet[Int]): Unit = {
+    val xs1 = xs.map((x: Int) => x.toString)
+    val xs2: SortedSet[String] = xs1
+    val xs3: strawman.collection.immutable.Set[String] = xs.map((x: Int) => x.toString)
```
julienrf: Note that the type annotation `x: Int` is required.

src/main/scala/strawman/collection/immutable/SortedSet.scala
```diff
@@ -0,0 +1,21 @@
...
+trait SortedSet[A]
+  extends Set[A]
+    with Sorted[A]
+    with SortedSetLike[A, SortedSet] // Inherited SortedSet operations return a `SortedSet`
+
+trait SortedSetLike[A, +C[X] <: SortedSet[X]]
+  extends SortedLike[A, C]
+    with SetLike[A, Set] // Inherited Set operations return a `Set`
+    with SetOps[A, C[A]] // Override the return type of Set ops to return C[A]
+    with SetMonoTransforms[A, C[A]] {
```
julienrf: There is subtlety here: we inherit from `SetLike[A, Set]` so that methods like `map` return a `Set` rather than a `SortedSet`. However we don’t want `+` or `++` to return a `Set` but a `SortedSet`, hence the `SetOpt[A, C[A]]` and `SetMonoTransforms[A, C[A]]` mixins.

src/main/scala/strawman/collection/Set.scala
```diff
@@ -1,4 +1,27 @@
...
+trait SetLike[A, +C[X] <: Set[X]]
+  extends IterableLike[A, C]
+    with SetMonoTransforms[A, C[A]] {
+
+  def contains(elem: A): Boolean
+
+}
+
+/** Monomorphic transformation operations */
+trait SetMonoTransforms[A, +Repr] {
```
szeiger: `extends IterableMonoTransforms`?

src/main/scala/strawman/collection/Sorted.scala
szeiger: It would be possible to abstract over different evidence types (e.g. `Ordering[_]` vs `_ <:< Int`) but at least the `Int` case could be done with a simpler API. Do we have any other similar cases?

src/main/scala/strawman/collection/Sorted.scala
szeiger: Same here.

src/main/scala/strawman/collection/immutable/SortedSet.scala
szeiger: Couldn't this be resolved by making `SortedPolyTransforms` extend `IterablePolyTransforms`?

src/test/scala/strawman/collection/test/Test.scala
szeiger: This looks wrong. Scala 2.12 should be able to perform overload resolution correctly without a type annotation.

src/main/scala/strawman/collection/Set.scala
```diff
@@ -1,4 +1,27 @@
...
+import scala.Boolean
+
+/** Base trait for set collections */
+trait Set[A]
+  extends Iterable[A]
+    with SetLike[A, Set]
+
+/** Base trait for set operations */
+trait SetLike[A, +C[X] <: Set[X]]
+  extends IterableLike[A, C]
```
szeiger: Coming back to my argument for `StableIterable` from the other PR, we also inherit methods like `take` and `drop` from `Iterable` which only make sense with a stable ordering. Should we move them down to `StableIterable`? Or is this not important enough in practice and we should simply allow it for non-stable orderings, too (like in 2.12 and the current strawman design)?

src/main/scala/strawman/collection/Sorted.scala
julienrf: Maybe `ClassTag[_]`, depending on how we deal with arrays. I don’t understand what you mean with “the `Int` case”.

src/main/scala/strawman/collection/Set.scala
julienrf: Indeed.

src/main/scala/strawman/collection/immutable/SortedSet.scala
julienrf: Probably, but this is not what we want: we really want `SortedSet[A]` to extend `SortedPolyTransforms[A, SortedSet]` and `IterablePolyTransforms[A, Set]` (note that we have `Set` in the latter case).

We could define `SortedPolyTransforms` as follows, as a workaround:

~~~ scala
trait SortedPolyTransforms[A, +C[X] <: Sorted[X], +D[X] <: Iterable[X]]
  extends IterablePolyTransforms[A, D]
~~~

src/main/scala/strawman/collection/Set.scala
julienrf: I like this idea.

src/main/scala/strawman/collection/Sorted.scala
szeiger: I mean that you wouldn't want to overload `BitSet.map` as `def map[B](f: Int => B)(implicit ev: B <:< Int)` but as `def map(f: Int => Int)`

szeiger: But `ClassTag` could indeed be interesting. It would allow us to reuse a generic `PolyTransformsWithEvidence` for specialized collections.

julienrf: I created #28 to keep track of this issue.

src/main/scala/strawman/collection/immutable/SortedSet.scala
szeiger: I wouldn't call it a workaround, I think that's how it ought to be defined. You also have `SeqPolyTransforms <: IterablePolytransforms`. Enforcing the correct linearization makes overrides work in the desired order.

julienrf: I just pushed a commit that fixes the linearization order with a similar trick.

src/main/scala/strawman/collection/Set.scala
julienrf: I added a commit with this change.

src/test/scala/strawman/collection/test/Test.scala
julienrf: Where does the problem come from?

src/main/scala/strawman/collection/Set.scala
julienrf: I created issue #30 to keep track of this idea.

src/test/scala/strawman/collection/test/Test.scala
julienrf: @szeiger How could we investigate where does this issue come from?

szeiger: Simplify and run through `-Ytyper-debug`?

szeiger: Looks like a type inferencer bug. This is as far as I got: https://issues.scala-lang.org/browse/SI-10194

szeiger: https://github.com/scala/scala/pull/5708

julienrf:
Rebased to master and added `TreeMap`.

src/test/scala/strawman/collection/test/Test.scala
```diff
@@ -288,6 +289,37 @@ class StrawmanTest {
...
   }
 
+  def sortedSets(xs: immutable.SortedSet[Int]): Unit = {
+    val xs1 = xs.map((x: Int) => x.toString)
+    val xs2: immutable.SortedSet[String] = xs1
+  }
+
+  def mapOps(xs: Map[Int, String]): Unit = {
+    val xs1 = xs.map((k, v) => (v, k))
+    val xs2: strawman.collection.Map[String, Int] = xs1
```
julienrf: This example demonstrates that calling `map` with a function that returns a tuple returns a `Map`.

src/test/scala/strawman/collection/test/Test.scala
```diff
@@ -288,6 +289,37 @@ class StrawmanTest {
...
+    val xs3 = xs.map(kv => (kv._2, kv._1))
+    val xs4: strawman.collection.Iterable[(String, Int)] = xs3
```
julienrf: Note that here the distinction between the `map` overload that returns a `Map` and the one that returns an `Iterable` is based on the arity of the supplied function.

This is because `map` method of `Map` is defined as follows:

~~~ scala
def map[K2, V2](f: (K, V) => (K2, V2)): C[K2, V2]
~~~

We could alternatively define it as follows:

~~~ scala
def map[K2, V2](f: ((K, V)) => (K2, V2)): C[K2, V2]
~~~

Note that here,  `f` takes one parameter, which is a tuple, where in the first snippet, it takes two parameters.

I’m not sure which solution is better… I think both can work and make sense.

With the binary function I like the ability to just write `kvs.map((k, v) => …)` without having to write `{ case (k, v) => … }`. But note that dotty already makes it possible to write `kvs.map((k, v) => …)` even though `f` would have type `((K, V)) => (K2, V2)`. Maybe we could put this feature into scalac too…

Not sure I’m clear… Do you have any strong preference and why?

src/test/scala/strawman/collection/test/Test.scala
```diff
@@ -288,6 +289,37 @@ class StrawmanTest {
...
+
+  def sortedMaps(xs: immutable.SortedMap[String, Int]): Unit = {
+    val x1 = xs.get("foo")
+    val x2: Option[Int] = x1
+    val xs1 = xs + ("foo", 1)
+    val xs2: immutable.SortedMap[String, Int] = xs1
+    val xs3 = xs.map(kv => kv._1)
+    val xs4: immutable.Iterable[String] = xs3
+    val xs5 = xs.map((k: String, v: Int) => (v, k))
+    val xs6: immutable.SortedMap[Int, String] = xs5
```
julienrf: Note that here we have to explicit the types of the parameters `k` and `v`. I think this type inference issue is already fixed by https://github.com/scala/scala/pull/5708.

src/test/scala/strawman/collection/test/Test.scala
```diff
@@ -288,6 +289,37 @@ class StrawmanTest {
...
+    class Foo
+//    val xs7 = xs.map((k: String, v: Int) => (new Foo, v)) Error: No implicit Ordering defined for Foo
+    val xs7 = (xs: immutable.Map[String, Int]).map((k, v) => (new Foo, v))
+    val xs8: immutable.Map[Foo, Int] = xs7
```
julienrf: The commented line fails as expected because there is no implicit ordering for `Foo`. (Note that, alternatively, the overloading resolution could have chosen to use the `map` method of `Map`, which does not require an additional `Ordering`, but that’s not the case and I think that’s better like that)

However, if we upcast the `OrderedMap` to a simple `Map`, then the overload that is picked is the one that requires no `Ordering` and returns a `Map`.

julienrf:
This PR is ready for review :)

If you approve the design, I’ll start replacing the `???` by the actual implementations.

src/test/scala/strawman/collection/test/Test.scala
szeiger: The binary function is potentially more efficient because it avoids tupling. Does the untupling in Dotty also work for overloaded functions? I thought it was done during type adaptation after overload resolution. If we can get https://github.com/scala/scala/pull/5698 into Scala to make the case-syntax work here we will need a Tuple1 version but I still think we should have the Tuple2 version in addition.

src/test/scala/strawman/collection/test/Test.scala
szeiger: I've been thinking about Rex's CanBuildFrom idea and there are a few things I'd like to try here. I think it may be possible to find a solution that is close to CanBuild without the implementation complexity of CanBuildFrom, that can provide a fallback to the unconstrained collection type when the evidence is not available.

julienrf: In that specific case, though, I would expect a failure if the evidence is not available. When do you think you could push your experiment? For reference, I tried something here but it’s probably different from what you are thinking about: https://github.com/scala/collection-strawman/compare/poly-transforms#diff-e325090939d0c05b24284230016c77e7R189

Ichoran: This pattern works when classes don't have type parameters of their own:
```scala
trait Foo; trait Bar; trait TypeClass[C] {}
class Super {
  def foo[A](f: () => A): Foo = ???
}
class Sub extends Super {
  def foo[A: TypeClass](f: () => A): Bar = ???
}
```
If that pattern would work for collections it would be adequate to handle things like "this has to be a Tuple2 to stay a Map".  Right now you get "missing parameter type" on `(new Sub[String]).foo(_ + ".")` if you give the classes type parameters.  There's no real reason that has to be the case, though, AFAICT.

Ichoran: Anyway, that would be my proposal for the "less complex than CanBuildFrom", @szeiger. As long as the pattern works we can put in whatever, but I think a simple witness that the arguments are the correct type will be sufficient for most use cases (i.e. not replacing breakOut, but everything else).

julienrf: I don’t understand where you want to go. I’d like this strawman to move forward and address the remaining challenges (e.g. `BitSet`, `Future.sequence`, etc.). If you think the design proposed in this PR is wrong would you mind creating another PR with an alternative design?

src/test/scala/strawman/collection/test/Test.scala
odersky: Dotty can compile this PR as-is. It would also have no probblem with the Function1 version. In fact, it would let you transparently use binary closures such as "(k, v) => ..." as unary function arguments, expanding them to `{ case (k, v) => ... "`. So as far as dotty is concerned the `Function1` version is better since it can make the `Function2` variant redundant.

src/test/scala/strawman/collection/test/Test.scala
odersky: The strawman LGTM so far. I propose let's go ahead and flesh it out. I did not quite get the simplified CBF idea. The example by @Ichoran looks good, but it seems to be exactly what is done now. After all, `TreeSet`'s `map`

    def map[B](f: (A) => B)(implicit ordering: Ordering[B]): TreeSet[B] = ???

is exactly the same as

    def map[B: Ordering](f: (A) => B): TreeSet[B] = ???

which looks like the pattern that @ichoran was proposing. Or am I missing something here?

szeiger: Let's go ahead with this version. I'll start working in my idea now, but it's only a generalization of what we're already doing, so it doesn't have to hold up any work on this PR.

Ichoran: @odersky - The type inference doesn't currently work the way it probably ought to.  It picks the most specific version and runs with it, then gives you an error if it doesn't work.  So, in my toy example,
```scala
trait Foo; trait Bar; trait TypeClass[C] {}
class Super[+A] {
  def foo[B](f: A => B): Foo = ???
}
class Sub[+A] extends Super[A] {
  def foo[B: TypeClass](f: A => B): Bar = ???
}
```
you can't actually get a `Foo` when your type is `Sub` unless you ask for it:
```scala
scala> val s = new Sub[String]
s: Sub[String] = Sub@22495dd1

scala> s.foo(_ => "")
<console>:14: error: could not find implicit value for evidence parameter of type TypeClass[String]
       s.foo(_ => "")
            ^

scala> s.foo(_ => ""): Foo
scala.NotImplementedError: an implementation is missing
```
Maybe this is what we want to avoid accidentally selecting the wrong types, but it doesn't feel like `TreeMap` is a `Map` if you can't throw whichever function you want at it.

Ichoran:
I haven't formally reviewed it but it looks promising enough to me so that with Stefan's blessing I'm happy moving ahead.  The type inference tweaks I'm suggesting wouldn't change the overall strategy much, just deliver a different user experience, so that's no reason to wait or change designs.

src/test/scala/strawman/collection/test/Test.scala
julienrf: @Ichoran Another solution consists in pre-upcasting to `Super`: `(s: Super[String]).foo(_ => "")`.

In my opinion it is fine to get a failure if I want to `map` a `TreeSet` into something that has no `Ordering` instance, or if I try to `map` an `Array` into something that has no `ClassTag` available. I prefer to have to explicitly upcast my `TreeSet` into a `Set`.

Ichoran: When type signatures are complex, explicit upcasting is a large burden.

Another solution is to add an `asSuper` method that does nothing but the upcast.  (E.g. for `TreeMap` it would be `asMap`.)

PR #26, opened 2017-02-09
Maps

julienrf:
I did a quick experiment to see if overloading `map` methods for `Map` collections would work.

### DIFF
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -33,7 +33,7 @@ trait IterableLike[+A, +C[X] <: Iterable[X]]
   /** Create a collection of type `C[A]` from the elements of `coll`, which has
     *  the same element type as this collection. Overridden in StringOps and ArrayOps.
     */
-  protected[this] def fromIterableWithSameElemType(coll: Iterable[A]): C[A] = fromIterable(coll)
+  protected[this] def fromIterableWithSameElemType(coll: Iterable[A]): C[A]
 }
 
 /** Base trait for instances that can construct a collection from an iterable */
--- a/src/main/scala/strawman/collection/Map.scala
+++ b/src/main/scala/strawman/collection/Map.scala
@@ -1,0 +1,44 @@
+package strawman.collection
+
+import strawman.collection.mutable.Builder
+
+import scala.Option
+import scala.annotation.unchecked.uncheckedVariance
+import scala.Predef.???
+
+/** Base Map type */
+trait Map[K, +V]
+  extends Iterable[(K, V)]
+    with MapLike[K, V, Map]
+
+/** Base Map implementation type */
+trait MapLike[K, +V, +C[X, Y] <: Map[X, Y]]
+  extends IterableLike[(K, V), Iterable]
+    with IterableMonoTransforms[(K, V), C[K, V @uncheckedVariance]]
+    with MapPolyTransforms[K, V, C] {
+
+  def get(key: K): Option[V]
+
+}
+
+/** Polymorphic transformation methods */
+trait MapPolyTransforms[K, +V, +C[X, Y] <: Map[X, Y]] extends IterablePolyTransforms[(K, V), Iterable] {
+
+  def map[K2, V2](f: (K, V) => (K2, V2)): C[K2, V2]
+
+  def flatMap[K2, V2](f: (K, V) => IterableOnce[(K2, V2)]): C[K2, V2]
+
+}
+
+/** Factory methods for collections of kind `* −> * -> *` */
+trait MapFactories[C[_, _]] {
+
+  def newBuilder[K, V]: Builder[(K, V), C[K, V]]
+
+  def empty[K, V]: C[K, V] =
+    newBuilder[K, V].result
+
+  def apply[K, V](elems: (K, V)*): C[K, V] =
+    newBuilder[K, V].++=(elems.toStrawman).result
+
+}
--- a/src/test/scala/strawman/collection/test/Test.scala
+++ b/src/test/scala/strawman/collection/test/Test.scala
@@ -130,7 +130,7 @@ class StrawmanTest {
     val ys9: Seq[Int] = xs9
     val xs9a = xs.map(_.toUpper)
     val ys9a: String = xs9a
-    val xs10 = xs.flatMap((x: Char) => s"$x,$x")
+    val xs10 = xs.flatMap(x => s"$x,$x")
     val ys10: String = xs10
     val xs11 = xs ++ xs
     val ys11: String = xs11
@@ -286,6 +286,18 @@ class StrawmanTest {
     }
     println(xs17)
     println(xs17.to(List))
+  }
+
+  def mapOps(xs: strawman.collection.Map[Int, String]): Unit = {
+    val xs1 = xs.map((k, v) => (v, k))
+//    val xs1Bis = xs.map { case (k, v) => (v, k) } Does not compile :(
+    val xs2: strawman.collection.Map[String, Int] = xs1
+    val xs3 = xs.map(kv => (kv._2, kv._1))
+    val xs4: strawman.collection.Iterable[(String, Int)] = xs3
+    println(xs1)
+    println(xs2)
+    println(xs3)
+    println(xs4)
   }
 
   @Test
--- a/src/main/scala/strawman/collection/Seq.scala
+++ b/src/main/scala/strawman/collection/Seq.scala
@@ -50,8 +50,11 @@ trait IndexedSeq[+A] extends Seq[A] { self =>
 /** Base trait for Seq operations */
 trait SeqLike[+A, +C[X] <: Seq[X]]
   extends IterableLike[A, C]
-    with SeqMonoTransforms[A, C[A @uncheckedVariance]] // sound bcs of VarianceNote
+    with SeqMonoTransforms[A, C[A @uncheckedVariance]] { // sound bcs of VarianceNote
 
+  protected[this] def fromIterableWithSameElemType(coll: Iterable[A]): C[A] = fromIterable(coll)
+
+}
 /** Base trait for linear Seq operations */
 trait LinearSeqLike[+A, +C[X] <: LinearSeq[X]] extends SeqLike[A, C] {
 
--- a/src/main/scala/strawman/collection/View.scala
+++ b/src/main/scala/strawman/collection/View.scala
@@ -12,6 +12,9 @@ trait View[+A] extends Iterable[A] with IterableLike[A, View] {
     case c: View[B] => c
     case _ => View.fromIterator(c.iterator())
   }
+
+  protected[this] def fromIterableWithSameElemType(coll: Iterable[A]): View[A] = fromIterable(coll)
+
   override def className = "View"
 }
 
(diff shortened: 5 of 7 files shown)


src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -33,7 +33,7 @@ trait IterableLike[+A, +C[X] <: Iterable[X]]
   /** Create a collection of type `C[A]` from the elements of `coll`, which has
     *  the same element type as this collection. Overridden in StringOps and ArrayOps.
     */
-  protected[this] def fromIterableWithSameElemType(coll: Iterable[A]): C[A] = fromIterable(coll)
+  protected[this] def fromIterableWithSameElemType(coll: Iterable[A]): C[A]
```
julienrf: I removed this default implementation because in the case of `Map[K, V]`, `fromIterable` does not return a `Map[K, V]` but an `Iterable[(K, V)]`.

src/test/scala/strawman/collection/test/Test.scala
```diff
@@ -288,6 +288,15 @@ class StrawmanTest {
     println(xs17.to(List))
   }
 
+  def mapOps(xs: strawman.collection.Map[Int, String]): Unit = {
+    // val xs1 = xs.map { case (k, v) => (v, k) }
+    // val xs2: strawman.collection.Map[String, Int] = xs1
+    val xs1 = xs.map(kv => (kv._2, kv._1))
+    val xs2: strawman.collection.Iterable[(String, Int)] = xs1
```
julienrf: As you can see in the above lines:
- Using the `.map { case (k, v) => …` syntax breaks type inference ;
- The inferred type of `xs1` is `Iterable[(String, Int)]` instead of `Map[String, Int]`.

Ichoran: You haven't given it any way to get `Map[String, Int]`, have you?  The standard trick is something like
```
def map[C, U](f: ((K, V)) => ((C, U)))(implicit ev: ((C, U)) <:< Tuple2[_, _]): Map[C, U] = ???
```
or somehow get implicit resolution in there.  That still gives broken type inference, but type inference and implicit resolution could be tweaked so that even *without* inferring all the types, the compiler would know to try this one first (and then be able to get through the case without an issue).  I unfortunately don't know whether the internals can support this with a reasonable amount of work, but conceptually it seems like it would solve the issue.

julienrf: Oh, I just realized that I defined the `map` overload that returns a `Map` with the following signature:

~~~ scala
def map[K2, V2](f: (K, V) => (K2, V2)): Map[K2, V2]
~~~

So, it takes as parameter a function with two parameters, not a function with one (tuple) parameter. But at call site I wrote code that was supplying a function taking one tuple parameter. That’s why the type inference didn’t result to what I expected.

I changed the call site and now the overloading trick works.

julienrf:
I did a few fixes since the first push and now I have something that works as follows.

If you write:

~~~ scala
xs.map((k, v) => …)
~~~

Then it selects the `map` overload that returns a `Map`.

If you write:

~~~ scala
xs.map(kv => …)
~~~

Then it selects the `map` overload that returns an `Iterable`.

However, the following does not compile:

~~~ scala
xs.map { case (k, v) => … } // missing parameter type for expanded function
~~~

So, this solution works but is not compatible with existing usage of `Map`s.

Ichoran:
As I mentioned before, you wouldn't expect the `map(kv => ...)` version to work because it explicitly has the wrong type signature.  The normal way to get around this--adding an implicit parameter--also doesn't work because the return type is not, in general, inferred.  But it *could* work if the implicit parameter could interact with the type checker in selecting which flavor of `map` to try first.

julienrf:
@Ichoran I was not clear but I tried different type signatures and no ones worked excepted the one that takes a `Function2` as parameter.

This one didn’t work, for instance:

~~~ scala
def map[U, K2, V2](f: ((K, V)) => U)(implicit ev: U <:< ((K2, V2))): C[K2, V2]
~~~

odersky:
Note that in dotty we have an automatic conversion from a binary lambda to a lambda over pairs. We could think of adding this to Scala 1.13 as well, the change is pretty simple. But it seems that @julienrf found how to make it work without that conversion, which is better.

src/main/scala/strawman/collection/Iterable.scala
szeiger: Why should it? I see no particular reason why we'd want the binary type constructors in `fromIterableWithSameElemType`. This compiles: https://github.com/szeiger/collection-strawman/commit/0de8480fbd61a58c2b6b603827dbfb0b35b2789b

szeiger:
@odersky It doesn't work for PartialFunction lambdas. I'm not familiar with that part of scalac but it's not clear to me why it couldn't (even without full unification of tuples and parameter lists). The problem only occurs if you overload the existing `map(Function1)` with a `map(Function2)` but overload resolution should be smart enough to know that a lambda with `case` can only ever match the `Function1` version.

src/main/scala/strawman/collection/Iterable.scala
julienrf: Because we want `filter` to still return a `Map[K, V]`, right?

szeiger: Hm, I just ran into two different compiler bugs (one probably being https://issues.scala-lang.org/browse/SI-10081) while trying this out. We're pushing the limits here...

szeiger:
Here's a fix for the PartialFunction case: https://github.com/scala/scala/pull/5698

src/main/scala/strawman/collection/Map.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import strawman.collection.mutable.Builder
+
+import scala.Option
+import scala.annotation.unchecked.uncheckedVariance
+import scala.Predef.???
+
+/** Base Map type */
+trait Map[K, +V]
+  extends Iterable[(K, V)]
```
dwijnand: Is there any reason to use `Product2` instead of `Tuple2` here (and throughout)?

julienrf: I think I’m going to learn something here: I thought `(K, V)` was a syntactic shortcut for `Tuple2[K, V]`, is it?

dwijnand: It is. I mean why are you using Tuple2 instead of Product2? I think you could and should.

odersky: @dwijnand Why do you prefer `Product2` over `Tuple2`?

dwijnand: It's a design decision on specificity / principle of least power.

The subject here is Map, a top-level trait. Must we constrain its Iterable element type to be Tuple2 or should we allow it instead to return any type of Product2?

Does the correctness of the collections in this rewrite depend on the correctness of the implementation of equals and hashCode of the Iterable element type? If so then there's more safety in using Tuple2 (although not complete safety as Tuple2 isn't final so one could still _technically_ override and subvert its synthetic equals/hashCode).

An example of the benefits in using Product2 instead is that in one's own implementation of Map one could avoid a lot of unboxing and boxing _into_ Tuple2 by using an element type that implements Product2.

One's hand is slightly swayed (possibly unconsciously) by the fact that in Scala, at both the value and type level, the syntax (..) is tuple syntax, instead of product syntax. Therefore requiring the use of Product1/Product2/.. to be used for product syntax, which is more verbose.

Thank you for the question, it allowed me to think out my thoughts and present them in full.

sjrd: @dwijnand If `Map` iterates with `Product2`s, then my typical way of iterating over it is not safe any more:
```scala
for ((key, value) <- myMap) { ... }
```
The new way to iterate on the client side would be:
```scala
for (keyValue <- myMap) {
  val key = keyValue._1
  val value = keyValue._2
  ...
}
```
That's definitely a regression to me.

The principle of least power needs to be balanced with the principle of easiest use, sometimes.

dwijnand: > The principle of least power needs to be balanced with the principle of easiest use, sometimes.

I agree with you. But Product2 works just like you like it:

```
scala> val xs = Seq[Tuple2[Int, Int]]((1, 2))
xs: Seq[(Int, Int)] = List((1,2))

scala> val ys = Seq[Product2[Int, Int]]((1, 2))
ys: Seq[Product2[Int,Int]] = List((1,2))

scala> for ((key, value) <- xs) println(s"$key: $value")
1: 2

scala> for ((key, value) <- ys) println(s"$key: $value")
1: 2
```

DarkDimius: I agree with the Product approach, it allows for smarter memory management by not forcing to inherit __class__  tupple2.

sjrd: *Only* if there's actually a `Tuple2` behind it. But not otherwise:
```scala
scala> case class Foo(_1: Int, _2: Int) extends Product2[Int, Int]
defined class Foo

scala> val xs: Iterable[Product2[Int, Int]] = List(Foo(1, 2))
xs: Iterable[Product2[Int,Int]] = List(Foo(1,2))

scala> for ((key, value) <- xs) println(s"$key: $value")

```
That's right: it *does not even crash*! It filters out things that are not `Tuple2`s.

odersky: @sjrd Well spotted. So I think we stick to tuples, after all.

odersky: Also, it's hard to see how maps could profit from `Product2`. To gain any efficiency through that, they'd need to already have a Product2 of a key and a value stored somewhere, so they could return that. It's a possibility for mutable hashmaps. Under open hashing, entries could presumably be made into the right Product2 types. But we still don't want to hand out an entry directly, because that could cause a memory leak... So, no.

Regarding "principle of least power": It actually does the other way! Tuple2 is less powerful than Product2 because it constrains the implementation more. So "least power" and "ease of use" are aligned here, after all.

dwijnand: @sjrd You're right. You would need to use a Product2 extractor instead of the language Tuple2 extractor. Another example of the (..) tuple syntax swaying one's hand, yours in this case.

@odersky I think what is "least power" depends on whose side you're looking, the producer or the consumer of the type. Lots of times you're the only implementor, so you only have consumers. But in this case it's a collections library trait, so you have both. So it's a design decision/judgement call.

julienrf:
Superseeded by #24.

src/main/scala/strawman/collection/Map.scala
odersky: @dwijnand I agree, it's an interesting observation that the principle works both ways, and means opposite things depending on how you look at it. I think it would be good to find different names for each direction:

 - Principle of least power: Given some requirements, choose the most constraining ("obvious") API that fulfills the requirements, in order to maximize understanding what the implementation will do.

 - Principle of maximal abstraction: Given some requirements, choose the API that admits the largest set of possible implementations. 

Neither principle is inherently better than the other, but I believe that currently many libraries go too much towards maximal abstraction.

dwijnand: There we go. Nice, I like it.

PR #29, opened 2017-02-14
Range and NumericRange

julienrf:
This PR is a rebased version of #11.

I also tried to make `Range` extend `NumericRange[Int]` but the compiler was crashing with a StackOverflow when I wanted to override the `by(newStep: T): NumericRange[T]` method for range (`by(newStep: Int): Range`).

I `@specialized(Unit)` the `foreach` method in both `Range` and `NumericRange` but this didn’t seem to have any measurable effect on `NumericRange`.

### DIFF
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -3,8 +3,11 @@ package collection
 
 import scala.annotation.unchecked.uncheckedVariance
 import scala.reflect.ClassTag
-import scala.{Int, Boolean, Array, Any, Unit, StringContext}
-import java.lang.{String, UnsupportedOperationException}
+import scala.{Any, Array, Boolean, Int, StringContext, Unit, UnsupportedOperationException}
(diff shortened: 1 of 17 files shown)


src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -145,6 +156,14 @@ trait IterableOps[+A] extends Any {
   *  as the one they are invoked on.
   */
 trait IterableMonoTransforms[+A, +Repr] extends Any {
+  def filter(p: A => Boolean): Repr
```
DarkDimius: This operation shouldn't be here. If you filter a range, it may stop being a range.

julienrf: The way we solved that is to make `Range` extend `IterableMonoTransforms[Int, Seq[Int]]` and then covariantly override the return type of the other methods of `IterableMonoTransforms` (e.g. `take`).

julienrf:
Superseded by #102

PR #31, opened 2017-02-15
Immutable arrays

julienrf:
This is a rebased version of #6 with benchmarks.

### DIFF
--- a/benchmarks/memory/src/main/scala/strawman/collection/MemoryFootprint.scala
+++ b/benchmarks/memory/src/main/scala/strawman/collection/MemoryFootprint.scala
@@ -1,4 +1,4 @@
-package bench
+package strawman.collection
 
 import strawman.collection.immutable.{LazyList, List}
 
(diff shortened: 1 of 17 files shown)


julienrf:
Here are some charts generated by the benchmarks:

lookup

Random access can be up to 2x faster than `Vector` (and an order of magnitude faster than `List`, of course).

concat

Concatenation can be up to 10x faster than `Vector` and `List`.

map

`map` can be up to 3x faster than `Vector` and 5x faster than `List`.

uncons

`tail` is an order of magnitude slower than `Vector` or `List`.

julienrf:
@odersky @Ichoran @szeiger There is a non anecdotal design decision behind this PR: generic collection types (e.g. `Iterable`, `Seq`, `Set`, etc.) have no implemented operations. The idea is to avoid inheriting inefficient operations. Concrete collection types can still mix in “default” implementations by mixing in traits like `IterableMonoTransformsFromIterable` (which implements operations of `IterableMonoTransforms` on top of a `fromIterable(it: Iterable[A]): Repr` method).

Do you like this idea?

Ichoran:
I like the idea of having a way to avoid pulling in inefficient methods, so overall yes.  I'm not sure the number of methods should be literally _zero_ since there is sometimes only one right way to do things.  On the other hand, if the set of implementations is not very granular, it doesn't matter because usually you will want _something_ and so it's tempting to just pull in everything.  This is harder to notice at this stage of the strawman because the set of methods is not very complete.

szeiger:
It would be helpful to take a step back and look for who overwrites what in the current collection implementations. `fromIterable` may still end up being the best default, or at least the only generic implementation that can be reused by multiple collection types.

julienrf:
Superseeded by #87

PR #40, opened 2017-03-02
Attempt at basic NonEmptyLists

joshlemer:
* Overrides many methods in `::[A]` and `List[A]` to keep type information when doing size-preserving transformations
* Add a `NonEmptyList` factory, which cannot conform to `IterableFactories[NonEmptyList]` because it is impossible to construct a `Builder[A, NonEmptyList[A]]`, as that trait requires a `clear()` method, returning the underlying collection to an empty state

Really hopeful that some form of NonEmpty collections can make it into the next redesign in one way or an other :)

### DIFF
--- a/src/main/scala/strawman/collection/immutable/List.scala
+++ b/src/main/scala/strawman/collection/immutable/List.scala
@@ -4,8 +4,9 @@ import scala.annotation.unchecked.uncheckedVariance
 import scala.Nothing
 import scala.Predef.???
 import strawman.collection
-import strawman.collection.{IterableFactories, IterableOnce, LinearSeq, SeqLike}
+import strawman.collection.{IterableFactories, IterableOnce, LinearSeq, SeqLike, View}
 import strawman.collection.mutable.{Buildable, Builder, ListBuffer}
+import scala.{Unit, Option, Int, None, Some}
 
 
 /** Concrete collection type: List */
@@ -20,7 +21,7 @@ sealed trait List[+A]
   protected[this] def newBuilder = List.newBuilder
 
   /** Prepend element */
-  def :: [B >: A](elem: B): List[B] =  new ::(elem, this)
+  def :: [B >: A](elem: B): ::[B] =  new ::(elem, this)
 
   /** Prepend operation that avoids copying this list */
   def ++:[B >: A](prefix: List[B]): List[B] =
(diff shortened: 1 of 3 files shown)


odersky:
I fear this would negatively impact type inference and/or give confusing error messages. Also, why should non-empty lists be treated specially? Why not also have lists of length greater to some constant, or lists of length equal to some constant? It seems non-empty lists scratch only the surface here.

joshlemer:
@odersky you have a good argument, just a couple points:  

1) I do think the distinction between (length == 0) and (length > 0) is not really arbitrary. There are many methods on `List[A]` that for example throw exceptions when called on `Nil`:  
* head
* tail
* last
* init
* min / minBy / max / maxBy  
  
which means that when using `List` in practice, code is often obfuscated in `.headOption.map...`, pattern matches, or `Try(list.max).toOption`, and I think you end up with lots of error handling code because of the lack of NonEmptyList. Just my opinion though
  
2) 
>Why not also have lists of length greater to some constant, or lists of length equal to some constant?  
That would be cool! Sort of like an HList maybe?
```
case class :: [+A, +Tail <: List[A]](x: A, private[collection] var next: Tail @uncheckedVariance)

val atLeastTwo: Int :: Int :: List[Int] = 1 :: 2 :: 3 :: Nil
```

Cheers!

ritschwumm:
@odersky allow me a rethorical question: why treat lists of 0..1 entries specially?

 both Option and non-empty collections are things that occur quite frequently in day to day programming, or at least for me they do. weren't there places in scala macros where the AST used (A,List[A]) instead?

odersky:
The difference is that `Option` is a solid abstraction. A non-empty list abstraction is destroyed the moment you take `tail` because that loses the knowledge of nonemptyness.

ritschwumm:
i don't see how this would be "not solid" - of course taking the tail of a non-empty collection is partial, just like taking the head of a regular (possibly-empty) collection.
same as List has headOption, NonEmptyList would have tailOption - or tailList if you want.

odersky:
The original proposal was to retrofit non-emptyness on existing lists, which IMO will never work. Exploring a separate abstraction for non-empty lists with all the necessary methods - sure, why not? But such an abstraction should be developed and widely user-tested before we would consider putting this in the strawman.

Ichoran:
I think it is valuable to have a trial implementation in the collections strawman as a branch because it's another kind of use-case that is reasonably common (there are a number of NonEmptyList-like implementations for the current Scala collections).  If the strawman does not support it as well as the current collections do, we may wish to think if we can tweak the design to support it better even if we don't include it.  Because there are several battle-tested NonEmptyLists out there (in scalaz, cats, and scalautils (except that actually goes out to two+ elements with "many"), among others).

I agree that we likely don't want it as part of regular List because with List it is convenient for type assignment that `x :: Nil` has type `List` not type `::`, and since this is how it was done historically, changing that would cause a lot of hassles.  But making NonEmptyLists a lot harder to implement would also cause hassles.

edmundnoble:
It is "the point" (or rather, the `copoint` :wink:) of `NonEmptyList` to be "destroyed the moment you take head"; it's a `Comonad`. Extending the `NonEmptyList` context to another function that requires it is as easy as `nel.extend(f)` (i.e., `nel.tails.map(f)`).

However `::` does not at all make for a readable `NonEmptyList` type name.

Issue #42, opened 2017-03-07
A more sensible `groupBy` API?


LPTK:
This is the result of some personal experimentations. Please let me know if it's not appropriate to post that here.

## Introduction

The `groupBy` method is very powerful and pretty useful, but it often suffers from inefficiencies because of the way its API is specified:
 
 * It is supposed to return an immutable `Map`. While sometimes useful, it's often not what one needs. Many uses of `groupBy` end up mapping or simply iterating on the result, in which case having a `Map` early is counter-productive.
 Notice that the [standard implementation](https://github.com/scala/scala/blob/44971d104f4364a0ddaa4f05afc4cc61ee39cdf7/src/library/scala/collection/TraversableLike.scala#L318-L330) of `groupBy` constructs a `mutable.Map` _and then_ rebuilds it into an `immutable.Map`, a final step that is often redundant.
 **Proposal:** Make the method return an `Iterable` of the same type as the current collection, and let the user decide whether to call `toMap` on it or not. One can also write something like `xs.view.groupBy(f)` which returns an immutable view into the internal `mutable.Map` used by `groupBy`, avoiding any unnecessary intermediate container (note that the call to `groupBy` is still strict nonetheless).
 
 
 * A very common use-case of `groupBy` involves transforming the values after having grouped them. Doing this as a second step as in `xs.groupBy(_.a).mapValues(_.map(_.b))` is wasteful as it creates needless intermediate collections.
 **Proposal:** Have an optional parameter to specify the value mapping to perform while grouping values. As long as this parameter has a default value, it should not even break the current API, as both `groupBy(f)` and `groupBy(f,g)` would be valid.
 Note: after implementing this, I discovered that user @balefrost [had already mentioned the problem](https://github.com/lampepfl/dotty/issues/818#issuecomment-157086673), and that it turns out that's how it's already done in C#.
  
 * With the new design based on the `Iterable` hierarchy, it is currently not possible to directly build the buckets of the grouped elements because there is no way to obtain a builder of the current collection. Although [it should be possible eventually](https://gitter.im/scala/collection-strawman?at=58b96c06e961e53c7f942c71), I'm not even sure it makes much sense to have the buckets mirror the original collection type anyway. In this proposal `groupBy` simply returns buckets as `View[V]`.
 
## Implementation

I have implemented the traditional `groupBy` API (but it's currently suboptimal because of the lack of `newBuilder`), along with my proposal for a modified API called `viewBy`, the signature of which is:

```scala
[...]

PR #43, opened 2017-03-07
Implement method `groupBy` and alternative version `viewBy`

LPTK:
See issue #42.

### DIFF
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -168,6 +168,22 @@ trait IterableMonoTransforms[+A, +Repr] extends Any {
     if (coll.isEmpty) throw new UnsupportedOperationException
     drop(1)
   }
+
+  /** Traditional GroupBy */
+  def groupBy[K](f: A => K): scala.collection.immutable.Map[K, Repr] = {
+    import mutable.ArrayBuffer
+    val m = scala.collection.mutable.Map.empty[K, ArrayBuffer[A]]
+    for (elem <- coll) {
+      val key = f(elem)
+      val bldr = m.getOrElseUpdate(key, new ArrayBuffer)
+      bldr += elem
+    }
+    val b = scala.collection.immutable.Map.newBuilder[K, Repr]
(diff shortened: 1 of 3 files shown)


julienrf:
Hi @LPTK!

Sorry for the delay, I was busy on other things, but `groupBy` is one of the next things to do, now that `Map`s are implemented (well, we still miss mutable `Map`s), so I think it’s time to review this PR.

The numbers you posted in #42 are nice but there are several differences between `viewBy` and `groupBy`. I think that would be interesting to know the performance improvement brought by each change.

Also, in the discussion of #42, we mentioned another variant that performs an aggregation operation on each group. I think it would be nice to include it. And we should try your suggestion that consists of adding a new parameter to `groupBy` but with a default value to not break source compatibility.

Finally, there are still some discussion points:

- What should be the type of the collection returned by `groupBy`? A `Map[K, ?]`, an `Iterable[(K, ?)]` or the same collection type? I don’t think returning the same collection type would make sense, especially with ordered collections, I have a slight preference for returning `Map`, though I would be interested in knowing precisely if it has a cost over returning an `Iterable`.
- What should be the type of the groups? Here I think it makes sense to keep the same collection type (again, think of ordered collections).

So, my preference would be to have something like so:

~~~ scala
trait IterableOps[+A] {
  def groupAndAggregate[K, V](key: A => K, value: A => V, aggregate: (V, V) => V): Map[K, V]
}

trait IterablePolyTransforms[+A, +C[X]] {
  def groupBy[K, V](key: A => K, value: A => V = a => a): Map[K, C[V]]
}
~~~

What do you think? Would you be interested in working on that, @LPTK?

Ichoran:
`groupBy`'s use-case signature should not change.  If we want something similar that retains the original collection type we should give it a new name (e.g. `mapGroups(key: A => K)(f: (K, CC[A]) => B): CC[B]`).

julienrf:
@LPTK What’s the status of this PR?

LPTK:
@julienrf I think this was more or less superseded by https://github.com/scala/collection-strawman/pull/253, so I'm closing.

PR #45, opened 2017-03-08
Abstractions for constrained collection types

szeiger:
This PR takes the result of https://github.com/scala/collection-strawman/pull/39 and the discussion there to implement the necessary abstractions properly. This integrates the `() => Builder` implicits that were added to master in the meantime and unifies these builder factories with type-constrained/unconstrained collection factories and with monomorphic collection factories that do not need hacks like `BitSet with Iterable[E]` anymore.

- Add an upper bound in `FromIterable` and `IterableFactory`. This
  allows us to express type-constrained iterables like `BitSet` without
  requiring an implicit evidence.

- Drop the `C[X] <: Iterable[X]` constraint in `FromIterable`.
  Guaranteeing an `Iterable[X]` result is not necessary anywhere and
  it makes monomorphic collection types awkward, for example producing
  a `BitSet with Iterable[X]` for any `X <: Int`. What we really want
  is to be able to drop the precise element type in `C[X] = BitSet`.

- Add `FromIterableBase` which contains the target type constructor as
  a type member. It is extended by `FromIterable` which pulls the
  type constructor out into a type parameter. This allows us to write
  polymorphic code for arbitrary `FromIterableBase` subtypes without
  having to infer the type constructor (which is not possible when it
  is not a real type constructor but a type lambda computation like
  `C[X] = BitSet` via `MonomorphicIterableFactory`.

- Add `ConstrainedFromIterable` and `ConstrainedIterableFactory` to
  mirror `FromIterable` and `IterableFactory` but with an additional
  implicit evidence for the element type. An upper bound and a
  monomorphic version are not necessary: Constrained collection types
  are always polymorphic (otherwise they would already have the
  correct evidence value for the element type) and any type bound can
  be rolled into the evidence type.

- Overload `to` for `FromIterable` and `ConstrainedFromIterable`.

- Add `ConstrainedIterablePolyTransforms` which mirrors
  `IterablePolyTransforms` with an additional implicit evidence (so it
  is based on `ConstrainedFromIterable` instead of `FromIterable`). It
  provides an additional method `unconstrained` to perform a widening
  conversion to the most specific unconstrained collection type.

- Replace the implicit `() => Builder` with a more generic implicit
  `MonomorphicIterableFactory`. It works via implicit lookup for all
  unconstrained, type-constrained and value-constrained unary type
  constructors.

- Add `ConstrainedMapFactory` which mirrors `MapFactory` with an
  additional implicit evidence for the key type. I considered using
  a binary type constructor that allows the evidence to be based on
  both, key and value types but it would complicate the API with a
  projection type for the key-only constraints and I do not expect that
  we will need it for any collection.

- Implement a builder-based `fromIterable` in `MapFactory` and
  `ConstrainedMapFactory` and add implicit `MonomorphicIterableFactory`
  implementations to both.


src/main/scala/strawman/collection/IterableFactories.scala
```diff
@@ -4,23 +4,61 @@ package collection
...
+    def fromIterable[E <: B](it: Iterable[E]): C[B] = self.fromIterable[B](it)
+    def newBuilder[A <: B]: Builder[A, C[B]] = self.newBuilder[B]
+    override protected[this] lazy val monomorphicIterableFactoryProto = this
+  }
+
+  implicit def iterableFactory[E <: B]: MonomorphicIterableFactory[E, C[E]] =
+    monomorphicIterableFactoryProto.asInstanceOf[MonomorphicIterableFactory[E, C[E]]]
+}
+
+trait MonomorphicIterableFactory[-B, +Repr] extends IterableFactory[B, ({ type L[X] = Repr })#L]
```
szeiger: `MonomorphicIterableFactory` makes sense from the POV of collection implementations. Calling it `CanBuild` would better convey the meaning to users.

src/main/scala/strawman/collection/immutable/TreeSet.scala
```diff
@@ -32,14 +42,11 @@ final class TreeSet[A]()(implicit val ordering: Ordering[A])
 
   // From SortedLike
   def range(from: A, until: A): TreeSet[A] = ???
-
-  // From SortedPolyTransforms
-  def map[B](f: (A) => B)(implicit ordering: Ordering[B]): TreeSet[B] = ???
```
szeiger: `ConstrainedIterablePolyTransforms` extends `IterablePolyTransforms` in order to get the linearization right for the purpose of overload resolution, so this override is no longer required.

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -29,14 +30,21 @@ class TraverseTest {
   def traverseTest: Unit = {
 
     val xs1 = immutable.List(Some(1), None, Some(2))
-    optionSequence(xs1)(immutable.List.canBuild[Int]) // TODO Remove explicit CanBuild parameter after https://issues.scala-lang.org/browse/SI-10081 is fixed?
```
szeiger: Would still require an explicit parameter even with SI-10081 fixed

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -29,14 +30,21 @@ class TraverseTest {
...
+    val o1 = optionSequence(xs1)
+    val o1t: Option[immutable.List[Int]] = o1
 
     val xs2 = immutable.TreeSet(Some("foo"), Some("bar"), None)
-    optionSequence(xs2)(immutable.TreeSet.canBuild[String])
+    val o2 = optionSequence(xs2)
```
szeiger: No explicit factory required, even for constrained collections. Both the previous implementation here and my old attempt in https://github.com/scala/collection-strawman/pull/39 needed it.

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -29,14 +30,21 @@ class TraverseTest {
...
+    val o2t: Option[immutable.TreeSet[String]] = o2
 
     val xs3 = mutable.ListBuffer(Right("foo"), Left(0), Right("bar"))
-    eitherSequence(xs3)(mutable.ListBuffer.canBuild[String])
+    val e1 = eitherSequence(xs3)
+    val e1t: Either[Int, mutable.ListBuffer[String]] = e1
 
+    // This use case from https://github.com/scala/scala/pull/5233 still eludes us
```
szeiger: I'm considering to resurrect the idea of path-dependent types for `CanBuild` scoping, essentially unifying the new way I implement `to` (which doesn't need to infer a unary type constructor anymore) with the implicit builder factories. This would not enable a type-driven `breakOut` but a value-driven approach with an explicit factory similar to the syntax below (like in `to`).

julienrf: Indeed, I agree that we should try something like this.

szeiger:
I tried this trick for a `CanBuildFrom`-like abstraction:

```scala
trait Iterable[+A] extends IterableOnce[A] with IterableLike[A, Iterable] {
  final type Build[-B, +Repr] = MonomorphicIterableFactory[B, Repr]
  ...
}

// Testing:
def pair[A, Repr <: Iterable[A], To](xs: Repr with Iterable[A])(implicit fi: Repr#Build[(A, A), To]): To = ???

val xs5 = immutable.List(1, 2, 3)
val p1 = pair(xs5)
val p1t: immutable.List[(Int, Int)] = p1
```

My understanding of the implicit scopes in 7.2 (http://www.scala-lang.org/files/archive/spec/2.11/07-implicits.html) is that `Repr#Build[(A, A), To]` first puts the companion object of `Repr` into the implicit scope and then dealiases `Build` but scalac disagrees. The `Repr` companion is not part of the scope. Am I misunderstanding the spec or is this a bug?

Ichoran:
You know, I'm getting worried that this level of abstraction is no better than and possibly even worse than CanBuildFrom.  ConstrainedIterablePolyTransforms is particularly bad--I can't imagine that anyone who had trouble with CBF would find CIPT simple.  (And the various self-types and factories are not terribly transparent either, I don't think, to a less expert user.)

I think if this rewrite is going to be a success on the simplification dimension, we're going to have to rely less on clever and elaborate type encodings and more on the compiler to do the thing we want by default.  If this means altering the scope or search strategy of implicits, I think that's fine.  If we drop the simplification part--that is, understanding how collections work remains a dark art of the high wizards of Scala--then the existing design I think is quite good.  And I'm not totally opposed to dropping that part; having better behavior and better implementation improves things quite nicely.  But perhaps we do want the simplification, and then I think this is getting to be too weighty.

In particular, I would take a completely orthogonal approach to constraints: a constrained collection does *not even live in the same hierarchy*, but requires boxing to become a full-fledged collection member.  We're partway there with ConstrainedIterablePolyTransforms, and I think if you don't care to understand the four type parameters very carefully that using CIPT is actually quite straightforward.  But having coexisting constrained and unconstrained methods makes things especially tricky to implement.

I'll try to come up with a sketch this weekend where the hierarchy is held together more loosely, and where to get the full power of collections you may need to box into them.  It's more of a breaking change, so maybe it's not going to work.  But this experiment is convincing me that you can't really get CBF-like functionality without CBF-like complexity.

lrytz:
In its defense, CIPT doesn't almost leak to the parts of the collections hierarchy which have nothing to do with constrained collections. This is very different from CBF, which are everywhere.

It does add complexity related to overriding order and overloading resolution. But this looks good to me:
image

Compared to the 2.12 version
image

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -189,3 +193,27 @@ trait IterablePolyTransforms[+A, +C[A]] extends Any {
   def zip[B](xs: IterableOnce[B]): C[(A @uncheckedVariance, B)] = fromIterable(View.Zip(coll, xs))
   // sound bcs of VarianceNote
 }
+
+/** Transforms over iterables that can return collections of different element types for which an
+  * implicit evidence is required.
+  */
+trait ConstrainedIterablePolyTransforms[+A, +C[A], +CC[X] <: C[X], Ev[A]] extends ConstrainedFromIterable[CC, Ev] with IterablePolyTransforms[A, C] {
```
lrytz: `+C[A]` could be `+C[_]` (also in `IterablePolyTransforms`)

szeiger:
The simplification vs `CanBuildFrom` is two-fold:

1. You only see the constraint-related complexity in the collection types that need them (i.e. arrays and sorted collections).

2. Compiler errors and auto-complete results look the way you expect despite abstracting over the evidence type because you're always dealing with a concrete instantiation where the evidence type is known.

For users I think this is vastly simpler than `CanBuildFrom`. For collection implementors it makes things somewhat easier because they can reuse `ConstrainedIterablePolyTransforms` for new evidence types. I agree that implementing collections is still difficult but that seems to be inherent complexity that we cannot remove. Either we abstract over possible constraints everywhere (`CanBuildFrom`) or we have two versions of all poly-transforms for constrained and unconstrained collections.

Note that adding the unconstrained collection type constructor as another new type parameter is independent of abstracting over the evidence type. `SortedPolyTransforms` should also have done it so it can extend `IterablePolyTransforms`, otherwise you need to override all methods in all implementations to get the linearization right.

> But this experiment is convincing me that you can't really get CBF-like functionality without CBF-like complexity.

We're already sacrificing functionality in this design: There's no automatic fallback to building an unconstrained collection and you can't implement a `breakOut` equivalent. Maybe more importantly, while we can implement methods like `Future.sequence` with the current level of functionality we cannot implement the improved version from https://github.com/scala/scala/pull/5233, which is what I was trying to do with the `Repr#Build` trick above. This requires something similar to `breakOut`, not because we want to build a target type that is different from the source type but because the target type may require a binary type constructor.

src/main/scala/strawman/collection/Iterable.scala
szeiger: I already removed a bunch of unnecessary names. I must have overlooked these ones.

Ichoran:
@lrytz - Yes, I like the improvement of usage you get from CIPT, as I alluded to earlier!  And I do like that it's kept away from many of the collections.

@szeiger - Actually I view the lack of automatic fallback an advantage, and I'm not saying that we shouldn't have constrained and unconstrained collections inherit different things.  I'm actually wondering whether we can additionally simplify things by allowing the inheritance to vary even *more*.  I'm reasonably happy with the existing design.  I'm just looking for additional ways to simplify.

For instance, with CIPT we have a fourth type parameter, C[_] or C[A], that exists solely to implement a single widening method.  Nothing in CIPT cares (yet) that this method exists, but it complicates the type signature of the trait (both by having it and by requiring CC[X] <: C[X]).  Can we just drop it?  Or do we need it as a bridge to the less constrained methods?  If we need it as a bridge, can we instead require a box?  (Maybe this has too big a performance hit.)

I think actually a good bit of the problem I sense intuitively is the `VeryLongAndExplicitClassNames` combined with `[A, C[_], X]`.  It feels like there are these giant things you must focus on (why else would they be so long?) and then there's cryptic type muddle where you can't tell what is for what and why.  Maybe shorter trait names and longer generic parameter names would make it more self-documenting.  (But I'm trying to put myself in the shoes of someone less familiar, which is a difficult thing to do accurately.)

szeiger:
> Actually I view the lack of automatic fallback an advantage

I don't have strong feelings either way, but I'm OK with removing automatic fallback.

> For instance, with CIPT we have a fourth type parameter, `C[_]` or `C[A]`, that exists solely to implement a single widening method.

Implementing the `unconstrained` method there is just a side-effect of having `C` available. It is needed for `with IterablePolyTransforms[A, C]` to get the linearization right.

> It feels like there are these giant things you must focus on (why else would they be so long?) and then there's cryptic type muddle where you can't tell what is for what and why

I wrote this code in the established Scala style of one-letter type parameters (already deviating a bit with `CC` and `Ev` because I didn't just want to pick some random single letter), but in general I would often prefer more expressive names (see also https://github.com/scala/scala/pull/5265).

Ichoran:
Ah, right.  So, if you didn't need to inherit `IterablePolyTransforms`, you wouldn't need that `C`.  Which gets back to my thought of making things less tied into the hierarchy.

There is precedent for longer type parameter names in `CanBuildFrom`.

plokhotnyuk:
Could you, please, state performance as one of main requirements for new collections.
Currently, in hot path, replacing usage of Scala's collections by Java's, because client want to pay less AWS bills

SethTisue:
@plokhotnyuk use a separate ticket, please

Ichoran:
@szeiger - I played with a few variants of CIPT that have only three type parameters, and though I didn't get all the way through the necessary changes they were promising enough that I am now happy with this approach: I think there is a round or two of simplification that could be applied probably without any compiler changes.  So that makes going with this exactly and then slightly refactoring a pretty safe bet.  (Without getting everything compiling I couldn't tell whether type inference would work on the new case, but we can tweak type inference I imagine, as long as the inference is a sensible thing to do in general.

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -189,3 +193,27 @@ trait IterablePolyTransforms[+A, +C[A]] extends Any {
...
+
+  protected def coll: Iterable[A]
```
julienrf: What’s the advantage of that compared to having a self type annotation?

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -81,17 +81,21 @@ trait IterableOps[+A] extends Any {
...
-    *  convert this collection to one of type `C[A]`. Example uses:
+  /** Given a collection factory `fi`, convert this collection to the appropriate
+    * representation for the current element type `A`. Example uses:
     *
     *      xs.to(List)
     *      xs.to(ArrayBuffer)
+    *      xs.to(BitSet) // for xs: Iterable[Int]
     */
-  def to[C[X] <: Iterable[X]](fi: FromIterable[C]): C[A @uncheckedVariance] =
+  def to[F <: FromIterableBase[A]](fi: F): F#To[A @uncheckedVariance] =
```
julienrf: Why not use `fi.To[A]` as the return type instead of the `F#To[A]` projection?

src/main/scala/strawman/collection/IterableFactories.scala
```diff
@@ -4,23 +4,61 @@ package collection
...
+  def empty[A : Ev]: CC[A] = constrainedFromIterable(View.Empty)
 
-  def fill[A](n: Int)(elem: => A): C[A] = fromIterable(View.Fill(n)(elem))
+  def apply[A : Ev](xs: A*): CC[A] = constrainedFromIterable(View.Elems(xs: _*))
 
-  def newBuilder[A]: Builder[A, C[A]]
+  def fill[A : Ev](n: Int)(elem: => A): CC[A] = constrainedFromIterable(View.Fill(n)(elem))
 
-  implicit def canBuild[A]: () => Builder[A, C[A]] = () => newBuilder[A] // TODO Reuse the same instance
+  def constrainedNewBuilder[A : Ev]: Builder[A, CC[A]]
```
julienrf: I would name it `newConstrainedBuilder` instead of `constrainedNewBuilder`.

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -81,17 +81,21 @@ trait IterableOps[+A] extends Any {
...
   // variance seems sound because `to` could just as well have been added
   // as a decorator. We should investigate this further to be sure.
     fi.fromIterable(coll)
 
+  def to[C[_], Ev[_]](fi: ConstrainedFromIterable[C, Ev])(implicit ev: Ev[A @uncheckedVariance]): C[A @uncheckedVariance] =
+    fi.constrainedFromIterable(coll)
```
julienrf: Maybe that’s unrelated to this PR, but I guess this method does not work with collections having binary type constructors (e.g. `TreeMap`), right? BTW, does the unconstrained one work with `HashMap`?

julienrf:
:+1: to this work.

src/main/scala/strawman/collection/Iterable.scala
szeiger: This is the same as in `IterablePolyTransforms`. It allows the use for pseudo-collection types like `String` and `Array`.

src/main/scala/strawman/collection/Iterable.scala
szeiger: In this case it shouldn't matter but I generally try to avoid path-dependent types when type projections are sufficient. With path-dependent types you always need to take care to get the paths right. It is easy to get into a situation where two types are "obviously" the same but the compiler doesn't agree. This is easier with type projections. Everything's fine as long as they dealias to the same type.

julienrf: I think that this kind of type projections is going to be removed from Dotty because they are unsound. Since in our case the `fi` parameter would always be a static `object` I think that using a path-dependent type would not suffer from the problems you are mentioning.

src/main/scala/strawman/collection/IterableFactories.scala
szeiger: I considered that, too, but it sounds like a method that creates a `ConstrainedBuilder`, whereas `constrainedNewBuilder` is consistent with `constrainedFromIterable`.

src/main/scala/strawman/collection/Iterable.scala
szeiger: Right, this still needs to be improved. The unconstrained version cannot work for binary type constructors. We only have an element type `A` in `Iterable`. While we can restrict it to work only for `(A1,A2)` there is no way to compute these types `A1` and `A2` without implicits. Conversions to maps either need another special case or a `ConstrainedFromIterable`. Since we're special-casing `Map` in other places we should probably overload `to` with two additional cases for unconstrained and constrained maps.

julienrf: A `CanBuildFrom`-like would nicely solve these issues. It’s probably OK if we have it only at this place instead of having it everywhere as in the current collections.

src/main/scala/strawman/collection/Iterable.scala
szeiger: Ah, right, this looked safe enough for me but I haven't tested it on Dotty. It's pretty much the same as https://github.com/lampepfl/dotty/pull/1051#issuecomment-181577039 though, so it won't work. I'll change it to use the path-dependent type instead.

src/main/scala/strawman/collection/IterableFactories.scala
```diff
@@ -4,23 +4,61 @@ package collection
...
+
+trait FromIterableBase[-B] {
+  type To[X]
+  def fromIterable[E <: B](it: Iterable[E]): To[E]
+}
 
 /** Base trait for instances that can construct a collection from an iterable */
-trait FromIterable[+C[X] <: Iterable[X]] {
-  def fromIterable[B](it: Iterable[B]): C[B]
+trait FromIterable[-B, +C[_]] extends FromIterableBase[B] {
```
odersky: Why do we need the bound here? It would be great if we could avoid the additonal `B` parameter.

szeiger:
@odersky 

> It would be great if we could get rid of at least one. This seems to be possible since we seem to agree that automatic widening is not needed.

I don't see how. We need the unconstrained type constructor for *manual* widening and to get the linearization right. There is no automatic widening in this version.

> can we find a less visible way to represent the parameter, e.g. a as a type member.

I would have preferred that, too, but due to variance the default would be `Nothing` if you don't specify anything else but we need `Any`. We could have a type alias for the `Any` case but I think it would only complicate things further. OTOH, how often do you need to use `FromIterable`? Most users will only see it in `to` and in the factory methods (if they look at the actual source; scaladoc and IDEs show the instantiated `Any` bound in most cases).

> or, should we not bother at all about this? How much code duplication would we get if we special cased those cases where the bound is not `Any`

`String` and `BitSet` are the cases that we know about. They would either need special-casing or be modeled as constrained types with implicit evidence. I would like to avoid the latter due to the performance implications.

odersky:
@szeiger My tendency would be to special case `String` and `BitSet`. The reasoning is that it's better to require some duplication for implementers of non-parametric collection types than to complicate the framework to a point where it is too hard for them to extend it anyway. The existing collections were a lesson in that respect.

Ichoran:
Also in favor of special-casing: it is important that `String` especially but also `BitSet` be really fast, which means that many things will be written as a special case anyway.

I'll try to finish up some of my 3-parameter CIPTs tonight or the next day.  The basic idea is to drop the inheritance of IPT in CIPT and push responsibility for putting something sensible for `C[A]` down to the leaf class implementations and/or use protected methods to enforce sensible linearization.  I got multiple things that compiled with very little work, but haven't had time to test them all in enough depth to know what flaws they might have.

szeiger:
I tried again to create a version without the upper bound but it breaks everything. Having this upper bound is essential for the unification of constrained and unconstrained collection types and `MonomorphicIterableFactory`. Once you have an evidence value, you can create such a factory but only if you have an upper bound (that you can set to the element type for which you got the evidence).

Ichoran:
I ran into some problems also, but I haven't exhausted the list of things I want to try.  That said, they're all tweaks on this design.  I think we can merge this change as-is and improve it if possible rather than restarting from scratch.

szeiger:
I'm also still trying a few things. I managed to restrict the upper bound to `FromIterableBase` (now called `Build`). I'm still working on a possible unification of `Build` with `CanBuild` (previously `MonomorphicIterableFactory`) and I also have a new idea for `CanBuildFrom` without `CanBuildFrom`. I'll update this PR tomorrow with some minor changes tomorrow and then merge it, possibly following up with a new one with additional improvements.

Ichoran:
I've played around with a variety of things that don't really work, but the underlying problem is that the following two things have equal priority:
```
def map[B](f: A => B): C
def map[B: Ev](f: A => B): D
```
But this isn't sensible.  If the two are really equivalently good, you shouldn't be able to compose them without a warning, because you will typically be unable to use the right one.  Otherwise, the one with extra evidence should be presumed to be used first.

I think the best way to solve this is by modifying the compiler, not by creating elaborate type signatures or finding a peculiar trick that just happens to work.  (I've run into this very problem multiple times before, and never been saved by the default behavior.)

julienrf:
@Ichoran Which version of Scala are you using? In the custom 2.12.2-ebe1180-SNAPSHOT you should not have this ambiguity.

odersky:
@julienrf 

> Which version of Scala are you using? In the custom 2.12.2-ebe1180-SNAPSHOT you should not have this ambiguity.

That's news to me. What changed in the snapshot? I just checked what would be involved in supporting this and the change is quite substantial. Essentially, it means making overloading resolution and implicit search recursively depend on each other, which could have far reaching consequences for performance.

julienrf:
Oh, sorry I was wrong, this version of Scala contains this fix: https://github.com/scala/scala/pull/5742 which is unrelated to overloading.

odersky:
I still have the suspicion we are on the wrong track here. What exactly are we trying to achieve? The strawman before this PR already handled String and Array which are examples of specialized and constrained collections, respectively. So what added complications cause us to complicate key base traits with additional parameters?

szeiger:
I think Rex is referring to a design that doesn't require `ConstrainedIterablePolyTransforms extends IterablePolyTransforms`. You do need *some* mechanism for disambiguation and I doubt any alternative would be simpler than what we already have (methods in subtypes take precedence over methods in supertypes). The current implmentation in maser "solves" this by overriding the evidence-based `map` in `TreeSet` but this approach clearly doesn't scale. You would have to override every overloaded method in every leaf class.

`String` and `Array` are not real collection types, they only get some extension methods, so not all of the collections machinery is necessary or expected to work for them. For example, we do not have a working `.to` for anything but unconstrained collection types. The same goes for methods outside the collections framework like `Future.sequence`.

As mentioned yesterday, I have a version that doesn't add an upper bound to `FromIterable` and `IterableFactory` but, of course, that means that these key abstractions still cannot be used for all collection types (and basic methods like `to` need to use `Build` instead of `FromIterable`). They are designed to work with unconstrained unary type constructors only. If we want to abstract over other collection types we need more powerful abstractions.

Ichoran:
That's correct @szeiger.  I was trying to reduce the complexity by making traits that contain only the type parameters needed to express the functionality they contain, not extra ones that help resolution of implicit search with overloaded methods.  And the problem is that without inheritance providing a precedence that breaks ties in which overloaded method should be used, you can't pick the "right method".  I was unable to find any combination of self-types and type members that would work.

What does "work" is making the classes not even inherit from Iterable and using the same kind of boxing method that `Array` and `String` use, but these will then have overhead if used in a widened context (as they will be boxed) and also will be hard to maintain once we have the full set of methods that will need to be proxied.

What also works is a kind of diamond inheritance where the leaf classes inherit both from the constrained branch and the unconstrained branch, but the user gets to see only one with a no-op method to widen the type.  But this complicates implementation greatly, as you have to hide the actual constructors from the user for every leaf class.

And finally, not abstracting over evidence at all kind of works, I think, but I didn't work it all the way through, so I'm not sure yet.  You need the upstream machinery for building with evidence, but you don't need CIPT if you replicate it at every point in the hierarchy where you have different evidence.

@odersky - I don't think the change I suggested would complicate search beyond what it is now.  It's just an extra tie-breaking signal, "oh, there's an implicit evidence parameter!", to apply when considering what method to use.  Alternatively, a tie-breaking annotation could be introduced.  It *is* a general problem that Scala is not very good at solving, right now, in a simple way.

odersky:
> @odersky - I don't think the change I suggested would complicate search beyond what it is now. It's just an extra tie-breaking signal, "oh, there's an implicit evidence parameter!", to apply when considering what method to use. 

Yes, but that would be too confusing.

Consider

   def f[A](x: A)
   def f[A: Ordering](x: A)

    f(1)           // now: ambiguous, would be OK
    f(1: Any)   // would give error: No implicit Ordering found for Any

The behavior in the last line would be more confusing than the ambiguity error we get now. To do a proper job, we'd need to look at the implicit whn doing overloading resolution but that comes with its own challenges.

What's the downside of letting CIPT inherit from IPT?

Ichoran:
The downside of letting CIPT inherit from IPT is that it requires a fourth type parameter, which is the type of the nonconstrained (widened) type.  If we're willing to accept that, Stefan's scheme can go through as is.

szeiger:
I've added some new changes:

- The upper bound in `FromIterable` and `IterableFactory` is gone. It is only present in the `Build` abstraction (which replaces `FromIterableBase`).

- `MonomorphicIterableFactory` first becomes `CanBuild` and then gets incorporated into the `Build` hierarchy for a "CBF light" that enables `breakOut`-like functionality in a style that is consistent with the new way of building in `to` (pass a factory object instead of specifying the target type).

- Type projections are replaced by path-dependent types.

szeiger:
BTW, I'm still not happy with adding an implicit `CanBuildFrom`-like abstraction at all. It feels like we're trying to shoehorn old patterns into the new design. There has to be a simpler way. I'll try that tomorrow.

odersky:
I think it looks much better now. I am still not quite happy with the `Build` thing. The name is not very telling and I believe that's because it's function is not so obvious either. I stick to these points because `Build`, or a trait like it, is in the base types of every collection type, which makes it quite visible in ScalaDoc. If we could have something like Build only as parts of the factories that would be less problematic.

Ichoran:
I made a small modification that pushes the `Ev` type in as a type member of CIPT.  Code at https://github.com/szeiger/collection-strawman/pull/1

As far as I can tell it's a win on all fronts.

I also thought about renaming the factories, but there are a lot of traits that do seemingly similar things with slight variations.  Naming is hard, so I left them alone after thinking about it for far too long.  I do think we need to rename them at some point.  (For now, the long technically correct names are fine, but we should revisit them once the set of them stabilizes so we know what it is we need to name.)

szeiger:
I've just pushed a new commit. Ever since the implicit builders were merged I had this nagging feeling that we had the whole thing backwards. In the new design the *source collection* is responsible for building a target collection, we do not rely on an implicit CanBuildFrom for that, so why should we do that for implicit builders? It comes with the same old problem of getting different results depending on the static type (because we rely on implicit search in companion objects to decide which builder to use) and it also caused me all kinds of problems trying to unify implicit builders and companion objects.

So I went back to the drawing board and started with the assumption that the source collection has to decide what to build. We already had `Buildable` for collections that allow eager building of the same collection with the same element type. A more consistent name for it is `MonoBuildable` and the `newBuilder` method is now called `newBuilderWithSameElemType`. This is consistent with `IterableMonoTransforms` and `fromIterableWithSameElemType`. What we were missing though was the *poly* version of it, so I added `PolyBuildable`. Unlike `MonoBuildable` it is not limited by variance, so its `newBuilder` method can be public. This allows anyone who has a strict unconstrained collection (i.e. a `PolyBuildable`) to build the same collection type with any element type. For constrained collection types I added `ConstrainedPolyBuildable` with an extra implicit evidence to mirror `ConstrainedFromIterable`.

Now we have everything in place to write methods outside the collection framework (like `Future.sequence`) in the same style as methods within, i.e. by overloading for constrained and unconstrained collection types:

```scala
  def optionSequence1[C[X] <: Iterable[X] with PolyBuildable[X, C], A](xs: C[Option[A]]): Option[C[A]] =
    xs.foldLeft[Option[Builder[A, C[A]]]](Some(xs.newBuilder)) {
      case (Some(builder), Some(a)) => Some(builder += a)
      case _ => None
    }.map(_.result)

  def optionSequence1[Ev[_], CC[_], A](xs: ConstrainedPolyBuildable[Option[A], CC, Ev] with Iterable[Option[A]])(implicit ev: Ev[A]): Option[CC[A]] =
    xs.foldLeft[Option[Builder[A, CC[A]]]](Some(xs.newConstrainedBuilder)) {
      case (Some(builder), Some(a)) => Some(builder += a)
      case _ => None
    }.map(_.result)
```

Since having to overload everything (if you want to support constrained collection types) is rather awkward, the next step was how to unify both through an implicit value. The result is essentially `CanBuildFrom` but I changed the `To` type from a type parameter to a type member because it is always a computed result type and not a type that guides the computation:

```scala
trait BuildFrom[-From, -Elem] {
  type To
  def newBuilder(from: From): Builder[Elem, To]
  def fromIterable(from: From)(it: Iterable[Elem]): To
}
```

Note that there is no method to get a `Builder` without a `From` instance. The source collection is still responsible for building!

While usage is essentially the same as for `CanBuildFrom` (we could even define a type alias to make it look the same) the way the implicits are materialized is very different. The companion object of `BuildFrom` has two methods to get a `BuildFrom` from a `PolyBuildable` and from a `ConstrainedPolyBuildable` (if an evidence value is available). This is defined in a self-contained way on top of the collection framework. No collection type has to define its own `BuildFrom` instances.

All implicit builders in companion objects are gone and so are the `BuildStrict`, `BuildConstrained` and `BuildStrictConstrained` abstractions. None of that is needed anymore.

Now we can write a single method like this that abstracts over `PolyBuildable` and `ConstrainedPolyBuildable`:

```scala
  def optionSequence[CC[X] <: Iterable[X], A](xs: CC[Option[A]])(implicit bf: BuildFrom[CC[Option[A]], A]): Option[bf.To] = ...
```

We can even do `breakOut`, and in way that is much more consistent with the new design than my previous attempt. In addition to the implicit *values* that `BuildFrom` provides for buildables, it also defines implicit *conversions* from all four basic collection factory types (`IterableFactory`, `ConstrainedIterableFactory`, `MapFactory` and `ConstrainedMapFactory`) to `BuildFrom`. This allows you to pass any collection factory that you could use in a `.to` call to a method that expects a `BuildFrom`:

```scala
    val xs4 = immutable.List[Option[(Int, String)]](Some((1 -> "a")), Some((2 -> "b")))
    val o4 = optionSequence(xs4)(immutable.TreeMap)
```

In fact, we can use it to define `to`:

```scala
  def to[F <: TypeConstrainedFromIterable[A]](fi: F): fi.To[A @uncheckedVariance] =
    fi.fromIterable(coll)

  // Generic version of the method above that can build anything with BuildFrom. Note that `bf` is not implicit.
  // We never want it to be inferred (because a collection could only rebuild itself that way) but we do rely on
  // the implicit conversions from the various factory types to BuildFrom.
  def to(bf: BuildFrom[Iterable[A], A]): bf.To =
    bf.fromIterable(coll)(coll)
```

The first version is purely for performance: It allows collection types without an evidence value to be built in a more direct way. Everything else uses the implicit conversion to `BuildFrom`.

`FromIterable` still extends `TypeConstrainedFromIterable` (which was called `Build` in the previous version). We could easily change that so that most collection types do not have to extend `TypeConstrainedFromIterable`. It would require an extra overload for `to` to have efficient implementations for both.

Ichoran:
Aha!  I like this one.  I think it still needs a couple of rounds of simplification and renaming to make it comprehensible and to get rid of some of the veryLongNamesEspeciallyWhenTheyRepeat: LongNameEspeciallyWhenTheyRepeat = new LongNameEspeciallyWhenTheyRepeat things.

But this seems like a better design overall.  I have to think through and test cases where the source collection being more in-charge might be a problem.  But from first principles, as you say, this seems better: if the source is in charge, dynamic dispatch takes care of loss of type information.  Conceptually, then, the issue would be creating collections from scratch.  I need to look into that in more detail.

julienrf:
> Note that there is no method to get a Builder without a From instance.

That could be useful to implement `unfold`s, though.

src/main/scala/strawman/collection/BuildFrom.scala
```diff
@@ -0,0 +1,68 @@
...
+    def newBuilder(from: ConstrainedPolyBuildable[A, CC, Ev]): Builder[E, To] = from.newConstrainedBuilder[E]
+    def fromIterable(from: ConstrainedPolyBuildable[A, CC, Ev])(it: Iterable[E]): To = from.constrainedFromIterable(it)
+  }
+
+  /** Convert an IterableFactory to a BuildFrom */
+  implicit def buildIterableFactory[C[_], E](fact: IterableFactory[C]): BuildFrom[Any, E] { type To = C[E] } = new BuildFrom[Any, E] {
+    type To = C[E]
+    def newBuilder(from: Any): Builder[E, To] = fact.newBuilder
+    def fromIterable(from: Any)(it: Iterable[E]): To = fact.fromIterable(it)
+  }
```
julienrf: This sounds like smell: we should not have to introduce this `Any` parameter if we don’t need it.

src/main/scala/strawman/collection/IterableFactories.scala
```diff
@@ -3,24 +3,46 @@ package collection
...
-  def fromIterable[B](it: Iterable[B]): C[B]
+trait FromIterable[+C[_]] extends Any with TypeConstrainedFromIterable[Any] {
+  type To[X] = C[X] @uncheckedVariance
 }
 
-/** Base trait for companion objects of collections */
-trait IterableFactories[+C[X] <: Iterable[X]] extends FromIterable[C] {
+/** Base trait for instances that can construct a collection from an iterable for certain types
+  * (but without needing an evidence value). */
+trait TypeConstrainedFromIterable[-B] extends Any {
```
julienrf: `UpperBounderFromIterable`?

src/main/scala/strawman/collection/immutable/TreeSet.scala
julienrf: This was not an override, just an implementation, FYI.

src/main/scala/strawman/collection/mutable/SortedSet.scala
```diff
@@ -1,12 +1,17 @@
...
+import strawman.collection.ConstrainedIterablePolyTransforms
+
 trait SortedSet[A]
   extends collection.SortedSet[A]
     with Set[A]
     with SortedSetLike[A, SortedSet]
 
 trait SortedSetLike[A, +C[X] <: SortedSet[X]]
   extends collection.SortedSetLike[A, C]
+    with ConstrainedIterablePolyTransforms[A, Set, SortedSet]
```
julienrf: `with ConstrainedIterablePolyTransforms[A, Set, C]`

julienrf: By the way, would it be possible to pull up this line at the level of `collection.SortedSet` so that we don’t repeat it for `collection.mutable.SortedSet` and `collection.immutable.SortedSet`?

src/test/scala/strawman/collection/test/Test.scala
```diff
@@ -331,6 +336,8 @@ class StrawmanTest {
 //    val xs7 = xs.map((k: String, v: Int) => (new Foo, v)) Error: No implicit Ordering defined for Foo
     val xs7 = (xs: immutable.Map[String, Int]).map((k, v) => (new Foo, v))
     val xs8: immutable.Map[Foo, Int] = xs7
+    val xs9 = xs6.to(List).to(mutable.HashMap)
```
julienrf: :+1:

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -7,36 +7,69 @@ import strawman.collection.Iterable
...
-    optionSequence(xs2)(immutable.TreeSet.canBuild[String])
+    val o2 = optionSequence(xs2)
+    val o2t: Option[immutable.TreeSet[String]] = o2
 
-    val xs3 = mutable.ListBuffer(Right("foo"), Left(0), Right("bar"))
-    eitherSequence(xs3)(mutable.ListBuffer.canBuild[String])
+    // Breakout-like use case from https://github.com/scala/scala/pull/5233:
+    val xs4 = immutable.List[Option[(Int, String)]](Some((1 -> "a")), Some((2 -> "b")))
+    val o4 = optionSequence(xs4)(immutable.TreeMap) // same syntax as in `.to`
+    val o4t: Option[immutable.TreeMap[Int, String]] = o4
```
julienrf: :+1:

lrytz:
Also like the latest change!

I'm a bit worried about keeping / promoting the "implicit buildFrom" pattern. We are moving away from it for the internal collections implementation and use overloading instead, why should we recommend it for external methods? I see that collection types don't have to define their own BuildFrom instances, which certainly makes things simpler. But there are other issues with implicit BuildFrom, like the complex signatures, Scaladoc, IDE discovery. For example, I defined your `optionSequence` methods as extension methods, here's how it looks:

image

image

src/main/scala/strawman/collection/BuildFrom.scala
szeiger: We do need for the "proper" cases (implicit values for poly builders). When you do a manual breakout, the `from` value is ignored.

szeiger:
@Ichoran @julienrf Building an arbitrary target collection will require a factory object. Just as in the case of `to` it is not type-driven. You have to pass the actual object. You can write it as a single method by using a non-implicit `BuildFrom` the same way the new `to` overload does. The alternative is to overload it for the different factory kinds. In fact, this is something I also considered for `to` but you have to support four factory kinds (unconstrained iterable/map, constrained iterable/map) if you want it to be comprehensive (unlike the `to` method in the current collections library which does not support map types).

@lrytz I tried it with `To` as a type parameter instead of  a type member but it didn't help, either. I don't think there's much we can do other than using `BuildFrom`-based methods sparingly. If you want a single definition that can abstract over kinds, implicits are the only way to do it.

src/main/scala/strawman/collection/BuildFrom.scala
szeiger: Alternatively, we could have a subclass like `BuildFromAny[-Elem] extends BuildFrom[Any, Elem]` with overloads that do not take a `from` value. This would be enough for the non-implicit use case in `to`. Is getting rid of the `Any` parameter here worth the extra complexity of a new class with two new methods?

julienrf:
> Building an arbitrary target collection will require a factory object.

Then we will not be able to write generic code as you did with `optionSequence`, but in cases where we have no “from” collection. A typical use case would be “unfolds”.

With the current collections we can write the following:

~~~ scala
  class Unfolder[That] {
    def apply[S, A](init: S)(f: S => Option[(A, S)])(implicit cbf: CanBuildFrom[Nothing, A, That]): That = {
      val builder = cbf.apply()
      …
    }
  }
~~~

And then `That` can be `List[X]` as well as `TreeMap[X, Y]`.

szeiger:
Yes, that's correct. It is consistent with `to` and with the new way of doing `breakOut`-like operations and it doesn't look like a huge loss to me. In most cases you want to build a specific collection type, so you just use the correct factory object. In my experience with using `breakOut` the result is usually assigned to a variable, in which case the new syntax is actually shorter because you don't have to spell out the element type (in addition to importing and using `breakOut` itself). If you want to abstract over collection types you have to pass a `CanBuildFrom` around and repeat the necessary type parameters for that everywhere. This was non-trivial in the old design and doesn't change much here (it may be slightly simpler now with `To` being a type member).

src/main/scala/strawman/collection/immutable/TreeSet.scala
szeiger: Trying to compile master without this line:

```
[info] Compiling 37 Scala sources to /Users/szeiger/code/collection-strawman/target/scala-2.12/classes...
[error] /Users/szeiger/code/collection-strawman/src/main/scala/strawman/collection/immutable/TreeSet.scala:10: class TreeSet needs to be abstract, since method map in trait SortedPolyTransforms of type [B](f: A => B)(implicit ordering: Ordering[B])strawman.collection.immutable.TreeSet[B] is not defined
[error] (Note that A => B does not match A => B: their type parameters differ)
[error] final class TreeSet[A]()(implicit val ordering: Ordering[A])
[error]             ^
[error] one error found
```

src/main/scala/strawman/collection/mutable/SortedSet.scala
szeiger: Yes, it's possible.

src/main/scala/strawman/collection/immutable/TreeSet.scala
julienrf: Yes, that’s what I meant by “this was an implementation”.

szeiger: It's not though. The error message is misleading. I haven't tried to track this down further but there is no abstract `map` method anywhere in the inheritance hierarchy of `TreeSet`. It inherits two concrete implementations but without enforcing `CIPS <: IPS` the types don't line up.

julienrf:
My global feeling about this PR: I definitely think we need a `CanBuildFrom` abstraction as you proposed, because it makes it possible to write powerful generic methods such as `def sequence[F[X] <: Iterable[X], A](xs: F[G[A]])(implicit bf: BuildFrom[F[G[A]], A]): G[bf.To]`. However, the hierarchy really becomes more complex with this PR. Maybe we should keep the `CanBuildFrom`-like thing but keep a simpler hierarchy (at the price of duplicating more things between the four kinds of factories).

Ichoran:
@julienrf - I concur.  That was one of the "rounds of simplification" I envisioned, along with one more attempt to check whether any more type parameters can be made type members instead.  (Along with a renaming to things that might technically be a little less accurate but are more intuitive, e.g. `IterableOfSame` instead of `IterableMonoTransforms`.  Same/Other or Alike/Unlike can be useful here.)

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -81,17 +81,22 @@ trait IterableOps[+A] extends Any {
...
     *
     *      xs.to(List)
     *      xs.to(ArrayBuffer)
+    *      xs.to(BitSet) // for xs: Iterable[Int]
     */
-  def to[C[X] <: Iterable[X]](fi: FromIterable[C]): C[A @uncheckedVariance] =
-  // variance seems sound because `to` could just as well have been added
-  // as a decorator. We should investigate this further to be sure.
+  def to[F <: TypeConstrainedFromIterable[A]](fi: F): fi.To[A @uncheckedVariance] =
     fi.fromIterable(coll)
```
odersky: I don't think you need the `F` parameter here. 

    def to(fi: TypeConstrainedFromIterable[A]): fi.to[A]

should work, no?

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -81,17 +81,22 @@ trait IterableOps[+A] extends Any {
...
 
+  // Generic version of the method above that can build anything with BuildFrom. Note that `bf` is not implicit.
+  // We never want it to be inferred (because a collection could only rebuild itself that way) but we do rely on
+  // the implicit conversions from the various factory types to BuildFrom.
+  def to(bf: BuildFrom[Iterable[A], A]): bf.To =
+    bf.fromIterable(coll)(coll)
+
   /** Convert collection to array. */
```
odersky: I am not sure about overloading `to`. Why do we need both versions? Also, naming conventions vary a lot
`TypeConstrainedFromIterable` and `BuildFrom` seem to have nothing in common.

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -189,3 +194,74 @@ trait IterablePolyTransforms[+A, +C[A]] extends Any {
   def zip[B](xs: IterableOnce[B]): C[(A @uncheckedVariance, B)] = fromIterable(View.Zip(coll, xs))
   // sound bcs of VarianceNote
 }
+
+/** Transforms over iterables that can return collections of different element types for which an
+  * implicit evidence is required.
+  */
```
odersky: Should explain in comment with examples why we need both `C` and `CC`. It looked non-senscial at first tome, and I started to understand only when I looked at instantiations of the class. (of course if we found a way to merge `C` and `CC` that would be even better).

src/main/scala/strawman/collection/IterableFactories.scala
odersky: I would say `BoundedFromIterable`. The upper is implied if left out.

odersky:
Should we try to merge Bounded (aka TypeConstrained) and Constrained? We can always express Bounded as a Constraint where we demand an instance of `<:<`. How many examples are there where we need Bounded? I know of String and BitSet, are there others?

src/main/scala/strawman/collection/Iterable.scala
szeiger: I'll give it a try. (I'm used to writing type projections where this wouldn't work)

src/main/scala/strawman/collection/Iterable.scala
szeiger: We don't need both. The `TypeConstrainedFromIterable` version covers the most common case and doesn't require an implicit conversion from the factory type to a `BuildFrom`. It's only there for performance (but I haven't benchmarked it). The `BuildFrom` version alone is sufficient.

szeiger: Hm, actually `BitSet` doesn't currently work with the `BuildFrom` version. I'll try to fix it.

szeiger:
I pushed another commit to address @odersky's review comments.

- Added more docs to `ConstrainedIterablePolyTransforms`

- Removed unnecessary type parameter in `to(TypeConstrainedFromIterable)`

- Renamed `TypeConstrainedFromIterable` to `BoundedFromIterable`

The problem with `BitSet` was caused by the companion object not extending `IterableFactory` because it is type-constrained. I fixed it by adding a supertype `BoundedIterableFactory` that covers all cases. OTOH we don't really need `BoundedFromIterable` so I removed that in exchange. This means that the "bounded" abstraction is only visible in companion objects but not in the collection types.

julienrf:
@lrytz I noticed that I don’t have the same result as you:

bitset-map

What bothers me is this `BitSet.this.Ev` thing… Which version of IntelliJ are you using? (I’m using 2017.1.1)

lrytz:
@julienrf this changed at some point between my comment (https://github.com/scala/collection-strawman/pull/45#issuecomment-285624422) and now. On the current master branch I get the same as you.

julienrf:
@szeiger In practice, do you think we will have other instantiations of `Ev` than `Ordering`? The current design leads to some issues, notably when used with dotty: [this method](https://github.com/scala/collection-strawman/blob/46b16bafc303c8aacff8eed2b4e722f186fa7e24/src/main/scala/strawman/collection/BuildFrom.scala#L36), by taking an unconstrained type parameter `Ev` makes the implicit search explode. Also, there is the IDE issue that I reported above.

What would be the impact of removing this abstract layer `ConstrainedPolyBuildable` and having just an `OrderingConstrainedPolyBuildable` (with no `Ev` type parameter)?

szeiger:
I can't think of any other instantiations in the core collection library. Apart from some reusable interfaces we wouldn't lose much by not abstracting over `Ev[_]`. People could still define collection types with other constraints by duplicating the interfaces. The only additional work required in these cases is implementing the `CanBuild` implicits (but as you point out abstracting over `Ev[_]` for these implicits doesn't work correctly in Dotty anyway).

PR #51, opened 2017-03-27
Catenable steque

edmundnoble:
Moved from scala/scala:

Sketch of stack-ended queues for Scala 2.13 (`Steque`). A persistent, functional data structure with implementation and design courtesy of `fs2.util`, it provides:

- constant-time (very fast: one or two constructor calls) `:+` (append), `+:` (prepend), `++` (concat) and `fromSeq[A](seq: Seq[A]): Steque[A]`
- amortized-constant-time `head`, `tail`
- linear-time `foldLeft`, `foreach` and friends

In these tasks, it beats out `Vector` in speed by an order of magnitude - see benchmarks at edmundnoble/scabsbench with results at https://ethercalc.org/kknm1xz3mb5s_3ht2cokr07oy/app. Some things it does badly include:

- indexing (linear time)
- last/init (linear time)
- allocates more than Vector (because Vector is chunked, and uses Arrays underneath)

In terms of implementation, Steque is an association structure built from an unbalanced binary tree which may have entire `Seq`s or single values at its leaves. For this reason, the asymptotics of consuming a `Steque` will also depend on those of the underlying collection. The asymptotics of, `head`, `tail` and `foldLeft` will require equivalent asymptotics for all Seq values included in the `Steque`.

Steque is essentially a binary tree which will graft subtrees from the left to the right as needed (i.e. as uncons is called), eventually flattening out into the trivial right-associated binary tree (cons list). It may be possible to reverse the algorithm to graft subtrees to the left and create a left-associated binary tree to provide amortized-constant unsnoc, but alternating between unsnoc and uncons will be incredibly expensive.

Potential optimization notes:

- It may be more performant to remove single leaves and keep `Array` leaves of a fixed width.
- `splitAt` may be able to avoid traversing the entire structure.


julienrf:
Hi, thanks for working on that. That looks super interesting, however we already have lots of `LinearSeq` collection types (`Vector`, `immutable.Array`, `List`). From what you said, `Steque` would be a nice candidate to replace `Vector`. Could you run the benchmarks and show us how it compares with `Vector`?

DarkDimius:
> From what you said, Steque would be a nice candidate to replace Vector. Could you run the benchmarks and show us how it compares with Vector?

Doesn't look like it. By-Index access is slow in this collection.

Ichoran:
It can't replace vector but it *might* be able to replace `List` as the default `Seq`.  Depends what the benchmarks look like!

DarkDimius:
> It can't replace vector but it might be able to replace List as the default Seq. Depends what the benchmarks look like!

I'd be very suprised. List is as-good as it you can ever get for head&tail, while this implementation has both conditional jumps and and a cycle.
I won't be willing to sacrifice List.

I will be glad to be proven wrong by a good benchmark.

DarkDimius:
Note that list is chosen as default Seq because most collections are <4 elements.
To replace default sequence it's not about beating asymptotics. It's about constants for small sizes.

DarkDimius:
>  * Conversion from a `Seq` takes constant time, but maintaining `uncons`
>  * performance in that case depends on the underlying `Seq`'s `uncons` performance.

I'd also prefer default collection not to have such surprises.

Ichoran:
Agreed about small sizes.  If `Steque` can be special-cased for such sizes to work well, then I still think it could replace `List` potentially.  Most likely, though, it would be yet another `LinearSeq` to choose when neither `Vector` nor `List` are suitable.

LPTK:
So it looks like this would be a very good data structure for functionally _building_ sequences.

```scala
def boundVars(ast: AST): Seq[Var] = ast match {
  case Let(symbol, expr, body) => symbol +: (boundVars(expr) ++ boundVars(body))
  case Fun(params, body) => params ++ boundVars(body)
  case ...
}
```

It would make sense to use it in the case above as it would be the fastest way to construct the resulting Seq. Unfortunately, currently operations like `++` and `+:` determine the resulting Seq implementation based on that of the current collection (which may be wholly inadequate for inductive sequence building).

It would be nice to at least be able to orient the implementation used with a return type annotation as in:

```scala
def boundVars(ast: AST): Steque[Var] = ...
```

DarkDimius:
LPTK, in a use case that you're describing, I agree that it _may_ be better than all collections that we have currently in stdlib. The problem that I see with this is that "may" word:
>       Conversion from a Seq takes constant time, but maintaining uncons
>       performance in that case depends on the underlying Seq's uncons performance.

If we were to include collection for this use-case, I'd expect [Conc-trees](https://axel22.github.io/resources/docs/lcpc-conc-trees.pdf) to be superior.

Still, I don't think this would be a convincing use-case. I'd prefer to have less collections in stdlib to maintain and instead have collections that are perfectly-suited for a more rare use-cases as separate libraries. 

It is very good that we're trying if other collections can be accommodated in the current design without modifying internals as this proves that those separate libraries would be possible to have.

LPTK:
@DarkDimius I agree it probably doesn't have its place in the standard library, at least as it is now. After all, in the use case I describe I would rather use something that is _not necessarily_ a Seq, but more like a functional SeqBuilder that I can later convert to a Seq with `toLinearSeq`, `toIndexedSeq`, etc. depending on my use case. 

```scala
def boundVars(ast: AST): SeqBuilder[Var] = ast match {
  case Let(symbol, expr, body) => symbol ~: (boundVars(expr) ~~ boundVars(body))
  case ...
}
boundVars(myTree).toList match { case h :: tl => ... }
```

SeqBuilder could still be a Seq and lazily adapt to the way it's decomposed by consumers, a bit like this Steque does, but it would need to do it better (not just accommodate head/tail decomposition). In that case, that data structure could be the default _SeqBuilder_.

jvican:
> It is very good that we're trying if other collections can be accommodated in the current design without modifying internals as this proves that those separate libraries would be possible to have.
  
If this is the case, we can have a community-driven Scala Platform module to support extra collections, suitable for use cases that are not as common as to deserve citizenship in the core collections.

edmundnoble:
@Ichoran @DarkDimius I understand your concerns. I believe the best option to deal with them is to replace the `Single`/`OfSeq` leaf structure with an Array of a fixed size. Thank you for the paper, it is very readable and seems to show these reasons for Steque being more useful than conc-trees:
1. Concat is an incredibly small constant for Steque, whereas for conc-trees it's O(min(n1, n2)) and conc-tree ropes it's O(log(n))
2. Append is an incredibly small constant for Steque, whereas for conc-trees it's O(log(n))
3. Conc-tree ropes do not retain their performance when used persistently.
4. Conc-trees and conc-tree ropes allocate more than Steque.

@LPTK You are correct. Steque allows for every possible functional sequence building pattern; assuming you will consume the sequence via an ordinary fold, it's incredibly efficient, where the degenerate case of the fold constructs a different data structure.

edmundnoble:
@Ichoran @DarkDimius I took a look; so `Steque` is actually equivalent to `List` if you just use `cons`; the folding loop is even equivalent in that case. `Steque` just adds on to the concat and snoc cases. As well, adding the array-leaves does nothing for performance except increase the memory overhead and constant factors.

edmundnoble:
I'm closing this PR in favor of @drdozer's `InceptionQ` which performs the same as `Steque` for queue operations, and provides equally performant stack operations.

PR #49, opened 2017-03-24
Strawman implementation of mutable.ArrayDeque

pathikrit:
screenshot 2017-03-24 17 26 19

Prior discussion: 
https://contributors.scala-lang.org/t/using-circular-buffers-as-mutable-arraybuffer/454

ArrayDeque's (implemented using a growable circular buffers using modular arithmetic for indexing) are strictly better data structures than vanilla ArrayBuffers.

This implementation adds `O(1)` `append`, `prepend`, `removeHead`, `removeTail`, `get`, `update` and `clear` APIs. We can replace all current implementations of mutable data structures like stack, queues and array buffers with ArrayDeque.

PR #52, opened 2017-03-27
ArraySeq (f.k.a. Spandex)

odd:
An _ArraySeq_ is an (ostensible) immutable array-like collection designed to support the following performance characteristics:
  * constant time `head`, `last`, `tail`*, `init`*, `take`*, `takeRight`*, `takeWhile`*, `drop`*, `dropRight`*, `dropWhile`*, `slice`* and `reverse` (* = depending on the auto-trimming mode in use)
  * amortised constant time `prepended`/`prependedAll`, `appended`/`appendedAll` and `concat` (depending on the complexity of `java.lang.System.arraycopy`)
  * efficient indexed access
  * linear time iteration (i.e. `iterator` and `foreach`) with low constant factors
  * reasonable memory usage (approximately double that of an array with the same number of elements)

Elements can be added to both the front and the rear of the underlying array (via `prepended`/`prependedAll` and `appended`/`appendedAll`/`concat` respectively), but never more than once for any given array slot. For any operation trying to use an already occupied slot (such as `appended`, `prepended` or `updated`) an overlay is provided that can store the index and element pair using a `HashMap` (thus avoiding having to copy the underlying array).
The overlay will only be allowed to grow according to the used auto-trimming mode, after which it will be merged with the underlying array on the next transformation operation.

To guarantee that only a single thread can write to an array slot an atomic long is used to guard the low and high index assignments.

When the underlying array is too small to fit a requested addition a new array is allocated (the size is dependent on the auto trimming mode used).

To ensure that no longer accessible array slots (i.e. on the returned instance of a `slice`-based operation) are freed, trimming can be used. Trimming can either be performed manually (via the `trim` method) or automatically (by specifying an auto-trimming mode at creation or via the `withAutoTrimming` method).

The automatic trimming can be customized via these three properties:

Property | Default | Description
-------- | ------- | -----------
`minUsagePercentage` | `25` | The minimum `size / capacity` percentage allowed before auto trimming.
`maxOverlaidPercentage` | `50` | The maximum `overlaySize / capacity´ percentage allowed before auto trimming.
`usePadding` | `true` | Indicates whether to use padding or not.

There exists three predefined auto trimming modes:
  * `AutoTrimming.Always` will use auto trimming for all transformation operations and never use an overlay nor any padding. Safe but slow.
  * `AutoTrimming.Default` will use auto trimming for all transformation operations where the size of the returned instance is at most 25% of the capacity of the underlay, and will use an overlay with a size of at most 50% of the capacity of the underlay, and will use padding. Somewhat leaky but fast except for creating small slices.
 * `AutoTrimming.Never` will never use auto trimming, and will allow the overlay to grow to the same size as the capacity of the underlay, and will use padding. Leaky but fast.

**Examples**
```scala
// expression                                              (array, start, stop)
ArraySeq()                                             ==> ([], 0, 0)
ArraySeq() :+ a                                        ==> ([a, _, _, _, _, _, _, _], 0, 1)
a +: ArraySeq()                                        ==> ([_, _, _, _, _, _, _, a], -1, 0)
ArraySeq() :++ Seq(a, b, c, d)                         ==> ([a, b, c, d, _, _, _, _], 0, 4)
Seq(a, b, c, d) ++: ArraySeq()                         ==> ([_, _, _, _, a, b, c, d], -4, 0)
Seq(a, b, c, d) ++: ArraySeq() :++ Seq(e, f, g, h)     ==> ([e, f, g, h, a, b, c, d], -4, 4)
ArraySeq(a, b, c, d, e, f, g, h)                       ==> ([a, b, c, d, e, f, g, h], 0, 8)
ArraySeq() :++ Seq(a, b, c, d, e, f, g, h)             ==> ([a, b, c, d, e, f, g, h], 0, 8)
Seq(a, b, c, d, e, f, g, h) ++: ArraySeq()             ==> ([a, b, c, d, e, f, g, h], -8, 0)
(Seq(a, b, c, d) ++: ArraySeq()).slice(1, 3)           ==> ([_, _, _, _, a, b, c, d], -3, -1)
(ArraySeq() :++ Seq(a, b, c, d)).slice(1, 3)           ==> ([a, b, c, d, _, _, _, _], 1, 3)
(Seq(a, b) ++: ArraySeq() :++ Seq(c, d)).slice(1, 3)   ==> ([c, d, _, _, _, _, a, b], -1, 1)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ For more information, see the [CONTRIBUTING](CONTRIBUTING.md) file.
 - [ ] `EqSet`
 - [x] `BitSet`
 - [x] `View`
+- [x] `ArraySeq` (new immutable indexed sequence)
 
 ## Implemented operations (on the relevant collection types)
 
@@ -216,10 +217,11 @@ For more information, see the [CONTRIBUTING](CONTRIBUTING.md) file.
 
 - [x] `combinations`
 - [x] `updated`
-- [x] `prepend`
-- [x] `append`
+- [x] `+:` / `prepend`
+- [x] `:+` / `append`
 - [x] `++` / `concat` / `union`
 - [x] `++:` / `prependAll`
+- [x] `:++` / `appendAll`
 - [x] `flatMap`
 - [x] `grouped`
 - [x] `keys` / `keySet` / `keysIterator`
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/ImmutableArrayBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/ImmutableArrayBenchmark.scala
@@ -1,4 +1,6 @@
-package strawman.collection.immutable
+package strawman
+package collection
+package immutable
 
 import java.util.concurrent.TimeUnit
 
@@ -6,7 +8,7 @@ import org.openjdk.jmh.annotations._
 import org.openjdk.jmh.infra.Blackhole
 
 import scala.{Any, AnyRef, Int, Long, Unit, math}
-import scala.Predef.intWrapper
+import scala.Predef.{intWrapper, longWrapper}
 
 @BenchmarkMode(scala.Array(Mode.AverageTime))
 @OutputTimeUnit(TimeUnit.NANOSECONDS)
@@ -15,27 +17,98 @@ import scala.Predef.intWrapper
 @Measurement(iterations = 12)
 @State(Scope.Benchmark)
 class ImmutableArrayBenchmark {
-  @Param(scala.Array("0", "1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "4096", "131070", "7312102"))
+  //@Param(scala.Array("0", "1"/*, "2", "3", "4"*/, "7", "8"/*, "15"*/, "16", "17", "39"/*, "282", "4096", "131070", "7312102"*/))
+  @Param(scala.Array(/*"0", */"1"/*, "2", "3", "4", "7"*/, "8"/*, "15", "16"*/, "17"/*, "39"*/, "282", "4096", "73121", "7312102"))
   var size: Int = _
 
+  val random = new scala.util.Random(19740115L)
   var xs: ImmutableArray[Long] = _
+  var ys: ImmutableArray[Long] = _
   var zs: ImmutableArray[Long] = _
   var zipped: ImmutableArray[(Long, Long)] = _
   var randomIndices: scala.Array[Int] = _
   def fresh(n: Int) = ImmutableArray((1 to n).map(_.toLong): _*)
+  def freshBuilder() = ImmutableArray.newBuilder[Long]()
 
   @Setup(Level.Trial)
   def initTrial(): Unit = {
     xs = fresh(size)
+    val xsl = xs.splitAt(size / 2)._1
+    ys = xsl ++ xsl.reverse
     zs = fresh((size / 1000) max 2).map(-_)
     zipped = xs.map(x => (x, x))
     if (size > 0) {
-      randomIndices = scala.Array.fill(1000)(scala.util.Random.nextInt(size))
+      randomIndices = scala.Array.fill(1000)(random.nextInt(size))
     }
   }
 
   @Benchmark
-  def create(bh: Blackhole): Unit = bh.consume(fresh(size))
+  def create_apply(bh: Blackhole): Unit = bh.consume(fresh(size))
+
+  @Benchmark
+  def create_build(bh: Blackhole): Unit = {
+    var i = 0L
+    val builder = freshBuilder()
+    builder.sizeHint(size)
+    while (i < size) {
+      builder += i
+      i += 1
+    }
+    bh.consume(builder.result())
+  }
+
+  @Benchmark
+  def create_buildRange(bh: Blackhole): Unit = {
+    var i = 0L
+    val builder = freshBuilder()
+    builder.sizeHint(size * 10)
+    while (i < size) {
+      builder ++= xs.iterableFactory.range(i - 10, i)
+      i += 1
+    }
+    bh.consume(builder.result())
+  }
+
+  @Benchmark
+  def expand_foldAppend(bh: Blackhole): Unit = {
+    val ys = (1L to size).foldLeft(xs.slice(0, 0)) {
+      case (acc, i) =>
+        //bh.consume(acc :+ 42L)
+        acc :+ i
+    }
+    bh.consume(ys)
+  }
+
+  @Benchmark
+  def expand_foldAppendDouble(bh: Blackhole): Unit = {
+    val ys = (1L to size).foldLeft(xs.slice(0, 0)) {
+      case (acc, i) =>
+        bh.consume(acc :+ 42L)
+        acc :+ i
+    }
+    bh.consume(ys)
+  }
+
+  @Benchmark
+  def expand_foldAppendDoubleForeach(bh: Blackhole): Unit = {
+    val ys = (1L to size).foldLeft(xs.slice(0, 0)) {
+      case (acc, i) =>
+        bh.consume(acc :+ 42L)
+        acc :+ i
+    }
+    ys.foreach(bh.consume)
+  }
+
+  @Benchmark
+  def extract_palindrome(bh: Blackhole): Unit = {
+    def isPalindrome[A](xs: Seq[A]): Boolean = {
+      xs match {
+        case first +: middle :+ last => first == last && isPalindrome(middle)
+        case _ => true
+      }
+    }
+    bh.consume(isPalindrome(ys))
+  }
 
   @Benchmark
   @OperationsPerInvocation(1000)
@@ -138,7 +211,7 @@ class ImmutableArrayBenchmark {
   }
 
   @Benchmark
-  def expand_padTo(bh: Blackhole): Unit = bh.consume(xs.padTo(size * 2, 42))
+  def expand_padTo(bh: Blackhole): Unit = bh.consume(xs.padTo(size * 2, 42L))
 
   @Benchmark
   def traverse_foreach(bh: Blackhole): Unit = xs.foreach(x => bh.consume(x))
@@ -170,17 +243,17 @@ class ImmutableArrayBenchmark {
   }
 
   @Benchmark
-  def traverse_foldLeft(bh: Blackhole): Unit = bh.consume(xs.foldLeft(0) {
+  def traverse_foldLeft(bh: Blackhole): Unit = bh.consume(xs.foldLeft(0L) {
     case (acc, n) =>
       bh.consume(n)
-      acc + 1
+      acc + n
   })
 
   @Benchmark
-  def traverse_foldRight(bh: Blackhole): Unit = bh.consume(xs.foldRight(0) {
+  def traverse_foldRight(bh: Blackhole): Unit = bh.consume(xs.foldRight(0L) {
     case (n, acc) =>
       bh.consume(n)
-      acc - 1
+      acc - n
   })
 
   @Benchmark
@@ -222,7 +295,7 @@ class ImmutableArrayBenchmark {
   @Benchmark
   @OperationsPerInvocation(1000)
   def transform_updateLast(bh: Blackhole): Unit = {
-    var i = 0
+    var i = 0L
     while (i < 1000) {
       bh.consume(xs.updated(size - 1, i))
       i += 1
@@ -231,12 +304,28 @@ class ImmutableArrayBenchmark {
 
   @Benchmark
   @OperationsPerInvocation(1000)
-  def transform_updateRandom(bh: Blackhole): Unit = {
+  def transform_updateForeach(bh: Blackhole): Unit = {
+    var ys = xs
     var i = 0
     while (i < 1000) {
-      bh.consume(xs.updated(randomIndices(i), i))
+      ys = ys.updated(randomIndices(i), i.toLong)
+      bh.consume(ys)
       i += 1
     }
+    ys.foreach(bh.consume)
+  }
+
+  @Benchmark
+  @OperationsPerInvocation(1000)
+  def transform_updateRandom(bh: Blackhole): Unit = {
+    var ys = xs
+    var i = 0
+    while (i < 1000) {
+      ys = ys.updated(randomIndices(i), i.toLong)
+      bh.consume(ys)
+      i += 1
+    }
+    bh.consume(ys)
   }
 
   @Benchmark
@@ -256,17 +345,30 @@ class ImmutableArrayBenchmark {
   def transform_distinct(bh: Blackhole): Unit = bh.consume(xs.distinct)
 
   @Benchmark
-  def transform_distinctBy(bh: Blackhole): Unit = bh.consume(xs.distinctBy(_ % 2))
+  def transform_distinctBy(bh: Blackhole): Unit = bh.consume(xs.distinctBy(_ % 2L))
 
   @Benchmark
-  def transform_map(bh: Blackhole): Unit = bh.consume(xs.map(x => x + 1))
+  def transform_map(bh: Blackhole): Unit = bh.consume(xs.map((n: Long) => n + 1L))
+
+  @Benchmark
+  def transform_collect(bh: Blackhole): Unit = bh.consume(xs.collect { case n if n % 5L == 0L => n })
+
+  @Benchmark
+  def transform_flatMap(bh: Blackhole): Unit = bh.consume(xs.flatMap {
+    case n if n % 3L == 0L => List(n, -n)
+    case n if n == size => List.range(1L, n)
+    case _ => Nil
+  })
+
+  @Benchmark
+  def transform_filter(bh: Blackhole): Unit = bh.consume(xs.filter(_ % 5L == 0L))
 
   @Benchmark
   @OperationsPerInvocation(100)
   def transform_span(bh: Blackhole): Unit = {
     var i = 0
     while (i < 100) {
-      val (xs1, xs2) = xs.span(x => x < randomIndices(i))
+      val (xs1, xs2) = xs.span(x => x < randomIndices(i).toLong)
       bh.consume(xs1)
       bh.consume(xs2)
       i += 1
@@ -296,7 +398,7 @@ class ImmutableArrayBenchmark {
 
   @Benchmark
   def transform_groupBy(bh: Blackhole): Unit = {
-    val result = xs.groupBy(_ % 5)
+    val result = xs.groupBy(_ % 5L)
     bh.consume(result)
   }
 }
--- a/test/junit/src/test/scala/strawman/collection/mutable/ArraySeqTest.scala
+++ b/test/junit/src/test/scala/strawman/collection/mutable/ArraySeqTest.scala
@@ -1,0 +1,44 @@
+package strawman.collection.mutable
+
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+import strawman.collection.immutable.{List, ArraySeq}
+
+@RunWith(classOf[JUnit4])
+/* Test for scala/bug#8014 and ++ in general  */
+class ArraySeqTest {
+  val noArraySeq = ArraySeq.empty[Int]
+  val xArraySeq = ArraySeq(1, 2, 3)
+  val smallArraySeq = ArraySeq.range(0,3)
+  val bigArraySeq = ArraySeq.range(0,64)
+  val smsm = ArraySeq.tabulate(2 * smallArraySeq.length)(i => (i % smallArraySeq.length))
+  val smbig = ArraySeq.tabulate(smallArraySeq.length + bigArraySeq.length)(i =>
+    if (i < smallArraySeq.length) i else i - smallArraySeq.length
+  )
+  val bigsm = ArraySeq.tabulate(smallArraySeq.length + bigArraySeq.length)(i =>
+    if (i < bigArraySeq.length) i else i - bigArraySeq.length
+  )
+  val bigbig = ArraySeq.tabulate(2 * bigArraySeq.length)(i => (i % bigArraySeq.length))
+
+
+  val arraySeqs = List(noArraySeq, smallArraySeq, bigArraySeq)
+  val ans = List(
+    arraySeqs,
+    List(smallArraySeq, smsm, smbig),
+    List(bigArraySeq, bigsm, bigbig)
+  )
+
+  @Test
+  def ArraySeqCat(): Unit = {
+    val cats = arraySeqs.map(a => arraySeqs.map(a ++ _))
+    assert( cats == ans )
+  }
+
+  @Test
+  def arrayCat(): Unit = {
+    val ars = arraySeqs.map(_.toArray)
+    val cats = arraySeqs.map(a => ars.map(a ++ _))
+    assert( cats == ans )
+  }
+}
--- a/benchmarks/memory/src/main/scala/strawman/collection/MemoryFootprint.scala
+++ b/benchmarks/memory/src/main/scala/strawman/collection/MemoryFootprint.scala
@@ -1,10 +1,10 @@
 package bench
 
-import strawman.collection.immutable.{LazyList, List, Range, NumericRange, Vector}
+import strawman.collection.immutable.{LazyList, List, Range, NumericRange, Vector, ImmutableArray, ArraySeq}
 import strawman.collection.mutable.{ArrayBuffer, ListBuffer}
 
 import scala.{Any, AnyRef, App, Int, Long, Seq, StringContext}
-import scala.Predef.{ArrowAssoc, println, intWrapper}
+import scala.Predef.{ArrowAssoc, intWrapper}
 import scala.compat.Platform
 import java.lang.Runtime
 import java.nio.file.{Files, Paths}
@@ -48,11 +48,12 @@ object MemoryFootprint extends App {
       "TreeSet"       -> benchmark(n => strawman.collection.immutable.TreeSet((1 to n).map(_.toString): _*)),
       "ArrayBuffer"   -> benchmark(ArrayBuffer.fill(_)(obj)),
       "ListBuffer"    -> benchmark(ListBuffer.fill(_)(obj)),
-      "ImmutableArray" -> benchmark(strawman.collection.immutable.ImmutableArray.fill(_)(obj)),
-      "ImmutableArray (primitive)" -> benchmark(strawman.collection.immutable.ImmutableArray.fill(_)(123)),
+      "ImmutableArray" -> benchmark(ImmutableArray.fill(_)(obj)),
+      "scala.Array"   -> benchmark(scala.Array.fill(_)(obj)),
       "Range"         -> benchmark(Range(0, _)),
-      "NumericRange"  -> benchmark(NumericRange(0, _, 1))
-    )
+      "NumericRange"  -> benchmark(NumericRange(0, _, 1)),
+      "ArraySeq"      -> benchmark(ArraySeq.fill(_)(obj))
+  )
 
   // We use a format similar to the one used by JMH so that
   // our charts can be generated in the same way
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
@@ -1,4 +1,6 @@
-package strawman.collection.immutable
+package strawman
+package collection
+package immutable
 
 import java.util.concurrent.TimeUnit
 
@@ -15,8 +17,7 @@ import scala.Predef.intWrapper
 @Measurement(iterations = 15)
 @State(Scope.Benchmark)
 class ArrayBaselineBenchmark {
-
-  @Param(scala.Array("39", "282", "73121", "7312102"))
+  @Param(scala.Array("39", "282", "4096", "73121", "7312102"))
   var size: Int = _
 
   @Param(scala.Array("39"))
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/ArraySeq_AlwaysBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/ArraySeq_AlwaysBenchmark.scala
@@ -3,12 +3,11 @@ package collection
 package immutable
 
 import java.util.concurrent.TimeUnit
-
 import org.openjdk.jmh.annotations._
 import org.openjdk.jmh.infra.Blackhole
-
-import scala.{Any, AnyRef, Int, Long, Unit, math}
-import scala.Predef.intWrapper
+import scala.{math, Any, AnyRef, Int, Long, Unit}
+import scala.Predef.{intWrapper, longWrapper}
+import strawman.collection.immutable.ArraySeq.AutoTrimming
 
 @BenchmarkMode(scala.Array(Mode.AverageTime))
 @OutputTimeUnit(TimeUnit.NANOSECONDS)
@@ -16,17 +15,19 @@ import scala.Predef.intWrapper
 @Warmup(iterations = 12)
 @Measurement(iterations = 12)
 @State(Scope.Benchmark)
-class LazyListBenchmark {
+class ArraySeq_AlwaysBenchmark {
   //@Param(scala.Array("0", "1"/*, "2", "3", "4"*/, "7", "8"/*, "15"*/, "16", "17", "39"/*, "282", "4096", "131070", "7312102"*/))
   @Param(scala.Array(/*"0", */"1"/*, "2", "3", "4", "7"*/, "8"/*, "15", "16"*/, "17"/*, "39"*/, "282", "4096", "73121", "7312102"))
   var size: Int = _
 
-  var xs: LazyList[Long] = _
-  var ys: LazyList[Long] = _
-  var zs: LazyList[Long] = _
-  var zipped: LazyList[(Long, Long)] = _
+  val random = new scala.util.Random(19740115L)
+  var xs: ArraySeq[Long] = _
+  var ys: ArraySeq[Long] = _
+  var zs: ArraySeq[Long] = _
+  var zipped: ArraySeq[(Long, Long)] = _
   var randomIndices: scala.Array[Int] = _
-  def fresh(n: Int) = LazyList((1 to n).map(_.toLong): _*)
+  def fresh(n: Int) = ArraySeq((1 to n).map(_.toLong): _*).withAutoTrimming(AutoTrimming.Always)
+  def freshBuilder() = ArraySeq.newBuilder[Long](autoTrimming = AutoTrimming.Always)
 
   @Setup(Level.Trial)
   def initTrial(): Unit = {
@@ -36,7 +37,7 @@ class LazyListBenchmark {
     zs = fresh((size / 1000) max 2).map(-_)
     zipped = xs.map(x => (x, x))
     if (size > 0) {
-      randomIndices = scala.Array.fill(1000)(scala.util.Random.nextInt(size))
+      randomIndices = scala.Array.fill(1000)(random.nextInt(size))
     }
   }
 
@@ -46,12 +47,55 @@ class LazyListBenchmark {
   @Benchmark
   def create_build(bh: Blackhole): Unit = {
     var i = 0L
-    val builder = xs.iterableFactory.newBuilder[Long]()
+    val builder = freshBuilder()
+    builder.sizeHint(size)
     while (i < size) {
       builder += i
       i += 1
     }
     bh.consume(builder.result())
+  }
+
+  @Benchmark
+  def create_buildRange(bh: Blackhole): Unit = {
+    var i = 0L
+    val builder = freshBuilder()
+    builder.sizeHint(size * 10)
+    while (i < size) {
+      builder ++= xs.iterableFactory.range(i - 10, i)
+      i += 1
+    }
+    bh.consume(builder.result())
+  }
+
+  @Benchmark
+  def expand_foldAppend(bh: Blackhole): Unit = {
+    val ys = (1L to size).foldLeft(xs.slice(0, 0)) {
+      case (acc, i) =>
+        //bh.consume(acc :+ 42L)
+        acc :+ i
+    }
+    bh.consume(ys)
+  }
+
+  @Benchmark
+  def expand_foldAppendDouble(bh: Blackhole): Unit = {
+    val ys = (1L to size).foldLeft(xs.slice(0, 0)) {
+      case (acc, i) =>
+        bh.consume(acc :+ 42L)
+        acc :+ i
+    }
+    bh.consume(ys)
+  }
+
+  @Benchmark
+  def expand_foldAppendDoubleForeach(bh: Blackhole): Unit = {
+    val ys = (1L to size).foldLeft(xs.slice(0, 0)) {
+      case (acc, i) =>
+        bh.consume(acc :+ 42L)
+        acc :+ i
+    }
+    ys.foreach(bh.consume)
   }
 
   @Benchmark
@@ -68,10 +112,11 @@ class LazyListBenchmark {
   @Benchmark
   @OperationsPerInvocation(1000)
   def expand_prepend(bh: Blackhole): Unit = {
+    xs.reset()
     var ys = xs
     var i = 0L
     while (i < 1000) {
-      ys = i #:: ys
+      ys = i +: ys
       i += 1
     }
     bh.consume(ys)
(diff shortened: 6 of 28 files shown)


src/main/scala/strawman/collection/immutable/Spandex.scala
```diff
@@ -0,0 +1,247 @@
...
+  def +:[B >: A](b: B): Spandex[B]
+
+  def :+[B >: A](b: B): Spandex[B]
+
+  override def ++[B >: A](xs: IterableOnce[B]): Spandex[B] = xs match {
+    case bs: Spandex[B] =>
+      Spandex.tabulate(length + bs.length) {
+        case i if i < length => apply(i)
+        case i => bs(i - length)
+      }
```
julienrf: You could probably have even better performance by using `System.arraycopy` here.

src/main/scala/strawman/collection/immutable/Spandex.scala
```diff
@@ -0,0 +1,247 @@
...
+    }
+  }
+
+  private[immutable] final object Empty extends Primary[Nothing](0, 0, Array.empty) {
+    override def +:[B >: Nothing](b: B): Spandex[B] = Spandex(b)
+    override def :+[B >: Nothing](b: B): Spandex[B] = Spandex(b)
+    override def ++[B >: Nothing](xs: IterableOnce[B]): Spandex[B] = {
+      val builder = Spandex.newBuilder[B]
+      builder ++= xs
+      Spandex.fromIterable(builder.result)
```
julienrf: This is not the first time we face this situation: maybe our `fromIterable` method should be a `fromIterableOnce`, so that we avoid using the builder. Is there an issue in doing that?

julienrf (commented):
That looks very interesting. Could you run the benchmarks with `immutable.Array` so that we can compare them?

src/main/scala/strawman/collection/immutable/Spandex.scala
odd: Switched to using two `Array.copy` instead.

odd:
Do you mean running the benchmarks for `immutable.Array` at the same time or independently and comparing the results? (The immutable-arrays branch does not seem to be updated to the latest version of master so running both benchmarks at the same time seems difficult at the moment)

DarkDimius:
I'd argue that you should compare against current Vector instead.
I can't think of a non-sythetic situation when I'd prefer immutable.Array over current vector.

This is because I expect most of the operations that showcase immutability(drop, tail, filter) to be substantially better in Vector than in immutable array that was proposed.
If I don't care about those operations, than I don't care about immutability and I would use Array instead. Between the Vector and mutable array I don't see a convincing performance argument for immutable array.

Ichoran:
Did you run the benchmarks?  If yes, can you post the results here?

odd:
They are running right now but will not be finished until later tonight so I'll upload the results tomorrow morning.

odd:
Here are the benchmark results (with Vector and scala.Array included).

The result part sorted by (benchmark, size, score, error):
[time-benchmark-results.txt](https://github.com/scala/collection-strawman/files/878612/time-benchmark-results.txt)

The full log:
[time-benchmark-full.txt](https://github.com/scala/collection-strawman/files/878604/time-benchmark-full.txt)

jvican:
Are you making sure that contributors follow good practices while running benchmarks? Things like cpu governor, cpu frequency, Turbo boost, etc can greatly affect the outcome of benchmarks.
  
I suggest that if these are not documented, you create a section in CONTRIBUTING that describe them. These benchmarks will probably need to be double-checked manually by a core maintainer of this repo.

DarkDimius:
@jvican, TBH, I don't think we should require contributors to provide us clean benchmarking results.
It's too much of effort that may require custom setup of hardware. 
I think better strategy would be for one of us to do it once, after we have all collections contributed.
In particular, I have quite some considerations about the current benchmarking suite as it creates an unrealistic "everything is in cache" situation in most runs.

jvican:
@DarkDimius yes, that makes sense. Automating all this infrastructure and being able to get reliable results from CI could also be an option. AFAIK, we have a dedicated machine for performance that we could tweak and use to get predictable benchmarks, in the same spirit as https://github.com/scala/scala-dev/issues/338.

julienrf:
> I have quite some considerations about the current benchmarking suite as it creates an unrealistic "everything is in cache" situation in most runs.

I would be very happy to revise the benchmarking suite, can you create an issue and detail all the points that you think should be changed?

odd:
I have generated some graphs manually to better see the relative performance of the benchmarks I posted earlier:
[time-benchmarks-results-charts.pdf](https://github.com/scala/collection-strawman/files/879057/time-benchmarks-results-charts.pdf)

Here is the pivot table the charts are based on:
[time-benchmarks-results-table.pdf](https://github.com/scala/collection-strawman/files/879058/time-benchmarks-results-table.pdf)

odd:
I've added some alternative implementations in 6f60304 that use atomic values to guard the low and high primary indices during `prepend` and `append` (instead of using `synchronized` as is done in this pull request): 
1) `Spandex_AtomicIntegers` uses two `AtomicInteger` values to represent low and high.
2) `Spandex_AtomicIntegerArray` uses a single `AtomicIntegerArray` with two slots for low and high.
3) `Spandex_AtomicLong` uses a single `AtomicLong` to represent both low and high.    
4) `Spandex_Synchronize` is the same implementation as in this pull request and uses `this.synchronize` to guard both low and high.  

I will try to run the benchmarks (also including the improvements suggested by @DarkDimius above) during the night and post the results here tomorrow.

odd:
The difference between the different atomic implementations where not significantly different in my benchmark run so I switched from the current synchronization based implementation to the alternative using two atomic integers since that is the simplest of the atomic variants (commited in e85b046).

szeiger:
I haven't looked at the implementation details yet so I don't know if this is feasible, but can a Spandex be created in a degenerate state (that would be rebalanced when needed) from an existing array without copying? We need immutable arrays for varargs. If Spandex is universally better we could keep them hidden as an implementation detail but if a Spandex can also be built with similarly low overhead, we wouldn't need immutable arrays at all.

odd:
@szeiger I added a factory method that wraps a given array without copying (in commit 3a7ea7f).

src/main/scala/strawman/collection/immutable/Spandex.scala
szeiger: It's a bit odd currently that `Iterable` is required if you want multiple traversals but also gives you all the extra methods. The only reason I can think of why we don't want `Builder <: Iterable` is that it shouldn't get those methods. OTOH, I can't imagine a useful `fromIterable` that requires multiple traversals, either.

odd:
Updated benchmarking results as of last night (`charts -f 1 -bm avgt -bs 1 -w 100ms -r 300ms -to 10s -t 1 -wi 7 -i 7`)

Graphs:
[time-benchmark-run-21-graphs.pdf](https://github.com/scala/collection-strawman/files/927970/time-benchmark-run-21-graphs.pdf)

Table:
[time-benchmark-run-21-table.pdf](https://github.com/scala/collection-strawman/files/927979/time-benchmark-run-21-table.pdf)

Log:
[time-benchmark-run-21.txt](https://github.com/scala/collection-strawman/files/927976/time-benchmark-run-21.txt)

julienrf:
Hi @odd, thanks a lot for continuing on that (even in absence of feedback from us). I really think this contribution is interesting. We first focus on getting the right design for the collections, and then we’ll see (maybe in June) which collection implementations we put in the core…

odd:
Hi @julienrf, no problem, it's only prudent to get the design sorted out before looking too much into different implementations and optimizations. Please let me know if there is anything I can do to help besides keeping this PR up to date with master.

scottcarey:
I haven't looked at details, so please excuse me if this question is not relevant:

`Expansion occurs when the underlying array is full on the effected side`

Is it possible for the array to 'wrap around' like a circular buffer, as done with the mutable.ArrayDeque ?  https://github.com/scala/collection-strawman/pull/49

If so, expansion would only be needed after all slots are full, rather than either side, since the 'edge' would move from index 0 as necessary.  Additionally, the 'biasing' of the origin from the center when re-sized would not be necessary, so the overall complexity might not increase.

This requires tracking more variables (as one needs the 'origin' index, as well as the start and end), which might break how the atomic integers are used to guard expansion.   However, one possibility is to pack two atomic integers into a single atomic long in order to hold more information atomically.

odd:
@scottcarey Thanks for the feedback and sorry for not answering sooner. 

Having the indexes wrap around the buffer (i.e. make it a circular buffer) would indeed allow more elements to be inserted before buffer expansion is required (so it would improve memory efficiency). The potential downside is that the overall complexity would increase (as you point out), but also that it would require two array copy operations instead of one at buffer expansion time (if low > high). It would also affect linear iteration negatively (instead of doing one low to high pass over the array, it could result in two non overlapping passes: low to end, 0 to high). As for packing the low and high marks into a single atomic long, this was tried earlier but the performance was not improved compared to using two atomic integers (using a single atomic long does however make the expansion at one end coupled to the expansion at the other end so that two simultaneous call to append and prepend would make one of them fail which is not the case now when using two separate atomic integers).

This is mainly just speculation on my part so preferably an alternative version using circular buffers could be implemented and benchmarked against the current one 🤓 🥇

NthPortal:
I noticed that the iterators returned by `Spandex.iterator()` do not do bounds checks.
I've written a basic fix [here](https://github.com/odd/collection-strawman/pull/1).

odd:
Time benchmark graphs for the latest version (2a840d8) comparing `ImmutableArray`, `Vector` and `Spandex`: [time-benchmarks-run-28-charts.pdf](https://github.com/scala/collection-strawman/files/1128097/time-benchmarks-run-28-charts.pdf)

julienrf:
Thanks @odd :)

The numbers are very good for Spandex! BTW, where does the name come from?

odd:
I liked the definition in Wikipedia and it seemed to be unused in this
domain:

​*Spandex, Lycra or elastane is a synthetic fiber
<https://en.wikipedia.org/wiki/Synthetic_fiber> known for its
exceptional elasticity
<https://en.wikipedia.org/wiki/Elasticity_(physics)>. It is stronger and
more durable than natural rubber <https://en.wikipedia.org/wiki/Rubber>.​*
*...*
*The name "spandex" is an anagram of the word "expands".*


*​*


On Thu, Jul 6, 2017 at 4:16 PM, Julien Richard-Foy <notifications@github.com

benchmarks/time/src/main/scala/strawman/collection/immutable/ImmutableArrayBenchmark.scala
```diff
@@ -88,9 +113,38 @@ class ImmutableArrayBenchmark {
   def map(bh: Blackhole): Unit = bh.consume(xs.map(x => x + 1))
 
   @Benchmark
+  def reverse(bh: Blackhole): Any = bh.consume(xs.reverse)
```
scottcarey: The benchmarks are inconsistent.  Some return Any, and some Unit.   This could affect the JVM optimizations.

Ideally the same benchmark suite is used for each of these as appropriate, only changing the underlying collection type each run.  `@Benchmark def reverse`, etc,  should not be written more than once.

Also, I am not sure why the form used here:

````scala
@Benchmark
def reverse(bs: Blackhole): Any = bh.consume(xs.reverse)
````

is preferred to 

````scala
@Benchmark
def reverse(): C = xs.reverse
````

(where `C` is the type of `xs`, though `Any` might do)

My understanding of JMH is that any value returned by the benchmark method is effectively blackhole'd anyway.  Obviously, in cases like foldLeft below we need the Blackhole to consume items that are not returned.

Many of the things above are not specific to this PR, but IMO may be affecting the sanity of the benchmark results.
Thoughts?

**Comment by scottcarey** [comment]
It is curious that foreach_filter does worse relative to ImmutableArray, when Spandex does better on the individual components (head, tail), maybe nonEmpty is the culprit?

In addition to benchmarks, It would be useful to have a chart of the heap used as size increases, per data structure.  This probably won't be all that flattering for Vector either.

**Comment by odd** [comment]
Here is a chart for the memory benchmarks comparing `ImmutableArray`, `Vector`, `scala.Array `and `Spandex`: 

memory-footprint

File: benchmarks/time/src/main/scala/strawman/collection/immutable/ImmutableArrayBenchmark.scala (hunk 3 above)
Comment (odd) [inline benchmarks/time/src/main/scala/strawman/collection/immutable/ImmutableArrayBenchmark.scala]: Yeah, one benchmark for each of the main abstractions (Seq, IndexedSeq, LinearSeq, Set, SortedSet etc) that could be extended once for each implementation would be super nice. I don't know if this negatively affects the quality of the results though (as the benchmarks will link towards the various abstract classes instead of the implementation class directly; on the other hand that is probably how they would be  used in the wild).

**Comment by scottcarey** [comment]
What if we do sizes of the form 2^n - 1 ?    Powers of two are cheating :)   The real-world average will be closer to 2^n * 3/2.

An interesting graph would be 2^n -1, 2^n, (2^n * 3/2), for n = 1, 3, 5, 7, ... 19

**Comment by Ichoran** [comment]
These results look great!  We should try to find *any* area in which `Vector` substantially outperforms `Spandex`.  If we can't find any, I suggest we just drop `Vector` and call `Spandex` `Vector` instead.

**Comment by scottcarey** [comment]
@Ichoran There are a lot of use cases where Vector absolutely _destroys_ Spandex.  They should be added as benchmarks.   Cherry-picking power of two sizes and read only / append / prepend use cases is the best possible situation for Spandex and similar 'single array backed' data structures.

To start with, `updated`.  Vector will be O(log32(n)), spandex will be O(n).
Vector can potentially optimize filter in cases where few things are filtered out, but I'm not sure if it does.   For example, filter out 1 in 1000 things from a large vector, and most of the tree could remain intact.  Spandex could be optimized to aggregate 'slices' of the backing array into a filtered view, but that has other downsides.

Composite patterns like  'tail then prepend' will suddenly look bad, when individually both tail and prepend are fast.  There should be benchmarks for `a +: seq.tail` and 'init :+ a`. 

Now lets get extreme.

Start with a Vector an a Spandex that both have 1 million elements.

Now decompose each via head/tail until it has one element left, and hold onto that last one-element sequence.  Spandex will still be indirectly referencing all 1 million elements in the backing array.  Vector will reference one, leaving the rest to be GC'd.  For Spandex, 'slice' is just a smaller view of the backing array, and many operations are based off of slice.

Vector is doing many things to prevent memory leaks that are not done here.  Many optimizations were proposed to make Vector faster at appending/prepending/tail etc, but were turned down because they would hold onto references to items that are not needed by the collection.  If you took the same approach in Vector to appending and slicing that Spandex does, its append time would dramatically improve, tail/init/slice would be just as fast as Spandex, but references to items in the collection would leak.
It was argued that users should use VectorBuilder, instead, if they wish to have fast incremental append/prepend/tail/init/etc.

So from some use cases, the real competitor for Spandex is VectorBuilder, which was not benchmarked.   And from other perspectives, Spandex is woefully lacking as a general purpose immutable data structure as any use case that ends up in `slice` can lead to memory leaks.  It is not IMO a persistent data structure, but rather merely an immutable one.  For me at least, to be persistent the data structure has to only reference what it requires, and no more.  Immutable is a lower bar.

Spandex looks extremely valuable for some use cases and obliterates Vector for those, but a immutable singular array backed structure will never defeat a tree/trie one once element removal comes into play and dereferencing of removed items is required.

**Comment by odd** [comment]
Thanks for the detailed analysis. You make a good argument for keeping Vector as is in the collection library. I believe the more elaborate benchmarks you describe should be added to the benchmarking suite so that the performance differences between Vector, Spandex and all the other implementations comes to light. 

The goal I had when creating Spandex was I'm fact not so much to replace Vector as it was to create an alternative to the original immutable array that was added to the strawman project; an implementation that could potentially be used as the default implementation of immutable.Seq offering both fast prepend as well as append operations with as little overhead over a plain array as possible (potentially also suitable for implementing varargs). 

The issue of keeping too many dereferenced items around after having done a narrow slice or filter could perhaps be mitigated by doing an automatic trim operation if the capacity/size ratio becomes too big.

**Comment by scottcarey** [comment]
Spandex is awesome.  It has a lot of potential for a lot of use cases I care about.  I just want to avoid having it end up being used where it should not, or at least without eyes wide open.

Vector has a lot of little problems.  I hope we find something better or fix it.  It could get some attention and be optimized more for some use cases.  We might want custom instances for very small Vectors for example.  I have seen a LOT of heap dumps with single or two element vectors eating up way more memory than needed.  I have also seen it used in a lot of code where it is being built up one at a time, leading to poor performance.    Scalaz-stream prior to the fs2 re-write was littered with one at a time vector growth/shrinking and I had to write my own low level implementations of some features in some cases (like async.Queue).

At minimum, having more quality immutable.Seq implementations around with different performance trade-offs will be beneficial.

**Comment by odd** [comment]
I pushed a reimplementation of Spandex that appends elements to the front and prepends elements to the rear of the array. This way no margins are needed as in the previous implementation (thus increasing memory efficiency). Thanks to @scottcarey for the idea.

**Comment by Ichoran** [comment]
@scottcarey - Thanks for the analysis.  I had not really looked at the implementation.  I agree that in contrast to `Vector`'s pretty-good-at-everything capability, `Spandex` performs extremely well in some areas but much worse in others.  I agree that it's something that we likely want to have, but we should keep looking for a better `Vector` replacement.  `Vector` still has quite a number of drawbacks, and there are alternate data structures that deliver superior performance in most use cases.

**Comment by julienrf** [comment]
To do some progress on this topic, it would be interesting to get performance numbers also for cases where `Spandex` is worse than `Vector`.

> Vector still has quite a number of drawbacks, and there are alternate data structures that deliver superior performance in most use cases.

@Ichoran So, which one should be use?

**Comment by Ichoran** [comment]
I think we should provide both.  Vector is a better general-purpose data structure, but it's not very good for some things Spandex is quite good at (e.g. everything where it's beaten badly in benchmarks here).

**Comment by odd** [comment]
I just pushed an update that adds the `patch` and `appendAll` operations (same implementations as in PR #184 and #185). I also added the `randomUpdate`, `randomPatch` and `prependAll` benchmark (following an earlier discussion with @scottcarey of use cases where`Vector` should out perform `Spandex`). The attached benchmarking results (for `List`, `Vector` and `Spandex`) shows that `Vector` does indeed perform much better for the `update` benchmark (in which a single element position is updated). All other benchmarks including the `patch` benchmark (in which a multiple of elements are inserted in between existing element positions) show similar or better performance for `Spandex` over `Vector`. 

[time-benchmark-run-34-charts.pdf](https://github.com/scala/collection-strawman/files/1202765/time-benchmark-run-34-charts.pdf)

**Comment by odd** [comment]
Latest benchmarking run (with new benchmark names) for `ImmutableArray`, `Vector` and `Spandex`: 
[time-benchmark-run-40-charts.pdf](https://github.com/scala/collection-strawman/files/1221753/time-benchmark-run-40-charts.pdf)

**Comment by odd** [comment]
Latest benchmarking run (with `update_last` and `update_random`) for `ImmutableArray`, `Vector`, `List` and `Spandex`:
[time-benchmark-run-42-charts.pdf](https://github.com/scala/collection-strawman/files/1224872/time-benchmark-run-42-charts.pdf)

**Comment by odd** [comment]
Latest benchmark run for `ImmutableArray`, `Vector`, `Spandex_0` and `Spandex_25`: 
[time-benchmark-run-57-charts.pdf](https://github.com/scala/collection-strawman/files/1245060/time-benchmark-run-57-charts.pdf).

`Spandex_0` uses a `Spandex` configured with no auto shrinking (i.e. a shrinking percentage threshold of zero), while `Spandex_25` uses a `Spandex` configured to use the default auto shrinking threshold of 25% (thus any operation which produce a result that is at most 25% of the size of its underlying array will copy its elements to a new smaller array). The main difference in performance between them can be seen in the `slice`-benchmarks where the auto-shrinking comes into play. 

I also added two new benchmarks (`appendInit` and `prependTail`) that together with the existing `updated`-benchmarks demonstrate cases where `Vector` is clearly outperforming `Spandex` (since they force `Spandex` to make a copy of its underlying array). The related `patch`-benchmark does not show the same performance advantage for `Vector` however (probably because the variable length of the patched-in collection forces `Vector` to rebuild much more of its internal structure). The benchmarks where `Spandex` on the other hand significantly outperforms `Vector` are: `reverse`, `appendAll`/`prependAll` (and thus `prependAllAppendAll`), `init`/`tail` and `loop_headTail`/`loop_initLast`.      

Also, please note that many benchmarks include the creation time of the tested collection since they use a fresh collection each time. This is done to ensure that no unwanted sharing is possible between invocations, but it also skews the benchmarks a bit since collections that are expensive to create but offers cheaper operations will get worse results than the operation would in isolation (to mitigate the effect, those benchmarks use the `@OperationsPerInvocation` annotation to perform the operation multiple times for each fresh collection).

**Comment by odd** [comment]
Latest benchmarks using `@Setup(Level.Invocation)` to create the collection instances run for `ImmmutableArray`, `List`, `Vector` and `Spandex_N` (with auto shrink threshold percentages `N=0` and `N=25`) using the new naming with category prefixes:
[time-benchmark-run-60-charts.pdf](https://github.com/scala/collection-strawman/files/1257337/time-benchmark-run-60-charts.pdf)

The `create`-benchmark has been added to compare creation performance and the multiple `slice`-benchmarks has been replaced by a single `access_slice`-benchmark.

**Comment by oxbowlakes** [comment]
This might be a case of attempting to shut the stable door after the horse has bolted but can we consider **not calling this `Spandex`**?

It is essentially unsearchable - try looking for "spandex collection" and see what comes up.

**Comment by odd** [comment]
Thanks for the feedback @oxbowlakes. The name "Spandex" is indeed a bit problematic. While it fits this collection type if you consider the word "spandex" being an anagram of "expands", it does, however, lead to some NSFW results when googled for. Originally it was intended as a better `ImmutableArray` so perhaps if it gets accepted it could simply replace the current `ImmutableArray` using the same name. Another potential goal is to replace `List` as the default `Seq` which would probably make most usages go via `Seq` instead of this collection directly.

So to get a naming discussion started I have made a list of some potential alternative names with some pros and cons for each one (some more exotic than others).

### ImmutableArray
##### Pros
* descriptive
* noncryptic
##### Cons
* quite long
* does not signal expansiveness
* already exists in the strawman (but could be replaced without loss of functionality)

### Line
##### Pros
* short
* fits with Array
* fits with Vector
##### Cons
* too generic?
* common word / other usages
* does not signal expansiveness
* bad Google factor

### Band
##### Pros
* short
* fits with Array
* fits with Vector
* does signal expansiveness somewhat (via association with "rubber" in "rubber band") 
##### Cons
* bad Google factor (existing usage is common but not associated with this domain, however)

### Ray
##### Pros
* short
* fits with Array (which it mimics)
* fits with Vector
##### Cons
* does not signal expansiveness

### Xray
##### Pros
* short
* fits with Array (which it mimics)
* does signal expansiveness somewhat (if you can look past the radiation connotation)
##### Cons
* at least one existing Scala related OSS project does use it but with different spelling ("X-ray").

### Exray
##### Pros
* short
* fits with Array (which it mimics)
* does signal expansiveness
##### Cons
* cryptic
* hard to spell/remember?

### Expander
##### Pros
* relatively short
* signals expansiveness
##### Cons
* sounds more like an actor than a type of collection

### Expandex
##### Pros
* relatively short
* signals expansiveness
* close to current name (if that matters)
##### Cons
* a bit repetitive

### Xpandex
##### Pros
* relatively short
* unique
* signals expansiveness
* close to current name (if that matters)
##### Cons
* cryptic
* hard to spell/remember?

**Comment by oxbowlakes** [comment]
I think that your goal here should be to name things expressively. AFAICT there is not one single class in the scala collections library that is not an established or descriptive name. I'm also a bit confused why you want the name to convey "expansion"; `ArrayList` in java expands but doesn't convey such.

The structure you've designed is intended for use as an `IndexedSeq`; what's wrong with `ArraySeq`, `ArrayIndexedSeq`, or `ArrayVector`?

**Comment by julienrf** [comment]
`ArraySeq` looks good to me

**Comment by odd** [comment]
I agree, `ArraySeq` looks good, thanks for the suggestion.

File: README.md [hunk 4]
```diff
@@ -119,6 +119,7 @@ For more information, see the [CONTRIBUTING](CONTRIBUTING.md) file.
 - [ ] `EqSet`
 - [x] `BitSet`
 - [x] `View`
+- [x] `Spandex` (new immutable indexed sequence)
```
julienrf: `ArraySeq`

src/main/scala/strawman/collection/immutable/ArraySeq.scala
```diff
@@ -0,0 +1,832 @@
...
+    else if (stop < 0) (capacity + stop, capacity, 0, start)
+    else (0, 0, start, stop)
+  }
+
+  @`inline`
+  private[ArraySeq] final def distance(x: Int, y: Int): Int =
+    if (x < y) y - x
+    else x - y
+}
+
```
julienrf: The implementation looks good but I’m wondering what are the costs of having specialized implementations for almost all operation in terms of bytecode size and what are the performance boost?

src/test/scala/strawman/collection/immutable/test/ArraySeqTest.scala
```diff
@@ -0,0 +1,865 @@
...
+import org.junit.Assert._
+import org.junit.Test
+
+import scala.{Any, Exception, Int, NoSuchElementException, Nothing, Unit}
+import scala.Predef.{identity, intWrapper}
+import scala.runtime.RichInt
+import scala.util.Try
+import scala.util.control.NonFatal
+
+class ArraySeqTest {
```
julienrf: Could you move this file to the `test/junit/src/test/scala/strawman/collection/immutable/` directory?

test/junit/src/test/scala/strawman/collection/mutable/ArraySeqTest.scala
```diff
@@ -0,0 +1,44 @@
+package strawman.collection.mutable
```
julienrf: Why is it in the `mutable` package?

julienrf (commented):
Thanks @odd for this long effort!

I left a few comments but overall that looks great! The benchmarks you posted are really good, so I think we should merge it and replace `ImmutableArray` with `ArraySeq`.

To make it mergeable you’ll have to squash the commits into a single one.

src/test/scala/strawman/collection/immutable/test/ArraySeqTest.scala
odd: Sure, I only put it in a subdirectory named "test" because the other tests were also in a "test" subdirectory.

test/junit/src/test/scala/strawman/collection/mutable/ArraySeqTest.scala
odd: I copied the existing test for `Vector` that already exist in the `mutable` package.

src/main/scala/strawman/collection/immutable/ArraySeq.scala
odd: Any operation that delegates to `Iterator` has been overridden to avoid the creation of the iterator in order to get better performance.

The size of the generated classfiles are as follows (according to a manual summation):

Collection | Scala 2.12 | Scala 2.13 |  Dotty
---------- | ---------- | ---------- | -------
ArraySeq | 113 KB | 112 KB | 115 KB
Vector | 115 KB | 115 KB | 90 KB
List | 112 KB | 128 KB | 66 KB

odd:
Hi, glad to hear you like it! 🤓

When you say that I should squash this PR, do you mean that I should rebase this PR onto current master folding all but the first commit into the first or do you mean something else? (I have merged with master multiple times during the work on this PR so the task of rebasing it will require some effort, so I thought I'd ask before starting).

Here are the charts for the latest benchmarking results comparing the performance of `ArraySeq`, `List`, `ImmutableArray` and `Vector` (one for a variety of sizes and one for only small sizes):
Various sizes: [time-benchmark-run-71-charts.pdf](https://github.com/scala/collection-strawman/files/1319093/time-benchmark-run-71-charts.pdf)
Only small sizes: [time-benchmark-run-72-charts.pdf](https://github.com/scala/collection-strawman/files/1319075/time-benchmark-run-72-charts.pdf).

odd:
All commits have now been squashed into a single one: 5186c1d.

test/junit/src/test/scala/strawman/collection/mutable/ArraySeqTest.scala
julienrf: Oh that’s a mistake. I’ll fix that one :)

julienrf:
> When you say that I should squash this PR, do you mean that I should rebase this PR onto current master folding all but the first commit into the first or do you mean something else?

I see that you just did it, but for reference what I meant was to clean up the history a bit. If everything fits well into a single commit, that’s better.

alexnixon:
This is interesting - always good to see thought going into collections (and benchmarks!), so thanks Odd!

As a disclaimer, my background is vanilla Scala so I'm coming to both this and `ImmutableArray` fresh.

Firstly, on `ImmutableArray`. It seems really dangerous to me that it lives in a package alongside all the other immutable collections yet has a completely different design - it's not immutable in the "persistent, functional data structures" sense like all the collections alongside it. They support cheap updates but this does not. This shows in several places, for example `IndexedSeqOps` which it implements:

```
  /** The rest of the collection without its `n` last elements. For
    * linear, immutable collections this should avoid making a copy. */
  override def dropRight(n: Int): C = fromSpecificIterable(view.dropRight(n))
```

Perhaps it would be better if it were named `CopyOnWriteArray`, and if it didn't share a hierarchy with collections with such different performance characteristics.

Then on this `ArraySeq` proposal, I find the lack of predictability of its performance worrying. For example if I write:

`def wat(x: ArraySeq[Int]): ArraySeq[Int] = x :+ 42` 

I have no idea whether this is going to double my heap size - it entirely depends on how the caller constructed the `ArraySeq` and whether it has already had its "cheap append" used up. Personally I would take no performance optimization over one which only kicks in if the wind's blowing east.

So in general I can see the utility of a copy-on-write array for when you really need the performance -
 and perhaps it should have convenience methods like slices and O(1) reversal <sup>[1]</sup>. And I can see the benefit of `Vector` as a general-purpose collection which trades off performance<sup>[2]</sup> for cheap updates. However, I'm not yet convinced about the middle ground in which I think this collection lies.

I'm very open to being swayed by arguments though so would love to hear a response to this, particularly if I've there are any misunderstandings in the above.

<sup>[1] though I also see good arguments against these features - if a user cares so much about performance that they're using a copy-on-write array, then maybe it's best to get them as close to the metal as possible. They might care about the extra memory used by metadata tagged onto the data structure, and they might care how the array is laid out in memory, for example if they want to cheaply wrap a `ByteBuffer` around it to send to some other data-reading API</sup>

<sup>[2] I would love to see improvements to `Vector`, particularly in the memory usage of small `Vector`s</sup>

odd:
Hi @alexnixon and thanks for the feedback!

The example you provide for the lack of predictability is not specific for `ArraySeq` but rather common for most the `Seq` implementations that exists in the strawman.
```scala
def wat(x: Seq[Int]): Seq[Int] = x :+ 42
```
The function above will lead to all (or some in the case of `immutable.Vector`) elements of `x` being copied into a new structure if the next position in `x` is unavailable. The different `Seq` implementations will decide the availability of the next position something like this: 
* `immutable.List`: the next position is never available since it is occupied by `Nil` (so always make a full copy of `x`)
* `immutable.ImmutableArray`: the next position is never available since the underlying array is always full (so always make a full copy of `x`)
* `immutable.ArraySeq`: the next position is available if the next position already holds an equal element (to the element being appended) or if the underlying array is not full (otherwise make a full copy of `x`)
* `immutable.Vector`: the next position is never available (so always make a copy of the segment of `x` holding the next position and so on up the levels [segment size is 32, up to 5 levels deep])
* `mutable.ArrayBuffer`: the next position is available if the underlying array is not full at the rear (otherwise make a full copy of `x`)
* `mutable.ListBuffer`: the next position is available if `toList` has not been called on `x` (otherwise make a full copy of `x`)

alexnixon:
Thanks for the quick reply! 

I think I was unclear so let me rephrase: I cannot see another _immutable_ collection for which, when adding an element in the most efficient way for that collection, you don't know whether it will use `O(n)` extra memory. Do you know of one? If so, which?

Even if there is one, I would still be happy provided the cost is amortized to constant. I have no problem with java's `ArrayList` for example.

But for `ArraySeq` this isn't amortized and this will lead to spooky action at a distance, for example:
```
var coll = (1 to 1000000).foldLeft(ArraySeq.empty[Int]) { case (acc, i) =>
  // Try uncommenting this
  // acc :+ 42
  acc :+ i
}
```

With one `:+` operation, the `:+` is constant time. With two `:+` operations, the first is constant and the second linear, which makes this program quadratic altogether. Is there any other scala collection with performance characteristics like this?

It's a slightly contrived example but not hard to imagine it biting in the real world, particularly when the `:+` calls aren't right next to each other.

A valid counterargument would be "we only make the guns, we can't stop people blowing off their toes". But there has to be a line somewhere and I'm still concerned we're at the wrong side of it here.

odd:
Repeated appends (and prepends) on the same instance do indeed lead to full copying for the implementations based on single arrays such as `ArraySeq` and `ImmutableArray` (except for the first append/prepend which you get for free using `ArraySeq`). For `List` the performance is even worse for appends (but free for prepends). `Vector` is, in this case, the best overall performer since it has much less copying to do in both cases and for all repetitions (except for prepends when compared to `List` and the first append/prepend when compared to `ArraySeq`). Please see the attached benchmarks below.

Do I understand you correctly in that it is the difference in complexity between the first and the latter prepends/appends for `ArraySeq` that bothers you? Would you prefer `ArraySeq` instead having the worst complexity all the time like `ImmutableArray` to get predictability (at the cost of performance), or is it rather an argument against any immutable sequence implementation not having a consistently cheap way of doing repeated prepends or appends on the same instance?

Here are the benchmarks for the single and double append scenarios:
expand_singleappend
expand_doubleappend

oxbowlakes:
I think the objection here is not that this specific operation may be
inefficient under certain conditions. As you rightly point out, a single
append to a `List` has the same characteristics.

The issue is that the egregiously-bad performance is an example of an
*action at a distance*. That is, one can have some nice, efficient working
code and then innocently add a `a :+ b` (knowing that `:+` is an efficient
operation for `ArraySeq`) and end up with a performance problem. The
performance problem is potentially very difficult to spot, being not
isolated to a single operation but rather the effects of two (possibly
unrelated) lines of code which would be otherwise fine by themselves.

I agree with Alex in the sense that this will lead to painful real-world
scenarios.

Chris

julienrf:
@odd How easy will your code be specializable for primitive types?

@szeiger Could you present your concerns about replacing `ImmutableArray` with `ArraySeq`?

odd:
A while back some issues were raised by @alexnixon and @oxbowlakes concerning the performance of repeated `prepended`/`appended` operations on `ArraySeq` (as manifested in the `expand_foldAppendDouble` benchmarks). The problem was that the performance and memory use could be difficult to predict and lead to hard to find performance issues. 

The performance best case occurred when the targetted underlying array slot was unoccupied (i.e. it had never been written to) in which case the `prepended`/ `appended` operation would be a simple array store (~constant). If, however, the targetted underlying array slot was already occupied (i.e. it had already been used to store an element), the accessible part of the underlying array would first have to be copied and then updated with the new element (~linear in the size of the instance). The same copy-on-repeated-write problem also occured for the `updated` operation (but for `updated` there existed no constant best case since the targetted array slot would always be preoccupied).

To solve this performance degredation on repeated writes, I have redesigned `ArraySeq` to provide a map backed _overlay_ in addition to the array backed _underlay_. The _overlay_ maps indices to elements and is used whenever the targetted array slot in the _underlay_ is occupied. This allows the `prepended`/`appended`/`updated` operations to have the same best case (single array store) while improving the worst case to a single hash map store (~effectively constant).

But all is not rosy, however. A danger with `ArraySeq` is that there exists a risk for memory leaks since elements in the _underlay_ can become _hidden_. A _hidden_ element can not be garbage collected even though it is not referenced from the `ArraySeq` instance directly, since it is still referenced from the associated _underlay_. The hiding of elements can happen in two different ways: by _indexing_ or by _shadowing_. 

Hiding by _indexing_ occurs when a `slice` based operation (such as `tail`, `init`, `take` or `drop`) returns an instance where only a subset of the indices available in the associcated _underlay_ are included. The elements in array slots outside the index subset will be _hidden_ since they can not be accessed via the returned `ArraySeq` instance (but are still referenced by the associated _underlay_) . Hiding through _indexing_ can also occur when an element is `prepended` or `appended` to an instance created with _padding_ (i.e. having a larger capacity than size); the added element is _hidden_ from the original instance since it is outside the index range of the original.

Hiding by _shadowing_ occurs when an element is added to the _overlay_ and the element in the slot with the same index in the _underlay_ becomes _hidden_.

To mittigate the risk for memory leaks  from _hidden_ elements, _trimming_ should be used, either manually via the `trim` operation, or automatically via the `withAutoTrimming` operation. 

The auto trimming can be customized via these three properties: 
* `minUsePercentage`:  the minimum percentage the size of an returned instance can have compared to the capacity of the associtated _underlay_.  
* `maxOverlayPercentage`: the maximum percentage the size of an _overlay_ can grow to compared to the capacity of the associated _underlay_. 
* `usePadding`: if true the _underlay_ will have a capacity that is the smallest power of two greater or equal to the size of the instance.

There are three predefined auto trimming modes: 
* `AutoTrimming.Always`: This mode will always trim the result of any transformation operations and never use an overlay nor any padding. It performs like an immutable array which copies on every change. Safe but slow.
* `AutoTrimming.Default`: This mode will trim the result of any transformation operations which return an instance with a size that is at most 25% of the _underlay_ capacity. It will allow the _overlay_ to grow to 50% of the capacity of the _underlay_ and will use padding. Somewhat leaky but fast except when creating small slices.
* `AutoTrimming.Never`: This mode will never trim the result of any transformation operations and will allow the _overlay_ to grow to the same size as the _underlay_. It will use padding. Leaky but fast.

One way of using `ArraySeq` to try to minimize the problem with element hiding without sacrificing too much perfomance is to use `AutoTrimming.Always` for all instances (to ensure safety) and use the `withoutAutoTrimming[B >: A](f: ArraySeq[A] => ArraySeq[B])` method to wrap any performance critical transformations. That way the transformations will be performed without any trimming happening during the transformation steps; the resulting `ArraySeq` instance will instead be trimmed once at the end, when the transformation steps are completed. 

The current default auto-trimming mode has been chosen to provide a middle ground between safety and performance but perhaps it is better to make `AutoTrimming.Always` the default (which would make an `ArraySeq` by default behave the same as an `ImmutableArray`) and to promote the above usage pattern as a potential optimization technique?

odd:
Here are the graphs for the latest benchmark results for `ArraySeq` with the three predefined auto-trimming modes compared to `List`, `Vector` and `ImmutableArray`:

access_init
access_last
access_random
access_slice
access_tail
create_apply
create_build
create_buildrange
expand_append
expand_appendall
expand_appendinit
expand_foldappend
expand_foldappenddouble
expand_foldappenddoubleforeach
expand_padto
expand_prepend
expand_prependall
expand_prependallappendall
expand_prependappend
expand_prependtail
extract_palindrome
transform_collect
transform_distinct
transform_distinctby
transform_filter
transform_flatmap
transform_groupby
transform_lazyzip
transform_map
transform_patch
transform_reverse
transform_span
transform_unzip
transform_updateforeach
transform_updatelast
transform_updaterandom
transform_zip
transform_zipmaptupled
transform_zipwithindex
traverse_foldleft
traverse_foldright
traverse_foreach
traverse_headtail
traverse_initlast
traverse_iterator

julienrf:
Wow, thanks @odd for the hard work!

In summary, it seems that the `AutoTrimming.Default` mode provides, in general, better performance than `Vector`. However, the `AutoTrimming.Always` mode has, in general, worse performance than `Vector`. On the other hand, the `AutoTrimming.Default` mode can lead to memory leaks, whereas the `AutoTrimming.Always` mode can not.

I’m afraid the result is a collection type that is very powerful but a bit hard to master. What do other people think? (@Ichoran? @szeiger?)

@odd What do you think about implementing macro benchmarks (as suggested in #292)? You can draw inspiration from those which are in `ArrayBaselineBenchmark.scala` or implement your own.

NthPortal:
What's the status of this?

odd:
Hi @NthPortal, since the last commit pushed here, I have experimented with the following:
1. Make it behave like `ImmutableArray` (which has been renamed to `ArraySeq` in 2.13) for all types `T <: AnyRef` (i.e. never reuse but always copy the underlying array) to avoid potentially leaking references to objects when shadowing elements (by update or windowing).
2. Keep the current behavior of reusing the underlying array as often as possible for primitive types (since these can not refer to other objects and thus the memory occupied by shadowed elements is bound by the size of the underlying array).
3. Specialize all operations for primitive types to avoid as much boxing as possible.

The development work has been moved to the odd/collection-benchmark repository (which contains a copy of the benchmarking suite in this repository but updated to target Scala 2.13.0-M4) but I haven't pushed in quite a while. I had some interesting conversations with @szeiger at ScalaDays Berlin which however made me a bit unsure of the future of specialization in Dotty (at least in its current form), so I haven't pressed on; and then it was summer, and the warmest one in ages at that... ☀️

SethTisue:

PR #53, opened 2017-03-27
HashSet, ListSet and TreeSet

julienrf:
In a process of getting something usable by users as quickly as possible, I’m starting by replacing triple question marks by actual implementations.

This PR adds immutable `HashSet`, `ListSet` and `TreeSet`. I mostly copy-pasted the current implementation from 2.12.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ libraryDependencies += "ch.epfl.scala" %% "collection-strawman" % "0.2.0-SNAPSHO
 
 ## Implemented operations (on the relevant collection types)
 
-### Terminal operations
+### Operations not returning a collection
 
 - [x] `forall`
 - [x] `foreach`
@@ -89,7 +89,7 @@ libraryDependencies += "ch.epfl.scala" %% "collection-strawman" % "0.2.0-SNAPSHO
 - [x] `sum`
 - [x] `to`
 
-### Transformations keeping the same element type
+### Transformations to collections having the same element type
 
 - [x] `drop`
 - [x] `filter`
--- a/README.md
+++ b/README.md
@@ -45,3 +45,67 @@ resolvers += Resolver.sonatypeRepo("snapshots")
 
 libraryDependencies += "ch.epfl.scala" %% "collection-strawman" % "0.2.0-SNAPSHOT"
 ~~~
+
+## Implemented collection types
+
+- [x] `List`
+- [x] `LazyList` (aka `Stream` in 2.12)
+- [ ] `Queue`
+- [ ] `Stack`
+- [x] `ArrayOps`
+- [x] `StringOps`
+- [x] `ArrayBuffer`
+- [x] `ListBuffer`
+- [ ] `UnrolledBuffer`
+- [ ] `LinkedList`
+- [ ] `DoubleLinkedList`
+- [ ] `Range` / `NumericRange`
+- [ ] `HashMap`
+- [ ] `TreeMap`
+- [ ] `IntMap` / `LongMap` (?)
+- [ ] `ListMap`
+- [ ] `MultiMap`
+- [ ] `HashSet`
+- [ ] `ListSet`
+- [ ] `TreeSet`
+- [ ] `EqSet`
+- [ ] `BitSet`
+- [x] `View`
+
+## Implemented operations (on the relevant collection types)
+
+### Terminal operations
+
+- [x] `forall`
+- [x] `foreach`
+- [x] `foldLeft`
+- [x] `foldRight`
+- [x] `head`
+- [x] `indexWhere`
+- [x] `isEmpty` / `nonEmpty`
+- [x] `last`
+- [x] `mkString`
+- [x] `size`
+- [x] `sum`
+- [x] `to`
+
+### Transformations keeping the same element type
+
+- [x] `drop`
+- [x] `filter`
+- [x] `partition`
+- [x] `splitAt`
+- [x] `tail`
+- [x] `take`
+- [ ] `groupBy`
+
+### Transformations to collections that can have a different element type
+
+- [x] `++` / `concat`
+- [x] `flatMap`
+- [x] `map`
+- [x] `zip`
+
+### In-place mutating operations
+
+TODO
--- a/README.md
+++ b/README.md
@@ -65,9 +65,9 @@ libraryDependencies += "ch.epfl.scala" %% "collection-strawman" % "0.2.0-SNAPSHO
 - [ ] `IntMap` / `LongMap` (?)
 - [ ] `ListMap`
 - [ ] `MultiMap`
-- [ ] `HashSet`
-- [ ] `ListSet`
-- [ ] `TreeSet`
+- [x] `HashSet`
+- [x] `ListSet`
+- [x] `TreeSet`
 - [ ] `EqSet`
 - [ ] `BitSet`
 - [x] `View`
@@ -76,6 +76,7 @@ libraryDependencies += "ch.epfl.scala" %% "collection-strawman" % "0.2.0-SNAPSHO
 
 ### Operations not returning a collection
 
+- [x] `firstKey`
 - [x] `forall`
 - [x] `foreach`
 - [x] `foldLeft`
@@ -84,6 +85,7 @@ libraryDependencies += "ch.epfl.scala" %% "collection-strawman" % "0.2.0-SNAPSHO
 - [x] `indexWhere`
 - [x] `isEmpty` / `nonEmpty`
 - [x] `last`
+- [x] `lastKey`
 - [x] `mkString`
 - [x] `size`
 - [x] `sum`
@@ -92,8 +94,10 @@ libraryDependencies += "ch.epfl.scala" %% "collection-strawman" % "0.2.0-SNAPSHO
 ### Transformations to collections having the same element type
 
 - [x] `drop`
+- [x] `empty`
 - [x] `filter`
 - [x] `partition`
+- [x] `range`
 - [x] `splitAt`
 - [x] `tail`
 - [x] `take`
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
@@ -1,0 +1,57 @@
+package strawman.collection.immutable
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations._
+
+import scala.{Any, AnyRef, Int, Unit}
+import scala.Predef.intWrapper
+
+@BenchmarkMode(scala.Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Fork(1)
+@Warmup(iterations = 12)
+@Measurement(iterations = 12)
+@State(Scope.Benchmark)
+class HashSetBenchmark {
+
+  @Param(scala.Array("8", "64", "512", "4096", "32768", "262144"/*, "2097152"*/))
+  var size: Int = _
+
+  var xs: HashSet[AnyRef] = _
+  var obj: Any = _
+
+  @Setup(Level.Trial)
+  def initData(): Unit = {
+    xs = HashSet((1 to size).map(_.toString): _*)
+    obj = ""
+  }
+
+  @Benchmark
+  def cons(): Any = {
+    var ys = HashSet.empty[Any]
+    var i = 0
+    while (i < size) {
+      ys = ys + obj // Note: we should test different cases: colliding values as well as non-colliding values, etc.
+      i += 1
+    }
+    ys
+  }
+
+  @Benchmark
+  def uncons(): Any = xs.tail
+
+  @Benchmark
+  def concat(): Any = xs ++ xs
+
+  @Benchmark
+  def foreach(): Any = {
+    var n = 0
+    xs.foreach(x => if (x eq null) n += 1)
+    n
+  }
+
+  @Benchmark
+  def map(): Any = xs.map(x => if (x eq null) "foo" else "bar")
+
+}
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
@@ -27,10 +27,10 @@ class HashSetBenchmark {
   def initData(): Unit = {
     def freshCollection() = HashSet((1 to size).map(_.toLong): _*)
     xs = freshCollection()
-    xss = scala.Array.fill(1000)(freshCollection())
-    if (size > 0) {
-      randomIndices = scala.Array.fill(1000)(scala.util.Random.nextInt(size))
-    }
+//    xss = scala.Array.fill(1000)(freshCollection())
+//    if (size > 0) {
+//      randomIndices = scala.Array.fill(1000)(scala.util.Random.nextInt(size))
+//    }
   }
 
   @Benchmark
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
@@ -28,7 +28,9 @@ class HashSetBenchmark {
     def freshCollection() = HashSet((1 to size).map(_.toLong): _*)
     xs = freshCollection()
     xss = scala.Array.fill(1000)(freshCollection())
-    randomIndices = scala.Array.fill(1000)(scala.util.Random.nextInt(size))
+    if (size > 0) {
+      randomIndices = scala.Array.fill(1000)(scala.util.Random.nextInt(size))
+    }
   }
 
   @Benchmark
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
@@ -16,7 +16,7 @@ import scala.Predef.intWrapper
 @State(Scope.Benchmark)
 class HashSetBenchmark {
 
-  @Param(scala.Array("0", "1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "73121", "7312102"))
+  @Param(scala.Array(/*"0", */"1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "73121", "7312102"))
   var size: Int = _
 
   var xs: HashSet[Long] = _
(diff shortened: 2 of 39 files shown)


src/main/scala/strawman/collection/Set.scala
```diff
@@ -42,9 +58,13 @@ trait SetLike[A, +C[X] <: Set[X]]
...
 
-  def & (that: Set[A]): Repr
+  def & (that: Set[A]): Repr = this.filter(that)
+
+}
+
+trait SetPolyTransforms[A, +C[X]] extends IterablePolyTransforms[A, C] {
 
-  def ++ (that: Set[A]): Repr
+  def ++ (that: Set[A]): C[A]
```
julienrf: Note that I moved `++` into a new `SetPolyTransforms` trait despite the method returns a `Set` with the same element type `A`. This is just in order to be able to provide a default implementation (see in `immutable.Set`) that gets the self-type right.

src/main/scala/strawman/collection/immutable/HashSet.scala
```diff
@@ -2,40 +2,325 @@ package strawman
...
+
+    override def size: Int = 0
+
+    protected def get0(elem: Any, hash: Int, level: Int) = false
+
+    protected def updated0(elem: Any, hash: Int, level: Int) = new HashSet1(elem, hash)
+
+    protected def removed0(key: Any, hash: Int, level: Int) = this
+
+  }
```
julienrf: One difference between this version and the version that is currently in 2.12 is that in 2.12 the protected methods above have a default implementation that corresponds to the meaning of an empty set, and these methods are then overriden in all the subtypes of `HashSet` excepted `EmptyHashSet`.

Here, I chose to leave these protected methods abstract and to implement them only in the leaves of the hierarchy. I don’t know if it has an impact on performance. Any hint?

DarkDimius: > I don’t know if it has an impact on performance

I wouldn't expect a measurable difference, but it's better to validate using benchmarks.

src/main/scala/strawman/collection/Set.scala
julienrf: Note also that `IterablePolyTransforms` provides another `++` method with the following signature:

~~~ scala
def ++ [B >: A](that: IterableOnce[B]): C[B]
~~~

Should we keep both methods? I don’t think so.

src/main/scala/strawman/collection/immutable/Set.scala
```diff
@@ -22,4 +23,15 @@ trait SetMonoTransforms[A, +Repr]
...
+}
+
+trait SetPolyTransforms[A, +C[X] <: Set[X] with SetLike[X, C]]
+  extends collection.SetPolyTransforms[A, C] {
+
+  protected def coll: C[A]
+
+  def ++ (that: collection.Set[A]): C[A] =
+    if (that.isEmpty) coll
+    else that.foldLeft[C[A]](coll)(_ + _)
```
julienrf: This is the generic implementation of `++`. In contrast to the one inherited from `IterablePolyTransforms`, this one does not rely on the `View` + `fromIterable` combination.

I think I should check whether the `View` + `fromIterable` combination offers the same performances as this implementation or not.

A major drawback of this implementation is that it requires complex bounds for the type parameter `C[_]`.

julienrf: So, I ran the benchmarks and here is the result:

concat

The best performance is achieved by the blue line (Scala 2.12 current `HashSet`). The green line show the performance of the above version of `++`, in our strawman, and the red line shows the performance of the `++` version inherited from [`IterablePolyTransforms`](https://github.com/scala/collection-strawman/blob/master/src/main/scala/strawman/collection/Iterable.scala#L189), based on `View.Concat` and `fromIterable`. Unfortunately, the version inherited from `Iterable` is ~2 times slower the one defined above…

julienrf:
Another performance regression revealed by the benchmark:

foreach

(the blue line shows Scala’s current `HashSet`)

Actually several operations are 1.5 to 2 times slower in our new `HashSet`.

src/main/scala/strawman/collection/immutable/HashSet.scala
julienrf: I ran the benchmarks and saw no difference at all between these alternatives.

DarkDimius: As expected. Good!

README.md
```diff
@@ -44,4 +44,68 @@ We also automatically publish snapshots on Sonatype:
...
+- [ ] `HashSet`
+- [ ] `ListSet`
+- [ ] `TreeSet`
+- [ ] `EqSet`
+- [ ] `BitSet`
+- [x] `View`
+
+## Implemented operations (on the relevant collection types)
+
+### Terminal operations
```
Ichoran: Since the rewrite is still eager, this terminology is especially confusing since it sounds like the Java 8 Streams terminology meaning "evaluate this now".  Maybe use "Methods not returning the same collection type"?

README.md
```diff
@@ -44,4 +44,68 @@ We also automatically publish snapshots on Sonatype:
...
+- [x] `take`
+- [ ] `groupBy`
+
+### Transformations to collections that can have a different element type
+
+- [x] `++` / `concat`
+- [x] `flatMap`
+- [x] `map`
+- [x] `zip`
+
```
Ichoran: Did my `collect` implementation stay in?  If yes, it's here too.  (It's only a mockup, though.)

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  @Benchmark
+  def foreach(): Any = {
+    var n = 0
+    xs.foreach(x => if (x eq null) n += 1)
+    n
+  }
+
+  @Benchmark
+  def map(): Any = xs.map(x => if (x eq null) "foo" else "bar")
```
Ichoran: This is not suitable for a hash benchmark because it results in a set with at most two values.  That lookup is pretty unlike a normal hash set lookup.

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
```diff
@@ -0,0 +1,57 @@
...
+    xs = HashSet((1 to size).map(_.toString): _*)
+    obj = ""
+  }
+
+  @Benchmark
+  def cons(): Any = {
+    var ys = HashSet.empty[Any]
+    var i = 0
+    while (i < size) {
+      ys = ys + obj // Note: we should test different cases: colliding values as well as non-colliding values, etc.
```
Ichoran: Indeed, we should.  There is little point looping `size` times while always trying to add the same element.  Let's do something more interesting, or take this test out.

benchmarks/time/src/main/scala/strawman/collection/immutable/ScalaHashSetBenchmark.scala
```diff
@@ -0,0 +1,57 @@
...
+import scala.{Any, AnyRef, Int, Unit}
+import scala.Predef.intWrapper
+
+@BenchmarkMode(scala.Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Fork(1)
+@Warmup(iterations = 12)
+@Measurement(iterations = 12)
+@State(Scope.Benchmark)
+class ScalaHashSetBenchmark {
```
Ichoran: Note: will need to be kept up to date with the strawman version.

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -68,6 +70,18 @@ trait IterableOps[+A] extends Any {
   /** The first element of the collection. */
   def head: A = iterator().next()
 
+  /** Selects the last element.
+    * $orderDependent
+    * @return The last element of this $coll.
+    * @throws NoSuchElementException If the $coll is empty.
+    */
+  def last: A = {
+    var lst = head
```
Ichoran: This is very unlikely to be efficient since you create an iterator in addition to going through a foreach, which by default will create a second iterator.

Just do this manually:

```scala
val i = iterator()
var lst = i.next()
while (i.hasNext) lst = i.next()
lst
```

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -109,12 +140,21 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
     def next() = throw new NoSuchElementException("next on empty iterator")
   }
+
+  def single[A](a: A): Iterator[A] = new Iterator[A] {
+    private var consumed: Boolean = false
+    def hasNext = !consumed
+    def next() = if (consumed) empty.next() else { consumed = true; a }
+  }
+
   def apply[A](xs: A*): Iterator[A] = new IndexedView[A] {
```
Ichoran: This is going to keep the pointless box of `IndexedView` around, isn't it?  This should be as thin a wrapper around `xs` as possible.  (Maybe this PR isn't the place to fix it, but if you're adding `single`, why not get this also?)

src/main/scala/strawman/collection/Set.scala
Ichoran: There is nothing set-specific about `++`.  It should just get it from `IterablePolyTransforms`.

src/main/scala/strawman/collection/immutable/HashSet.scala
Ichoran: I've never encountered a case where it mattered whether I overrode all but one leaf or left it abstract and overrode every leaf.

src/main/scala/strawman/collection/immutable/HashSet.scala
```diff
@@ -2,40 +2,325 @@ package strawman
...
+          Array.copy(elems, 0, elemsNew, 0, elems.length)
+          elemsNew(offset) = subNew
+          val sizeNew = size + (subNew.size - sub.size)
+          new HashTrieSet(bitmap, elemsNew, sizeNew)
+        }
+      } else {
+        this
+      }
+    }
+  }
```
Ichoran: I haven't checked this carefully.  I'm assuming it's the obvious cut-and-paste and the algorithm is unchanged.

src/main/scala/strawman/collection/immutable/ListSet.scala
```diff
@@ -0,0 +1,126 @@
...
+  override def size: Int = 0
+  override def isEmpty: Boolean = true
+
+  def contains(elem: A): Boolean = false
+
+  def + (elem: A): ListSet[A] = new Node(elem)
+  def - (elem: A): ListSet[A] = this
+
+  def iterator(): strawman.collection.Iterator[A] = {
+    def reverseList = {
```
Ichoran: Do we have any idea why this is a separate def rather than just creating the list inline?

src/main/scala/strawman/collection/immutable/Set.scala
Ichoran: Maybe the fold should be inlined into a bare iterator walk in `IterablePolyTransforms`.  Would that be faster?

README.md
julienrf: They are effectively terminal operations on `View`s. These operations do not return the same collection type (e.g. `size`).

README.md
julienrf: I think this `collect` is not in `IterablePolyTransforms` yet.

src/main/scala/strawman/collection/Iterator.scala
julienrf: I don’t understand what you mean by “keeping the box around”, can you detail?

src/main/scala/strawman/collection/Set.scala
julienrf: Yes, I agree!

src/main/scala/strawman/collection/immutable/HashSet.scala
julienrf: Yes, I copy pasted from the existing collections.

src/main/scala/strawman/collection/immutable/Set.scala
julienrf: Actually, `++` uses `+`, which is notably slower in the strawman than in 2.12:

cons

I will investigate what’s wrong…

julienrf: For reference: I tried inlining the `foldLeft` loop and only got a very slight improvement. On the other hand, removing the `if (that.isEmpty)` check had more impact.

BTW, in the current collections, the implementation is just a `foldLeft` (not inlined, and without first checking if the collection to concatenate is empty or not).

README.md
Ichoran: Yes, I know, but the interface presented is an eager one so using the language of a lazy interface is confusing.  I guess since it's just for us it's no big deal so long as we all know what it means.

src/main/scala/strawman/collection/Iterator.scala
Ichoran: The `IndexedView` is created just so you can get an `iterator` on it.  So you take elements, create an object to hold them, then create _another_ object to point to the first object, with all sorts of indirection on the way.  `IndexedView` is just a box around `xs` providing `iterator`, but it's so simple to write again that we should rather than doubling all allocations, indirections, etc..

src/main/scala/strawman/collection/immutable/ListSet.scala
Ichoran: It usually ends up as an extra method in the bytecode.  Sometimes it's faster because each method is below some important limit for inlining or optimization.  Sometimes it's slower because the extra method call either isn't elided or is but uncleanly leaving extra work to do.  Most of the time it doesn't matter.

But the thing that I don't understand is why it is even here.  The exact same algorithm with the exact same variables would be used if there wasn't a `def`.  So it seems a pointless extra layer.

Ichoran:
I haven't been able to eyeball where the performance regressions come from.  First step is to pick the cleanest of the benchmarks and/or make them cleaner, so we believe that the effect is real.  `foreach` seems like a good choice because it's pretty simple in the old implementation.  Change the benchmark to be decent--consume every element with `Blackhole.consume`, for instance. Maybe the effect will go away.

Next, assuming not, we need to benchmark parts of the implementation to see where the difference comes from.  One suspect: it's `TrieIterator` (but I don't know why, yet).  In the old implementation, `foreach` was overridden for `HashTrieSet`; if you do the equivalent operation on the iterator (a simple sum, for instance), you get a ~15% performance penalty.  What if it's overridden here?  If it shows a big difference, it may be that `TrieIterator` is having trouble keeping up.  If it doesn't, that suggests that more of the difference is at the leaves.  I think within a few rounds we should have an answer.

(Needless to say, a 2x performance decrease is not acceptable when we already get dinged by not keeping up with `java.util`.)

lrytz:
Note that Scala releases are built with `-opt:l:classpath`, it seems this repo is not

src/main/scala/strawman/collection/Set.scala
julienrf: Actually, I think this is not possible to implement this `++` signature in a sound way: see my attempt here: https://github.com/scala/collection-strawman/compare/hashet-benchmarks...unsound-set-concat

Let me explain a bit better the tradeoff. The `++` method inherited from `Iterable` has the following signature:

~~~ scala
def ++ [B >: A](that: IterableOnce[B]): Set[B]
~~~

While the `Set`-specific version has the following signature:

~~~ scala
def ++ (that: Set[A]): Set[A]
~~~

The former forces us to rebuild a whole `Set[B]` and to add to it all the elements of `this` and all the elements of `that`. The latter allows us to reuse `this` as it is and just add to it all the elements of `that`. This optimization can not be achieved with the former type signature because we would have to upcast `this` into a `Set[B]`, which is unsound because of the way `contains` is defined (`A` is in contravariant position). By the way, this makes me think that we will have the same issue with `contains`, if we define it in `Iterable`.

I would be tempted to remove these operations from `Iterable` but that would make `Iterable` less useful… Alternatively, we could live with overloads:

image

julienrf:
Some news: I pushed some more commits and now we have about the same performance as 2.12 collections. The main change is that now `splitAt` (which is used by `TrieIterator`) is now implemented in terms of `take` and `drop`, by default, and benefits from optimizations brought to these methods.

Here is the chart for concat:

concat

Remember that the vertical axis is logarithmic… In the worst case (282 elements) we are still ~1.6x slower than 2.12.

And here is the chart for cons:

cons

src/main/scala/strawman/collection/Set.scala
julienrf: Note that currently we have the same situation in 2.12: `Set` has two overloads of `++` (one inherited from `Iterable`, and one introduced in `Set`).

Ichoran:
Looks much better!  Nice work!

src/main/scala/strawman/collection/Set.scala
szeiger: I don't see why this would be unsound. I don't want to open up the can of worms with intentional vs extensional sets again but no matter which one you pick (which determines variance) we still have universal equality and hashing, A default implementation with a typecast may not be correct in all cases (`SortedSet` could be a problem here) but `Set[A]` -> `Iterable[A]` -> `Iterable[B]` -> `Set[B]` is always well defined, so there must always be a way to implement this operation.

szeiger:
What's the margin of error for these benchmarks? Can we add this to the plots? The performance going back and forth in the higher element count measurements doesn't make much sense.

julienrf:
The margin of error is actually drawn. But in the images I posted the error bar was too small so we don’t even see it. The uncons benchmark has a higher error bar, for instance:

uncons

src/main/scala/strawman/collection/Set.scala
julienrf: Hmm, indeed I tried to build a counter-example but I failed: https://scastie.scala-lang.org/julienrf/KLo8JYnZRKmiIuolI7lZxg (also available in [this gist](https://gist.github.com/julienrf/7317162fec0f1a746c94021abd494771)). Scalac refuses to compile the unsound code :)

Still, I’m not 100% sure that it would be safe…

julienrf:
I think we should merge the PR as it is, so that we can move forward to implementing other collection types. We can still improve these collections later.

src/main/scala/strawman/collection/Set.scala
julienrf: Thinking about this problem again, I ended up with the following question: do we really want to keep (immutable) `Set` invariant? Why not make it covariant like other immutable collection types? Do people often use the “intentional” facet of a `Set` very often in practice? We can anyway turn a `Set[A]` into a `[B >: A]B => Boolean`, like we do with other immutable collection types. The drawback is that we can get this `B` inferred to `Any` without noticing it… But we already have this issue with the other immutable collection types, right?

DarkDimius: One good reason to keep immutable Set invariant is that many immutable sets are based on comparators and it's not that easy to write a compactor that is able to compare all the children of a non-sealed class. But you want to be able to put  classes from non-sealed higherarchies into immutable sets.

PR #56, opened 2017-04-04
Implement HashMap and TreeMap

julienrf:
This work is based on the `hashset` branch. I targeted the PR to this branch so that the diff shows only the changes performed on top of that branch.

This PR adds implementation of `HashMap`, `TreeMap` and `ListMap`.

I mainly copy-pasted the implementation from 2.13.x branch without much changes.

However, I introduced a new abstraction: `MapValuePolyTransforms[K, V, C[?, ?]]`, which contains operations that are polymorphic on the type of the values of a Map-like collection. This abstraction makes it possible to reuse more code between `Map` and `SortedMap`. Indeed, `SortedMap[K, V]` does not inherit from `MapPolyTransforms[K, V, SortedMap]`, but from `MapPolyTransforms[K, V, Map]` (note that the self-type is `Map` instead of `SortedMap`), because these polymorphic transformation methods do not take an `Ordering[K]` parameter (required to build a `SortedMap`). As a result, all these inherited operation return a `Map`, not a `SortedMap`. However, not all of the operations defined in `MapPolyTransforms` actually require this `Ordering` parameter to work on `SortedMap`. An example of that is `concat`, which uses the same type of keys but might have a different type of values :

~~~ scala
def concat[V2 >: V](xs: collection.Iterable[(K, V2)]): C[K, V2]
~~~

Before this PR, all these methods (like also `updated`, for instance) were defined in `MapPolyTransforms` and then we had to take care of overriding them (and duplicating their implementation) in `SortedMapPolyTransforms`, just to get `SortedMap` as their return type instead of it being widened to `Map`.

With `MapValuePolyTransforms` it is now possible to define and implement them once and to inherit from them in both `HashMap` and `TreeMap`. Also, it has also been useful to define a generic `ImmutableMapBuilder` type used by both `HashMap` and `TreeMap`.


src/main/scala/strawman/collection/BitOperations.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import scala.Predef.{intWrapper, longWrapper}
+import java.lang.String
+
+/** Some bit operations.
+ *
+ *  See http://www.drmaciver.com/2008/08/unsigned-comparison-in-javascala/ for
+ *  an explanation of unsignedCompare.
+ */
+private[collection] object BitOperations {
```
Ichoran: Actually, why do we need *any* of this?  It doesn't seem to be used here.

Ichoran:
Looks reasonable overall, except for the bitops stuff that seems to serve no purpose.  The overall strategy seems in line with what we've already got.

src/main/scala/strawman/collection/BitOperations.scala
julienrf: I copy-pasted this file without any change but you are totally right we should definitely get rid of it if it’s not that useful. So far we are only using `unsignedCompare` (from the inner trait `Int`).

I see that in Scala collections with use these `BitOperations` in `IntMap` and `LongMap` (which are optimized `Map` implementations for `Map[Int, ?]` and `Map[Long, ?]`).

julienrf:
Alright @Ichoran, I removed `BitOperations` (although we might re-introduce it when we will add `IntMap` and `LongMap`). I kept the custom `unsignedCompare` method instead of using `java.lang.Integer.compareUnsigned`, because their behavior is not exactly the same (e.g. `unsignedCompare(0, 0)` returns `true` whereas `Integer.compareUnsigned(0, 0) < 0` return `false`) and I don’t want to introduce a regression.

julienrf:
(I deleted my previous comment, which was wrong about `compareUnsigned`)

I removed `BitOperations` and am now using `Integer.compareUnsigned`.

PR #57, opened 2017-04-12
Provide alphanumeric methods with descriptive names as alternatives to symbolic operators

### LINKED ISSUE #46: Provide descriptive names as alternatives to the most commonly used symbolic method names
As outlined in http://docs.scala-lang.org/style/naming-conventions.html the use of symbolic method names should be avoided. 

I suggest implementing descriptive alternatives to the most commonly used symbolic method names used in the Scala Collections API. I understand that this would add more methods to the API, but I strongly believe it would make the API more accessible for people without an extensive functional background.

As an example, `Map` already now provides `updated` method that is almost synonymous to `+`.

Similarly `++` in `Seq` (or its super class) could be called `concat`
As another example `:+` in `Seq` could be called `append`

julienrf:
Fixes #46.

I systematically added methods with alphanumeric names for all operations and defined symbolic operators as aliases.

I suffixed operations on mutable collections with `InPlace`, since that’s the new convention.

### DIFF
--- a/src/main/scala/strawman/collection/mutable/Builder.scala
+++ b/src/main/scala/strawman/collection/mutable/Builder.scala
@@ -7,9 +7,6 @@ import strawman.collection.{IterableMonoTransforms, IterableOnce}
 /** Base trait for collection builders */
 trait Builder[-A, +To] extends Growable[A] { self =>
 
-  /** Append an element */
-  def +=(x: A): this.type
-
   /** Clears the contents of this builder.
    *  After execution of this method the builder will contain no elements.
    */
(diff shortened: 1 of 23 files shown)


src/main/scala/strawman/collection/mutable/Builder.scala
```diff
@@ -20,22 +17,25 @@ trait Builder[-A, +To] extends Growable[A] { self =>
...
 
   def clear() = sb.setLength(0)
 
-  /** Overloaded version of `++=` that takes a string */
-  def ++= (s: String) = { sb.append(s); this }
+  /** Overloaded version of `addAllInPlace` that takes a string */
+  def addAllInPlace(s: String): this.type = { sb.append(s); this }
+
+  /** Alias for `addAllInPlace` */
+  def ++= (s: String): this.type = addAllInPlace(s)
```
szeiger: Should be marked `@inline`. And what happened to `+=`?

julienrf: `+=` is defined in `Growable`

PR #59, opened 2017-04-12
Compile under dotty

odersky:
This is a set of changes needed to compile the strawman under dotty. The first 4 commits are simple rewrites or only concern some details. The next two commits are more problematic.

fc09109 shows a failure to infer the right type instance of `buildFromConstrainedPolyBuildable`

It's signature is the epic

    implicit def buildFromConstrainedPolyBuildable[Ev[_], CC[_], Impl[_], A, E : Ev]: BuildFrom[ConstrainedPolyBuildable[A, CC, Ev], E] { type To = CC[E] }

Btw, `Impl[_]` is unused - why are we adding a 5th type parameter here? 

I have no idea why dotty fails to find in the implicit, and in fact have no idea how an implicit os supposed to be found in the first place. Can someone explain the intended sequence of instantiations here?

a475189 shows a failure of type inference due to union types. See the commit comment. The worrying thing here is that we see `ListBuffer.To` in the error message, which IMO should not happen.

### DIFF
--- a/src/test/scala/strawman/collection/test/TraverseTest.scala
+++ b/src/test/scala/strawman/collection/test/TraverseTest.scala
@@ -57,7 +57,8 @@ class TraverseTest {
     val o1t: Option[immutable.List[Int]] = o1
 
     val xs2 = immutable.TreeSet(Some("foo"), Some("bar"), None)
-    val o2 = optionSequence(xs2)
+    val o2 = optionSequence(xs2)(
+     BuildFrom.buildFromConstrainedPolyBuildable[math.Ordering, immutable.TreeSet, Option[String], String])
     val o2t: Option[immutable.TreeSet[String]] = o2
 
     // Breakout-like use case from https://github.com/scala/scala/pull/5233:
--- a/src/test/scala/strawman/collection/test/TraverseTest.scala
+++ b/src/test/scala/strawman/collection/test/TraverseTest.scala
@@ -69,9 +69,8 @@ class TraverseTest {
 
   @Test
   def eitherSequenceTest: Unit = {
-    val xs3 = mutable.ListBuffer(Right("foo"), Left(0), Right("bar"))
+    val xs3: mutable.ListBuffer[Either[Int, String]] = mutable.ListBuffer(Right("foo"), Left(0), Right("bar"))
     val e1 = eitherSequence(xs3)
     val e1t: Either[Int, mutable.ListBuffer[String]] = e1
   }
-
 }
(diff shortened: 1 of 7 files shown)


src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -69,9 +69,8 @@ class TraverseTest {
 
   @Test
   def eitherSequenceTest: Unit = {
-    val xs3 = mutable.ListBuffer(Right("foo"), Left(0), Right("bar"))
+    val xs3: mutable.ListBuffer[Either[Int, String]] = mutable.ListBuffer(Right("foo"), Left(0), Right("bar"))
```
Ichoran: If dotty initially infers `Right[Nothing, String] | Left[Int, Nothing]` and has a rule that: "if the type needs to be inferred, look for supertypes that avoid `Nothing`", we'll get `Either[Int, String]` which is what we want.  If this is too greedy, one could restrict it to only operate on enums.  (It should probably be symmetric with Any in contravariant position.)

smarter: > If this is too greedy, one could restrict it to only operate on enums

You wouldn't have this problem with enums because the return types of their apply methods are always widened to the enum type

smarter: If the collection strawman defined its own version of `Either` it could do the same by hand in the `apply` of `Left` and `Right`.

odersky: @ichoran That's not how implicit search works. We infer a type for certain type variables and then stick to it. If we selectively widen on failure we'd get another dimension of search complexity.

@smarter Yes, changing Either would work here, but I fear the problem is more general.

odersky: The problem is the type To in ListBuffer, which is non-variant. So even if we could widen, this would not work in this case.

julienrf:
Thanks a lot Martin for this work :)

I will continue on it and update the build so that anyone will be able to cross-compile for Dotty from sbt (that will make it possible to detect future dotty incompatibilities from the CI, for instance).

We also need to fix the `Ordering` and the `Left | Right` problems. I think `Ordering` should be fixed in the standard library, but I guess the `Left | Right` vs `Either` typing problem should be solved on the dotty side, right?

odersky:
I might have a fix on the dotty side for both the `Ordering` and `Left | Right` vs `Either` problem, but it is involved.

Issue #54, opened 2017-03-29
Benchmarking guidelines


DarkDimius:
I'm writing this as a guide on how to make contributed benchmarks "more valid" and actually measure what they are supposed to. Note that as of current state benchmarks in this repo don't follow them.

- make sure to benchmark for small collections(<4 elems), they are the most common
- make sure not to use sizes which are "magical" for your collection. E.g. Degrees of 2. I propose to standardise on those sizes: `0, 1, 2, 3, 4, 7, 8, 15, 16, 17, 39, 282, 73121, 7312102`.
- Use `Blackhole.consume` instead of returning a value from a method
- Use `Blackhole.consume` to ensure that data you are accessing is actually accessed
- don't use `foreach` as a way to run multiple iterations of benchmark. Use a hand-written while cycle and `@OperationsPerInvocation`, eg https://github.com/DarkDimius/DelayingArrays/blob/rewrite/src/main/scala/me/d_d/delaying/Benchmarks.scala#L151
- If you are accessing the same data in every benchmark iteration, eg. `list.head`, make sure to have a set-up run to ellide it from memory. This is hard to get right and this is why I recommend to not include such benchmarks.
- If you are accessing random data, make sure not to use random number generator provided in Java. It's slow. You'll be benchmarking RNG instead of your collection. Use this one instead: https://github.com/DarkDimius/DelayingArrays/blob/rewrite/src/main/scala/me/d_d/delaying/Benchmarks.scala#L58
- make sure you're not measuring boxing\string allocation. It may be slower than the actual operaion being performed in benchmark.
- best way to ensure this, read assembly resulting from JIT-ting your code;

If you want to get reliable numbers when running on your machine , make sure that you have:
 - make sure you're not running inside a VM;
 - everything that can do some background computation disabled, including:
   - cron;
   - ntpd;
   - OS updates, not only installation, but even checks for updates;
   - web browsers\music
   - filesystems and SSD drives can sometimes do operations that are similar to garbage collection in VM. Make sure not to use them(I recommend launching entire OS in ramdisk).
  - cpu fixed to the same speed: 
     - disabled turbo-boost
     - cpu scheduler fixed to a fixed frequency
     - to ensure that cpu doesn't clockdown due to overheating, fix it to a lower frequency,
     - disable hpet.

Additions and suggestions are welcome.

Ichoran:
Mostly I think this is good advice.  A few quibbles:

[...]

PR #62, opened 2017-04-21
Add ArrayView benchmarks (and baseline)

biboudis:
This set could be useful for examining, through experimentation, to see whether the finally executed code was fused, and by comparing it with the baseline set, to check whether inline+specialization gave interesting results. This would be also interesting for quickly testing auto specialization in dotty compared with the baseline or even Java 8 streams.

I have ported some of these stream benchmarks (or Views in Scala parlance) from our Stream Fusion, to Completeness paper ([pdf](https://arxiv.org/pdf/1612.06668.pdf)).

WDYT?

### DIFF
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
@@ -79,25 +79,6 @@ class ArrayBaselineBenchmark {
   }
 
   @Benchmark
-  def filters(bh: Blackhole) = {
-    var i=0
-    var ret=0L
-    while (i < v.length) {
-      if ((v(i) & 0xD) != 0xCAFED00D &&
-          (v(i) & 0xE) == 0xD15EA5E &&
-          (v(i) & 0xA) != 0xDABBAD00 &&
-          (v(i) & 0xD) == 0xDEADBAAD &&
-          (v(i) & 0xB) != 0xDEADDEAD &&
-          (v(i) & 0xE) == 0xDEADFA11 &&
-          (v(i) & 0xE) != 0xFFBADD11 &&
-          (v(i) & 0xF) == 0x4B1D)
-        ret += v(i)
-      i += 1
-    }
-    bh.consume(ret)
-  }
-
-  @Benchmark
   def maps(bh: Blackhole)= {
     var i=0
     var ret=0L
@@ -105,12 +86,21 @@ class ArrayBaselineBenchmark {
       ret += v(i) +
         (v(i) & 0xD) + 0xCAFED00D +
         (v(i) & 0xE) + 0xD15EA5E +
-        (v(i) & 0xA) + 0xDABBAD00 +
-        (v(i) & 0xD) + 0xDEADBAAD +
-        (v(i) & 0xB) + 0xDEADDEAD +
-        (v(i) & 0xE) + 0xDEADFA11 +
-        (v(i) & 0xE) + 0xFFBADD11 +
-        (v(i) & 0xF) + 0x4B1D
+        (v(i) & 0xA) + 0xDABBAD00
+      i += 1
+    }
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def filters(bh: Blackhole) = {
+    var i=0
+    var ret=0L
+    while (i < v.length) {
+      if ((v(i) & 0x13) != 0x11 &&
+          (v(i) & 0x12) == 0x12 &&
+          (v(i) & 0x11) != 0x10)
+        ret += v(i)
       i += 1
     }
     bh.consume(ret)
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
@@ -1,0 +1,143 @@
+package strawman.collection.immutable
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+import strawman.collection.{ArrayView, View}
+import scala.{Any, AnyRef, Int, Long, Unit, Array}
+import scala.Predef.intWrapper
+
+@BenchmarkMode(scala.Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Fork(1)
+@Warmup(iterations = 12)
+@Measurement(iterations = 12)
+@State(Scope.Benchmark)
+class ArrayBaselineBenchmark {
+
+  @Param(scala.Array("282", "73121", "7312102"))
+  var size: Int = _
+
+  @Param(scala.Array("39"))
+  var vLoSize: Int = _
+
+  var shortRangingFactor : Int = (size * 0.2).toInt
+
+  var v: Array[Long] = _
+  var vLo: Array[Long] = _
+
+  @Setup(Level.Trial)
+  def initData(): Unit = {
+
+    def fillArray(range: Int) = {
+      val array = new Array[Long](range)
+      var i = 0
+      while (i < range) {
+        array(i) = i.toLong
+        i += 1
+      }
+      array
+    }
+
+    v = fillArray(size)
+    vLo = fillArray(vLoSize)
+  }
+
+  @Benchmark
+  def sum(bh: Blackhole) = {
+    var i = 0
+    var ret = 0L
+    while (i < v.length) {
+      ret += v(i)
+      i += 1
+    }
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def sumOfSquares(bh: Blackhole) = {
+    var i = 0
+    var ret = 0L
+    while (i < v.length) {
+      ret += v(i) * v(i)
+      i += 1
+    }
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def sumOfSquaresEven(bh: Blackhole) = {
+    var i = 0
+    var ret = 0L
+    while (i < v.length) {
+      if (v(i) % 2L == 0L)
+        ret += v(i) * v(i)
+      i += 1
+    }
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def filters(bh: Blackhole) = {
+    var i=0
+    var ret=0L
+    while (i < v.length) {
+      if (v(i) > 1 && v(i) > 2 && v(i) > 3 && v(i) > 4 && v(i) > 5 && v(i) > 6 && v(i) > 7)
+        ret += v(i)
+      i += 1
+    }
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def maps(bh: Blackhole)= {
+    var i=0
+    var ret=0L
+    while (i < v.length) {
+      ret += v(i) * 1*2*3*4*5*6*7
+      i += 1
+    }
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def cart(bh: Blackhole)= {
+    var d, dp = 0
+    var ret = 0L
+    while (d < v.length) {
+      dp = 0
+      while (dp < vLo.length) {
+        ret += v(d) * vLo(dp)
+        dp += 1
+      }
+      d += 1
+    }
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def flatMap_take(bh: Blackhole) = {
+    var counter1 = 0
+    var counter2 = 0
+    var ret = 0L
+    var n = 0
+    var flag = true
+    val size1 = v.length
+    val size2 = vLo.length
+    while (counter1 < size1 && flag) {
+      val item1 = v(counter1)
+      while (counter2 < size2 && flag) {
+        val item2 = vLo(counter2)
+        ret = ret + item1 * item2
+        counter2 += 1
+        n += 1
+        if (n == 20000000)
+          flag = false
+      }
+      counter2 = 0
+      counter1 += 1
+    }
+    bh.consume(ret)
+  }
+}
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
@@ -10,13 +10,13 @@ import scala.Predef.intWrapper
 
 @BenchmarkMode(scala.Array(Mode.AverageTime))
 @OutputTimeUnit(TimeUnit.NANOSECONDS)
-@Fork(1)
-@Warmup(iterations = 12)
-@Measurement(iterations = 12)
+@Fork(15)
+@Warmup(iterations = 30)
+@Measurement(iterations = 15)
 @State(Scope.Benchmark)
 class ArrayBaselineBenchmark {
 
-  @Param(scala.Array("282", "73121", "7312102"))
+  @Param(scala.Array("39", "282", "73121", "7312102"))
   var size: Int = _
 
   @Param(scala.Array("39"))
@@ -34,7 +34,7 @@ class ArrayBaselineBenchmark {
       val array = new Array[Long](range)
       var i = 0
       while (i < range) {
-        array(i) = i.toLong
+        array(i) = scala.util.Random.nextInt(size).toLong
         i += 1
       }
       array
@@ -83,7 +83,14 @@ class ArrayBaselineBenchmark {
     var i=0
     var ret=0L
     while (i < v.length) {
-      if (v(i) > 1 && v(i) > 2 && v(i) > 3 && v(i) > 4 && v(i) > 5 && v(i) > 6 && v(i) > 7)
+      if ((v(i) & 0xD) != 0xCAFED00D &&
+          (v(i) & 0xE) == 0xD15EA5E &&
+          (v(i) & 0xA) != 0xDABBAD00 &&
+          (v(i) & 0xD) == 0xDEADBAAD &&
+          (v(i) & 0xB) != 0xDEADDEAD &&
+          (v(i) & 0xE) == 0xDEADFA11 &&
+          (v(i) & 0xE) != 0xFFBADD11 &&
+          (v(i) & 0xF) == 0x4B1D)
         ret += v(i)
       i += 1
     }
@@ -95,7 +102,15 @@ class ArrayBaselineBenchmark {
     var i=0
     var ret=0L
     while (i < v.length) {
-      ret += v(i) * 1*2*3*4*5*6*7
+      ret += v(i) +
+        (v(i) & 0xD) + 0xCAFED00D +
+        (v(i) & 0xE) + 0xD15EA5E +
+        (v(i) & 0xA) + 0xDABBAD00 +
+        (v(i) & 0xD) + 0xDEADBAAD +
+        (v(i) & 0xB) + 0xDEADDEAD +
+        (v(i) & 0xE) + 0xDEADFA11 +
+        (v(i) & 0xE) + 0xFFBADD11 +
+        (v(i) & 0xF) + 0x4B1D
       i += 1
     }
     bh.consume(ret)
@@ -132,7 +147,7 @@ class ArrayBaselineBenchmark {
         ret = ret + item1 * item2
         counter2 += 1
         n += 1
-        if (n == 20000000)
+        if (n == shortRangingFactor)
           flag = false
       }
       counter2 = 0
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayViewBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayViewBenchmark.scala
@@ -1,0 +1,131 @@
+package strawman.collection.immutable
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+import strawman.collection.{ArrayView, View}
+import scala.{Any, AnyRef, Int, Long, Unit, Array}
+import scala.Predef.intWrapper
+
+@BenchmarkMode(scala.Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Fork(1)
+@Warmup(iterations = 12)
+@Measurement(iterations = 12)
+@State(Scope.Benchmark)
+class ArrayViewBenchmark {
+
+  @Param(scala.Array("282", "73121", "7312102"))
+  var size: Int = _
+
+  @Param(scala.Array("39"))
+  var vLoSize: Int = _
+
+  var shortRangingFactor : Int = (size * 0.2).toInt
+
+  var v: View[Long] = _
+  var vLo: View[Long] = _
+
+  @Setup(Level.Trial)
+  def initData(): Unit = {
+
+    def fillArray(range: Int) = {
+      val array = new Array[Long](range)
+      var i = 0
+      while (i < range) {
+        array(i) = i.toLong
+        i += 1
+      }
+      array
+    }
+
+    v = ArrayView(fillArray(size))
+    vLo = ArrayView(fillArray(vLoSize))
+  }
+
+  @Benchmark
+  def sum (bh: Blackhole) = {
+    val ret : Long = v
+      .foldLeft(0L)(_+_)
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def sumOfSquares (bh: Blackhole) = {
+    val ret : Long = v
+      .map(d => d * d)
+      .foldLeft(0L)(_+_)
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def sumOfSquaresEven (bh: Blackhole) = {
+    val ret : Long = v
+      .filter(x  => x % 2L == 0L)
+      .map(x => x * x)
+      .foldLeft(0L)(_+_)
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def mapsMegamorphic (bh: Blackhole) = {
+    val ret : Long = v
+      .map(x => x + 1)
+      .map(x => x + 1)
+      .map(x => x + 1)
+      .map(x => x + 1)
+      .map(x => x + 1)
+      .map(x => x + 1)
+      .map(x => x + 1)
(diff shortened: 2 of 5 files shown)


benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
```diff
@@ -0,0 +1,143 @@
...
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+import strawman.collection.{ArrayView, View}
+import scala.{Any, AnyRef, Int, Long, Unit, Array}
+import scala.Predef.intWrapper
+
+@BenchmarkMode(scala.Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Fork(1)
+@Warmup(iterations = 12)
```
DarkDimius: for 282 elems this is very low

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
```diff
@@ -0,0 +1,143 @@
...
+
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+import strawman.collection.{ArrayView, View}
+import scala.{Any, AnyRef, Int, Long, Unit, Array}
+import scala.Predef.intWrapper
+
+@BenchmarkMode(scala.Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Fork(1)
```
DarkDimius: why?

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
```diff
@@ -0,0 +1,143 @@
...
+@Warmup(iterations = 12)
+@Measurement(iterations = 12)
+@State(Scope.Benchmark)
+class ArrayBaselineBenchmark {
+
+  @Param(scala.Array("282", "73121", "7312102"))
```
DarkDimius: The important use-cases are also small sizes, I've proposed some sizes here: https://github.com/scala/collection-strawman/issues/54

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
```diff
@@ -0,0 +1,143 @@
...
+  var vLo: Array[Long] = _
+
+  @Setup(Level.Trial)
+  def initData(): Unit = {
+
+    def fillArray(range: Int) = {
+      val array = new Array[Long](range)
+      var i = 0
+      while (i < range) {
+        array(i) = i.toLong
```
DarkDimius: for future, it's a bad idea to bias filling towards small numbers, because there's a lot of magic going with them for Ints.
This code is fine, as it uses longs, but it's in general better to use random numbers instead.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
```diff
@@ -0,0 +1,143 @@
...
+    }
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def sumOfSquaresEven(bh: Blackhole) = {
+    var i = 0
+    var ret = 0L
+    while (i < v.length) {
+      if (v(i) % 2L == 0L)
```
DarkDimius: I don't think that the baseline code should be fused, as it's not a real baseline for an array.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
```diff
@@ -0,0 +1,143 @@
...
+    }
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def filters(bh: Blackhole) = {
+    var i=0
+    var ret=0L
+    while (i < v.length) {
+      if (v(i) > 1 && v(i) > 2 && v(i) > 3 && v(i) > 4 && v(i) > 5 && v(i) > 6 && v(i) > 7)
```
DarkDimius: Same here, better unfused

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
```diff
@@ -0,0 +1,143 @@
...
+    }
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def maps(bh: Blackhole)= {
+    var i=0
+    var ret=0L
+    while (i < v.length) {
+      ret += v(i) * 1*2*3*4*5*6*7
```
DarkDimius: same here, but even more extreme

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  @Benchmark
+  def cart(bh: Blackhole)= {
+    var d, dp = 0
+    var ret = 0L
+    while (d < v.length) {
+      dp = 0
+      while (dp < vLo.length) {
+        ret += v(d) * vLo(dp)
```
DarkDimius: v(d) -> v(d + dp) // and guard the cycle.

This is because a _very_ smart jit may hoist multiplication out of the cycle. This is a trivial use-case for polyhedral optimizations. Afaik hotspot doesn't do it now, but it would be nice to have a baseline that withstands time.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
```diff
@@ -0,0 +1,143 @@
...
+  def flatMap_take(bh: Blackhole) = {
+    var counter1 = 0
+    var counter2 = 0
+    var ret = 0L
+    var n = 0
+    var flag = true
+    val size1 = v.length
+    val size2 = vLo.length
+    while (counter1 < size1 && flag) {
+      val item1 = v(counter1)
```
DarkDimius: unfuse

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
biboudis: What do you think about the sizes of the cartesian example? (which I will rename to `flatMap_map_sum`)

I can parameterize both inner and outer but the number of runs on for the benchmarks will explode (not only for cartesian).

WDYT?
`
  @Param(scala.Array("0", "1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "73121", "7312102"))
  var sizeOuter: Int = _

  @Param(scala.Array("0", "1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "73121", "7312102"))
  var sizeInner: Int = _
`

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
szeiger: Would it make sense to benchmark both? Just like most collections in practice are small I bet most Int collections contain small numbers.

Ichoran: If you're benchmarking primitives, it doesn't matter what the values are.  If you're benchmarking something that boxes, you probably should be using `String` instead of `Int` for your base case (and then when testing boxing you can decide whether it makes more sense to use values that may be cached, e.g. `2: Object` or not.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
Ichoran: This is a perfectly reasonable benchmark.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
Ichoran: This is not a good benchmark.  Any self-respecting JIT compiler will elide all the pointless comparisons and just do `v(i) > 7`.  It also shouldn't be called "filters" because this isn't filtering.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
Ichoran: Agreed

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
Ichoran: For reference, the hoisting operation--which does NOT given identical results with IEEE754 floating point arithmetic--is
```scala
var temp = 0L
while (dp < vLo.length) {
  temp += vLo(dp)
  dp += 1
}
ret += v(d) * temp
```
As an optimization it's a bit iffy because of the lack of binary identical results, but it's unquestionably faster.  And once you've done this, you can recognize that `temp` is the same each iteration, and pull that out too.

Thus, addition is not a good way to test a cartesian product like this.  `v(d + dp)` isn't a cartesian product any more, but `ret ^= (v(d) + dp)*(vLo(dp) + d)` or somesuch is pretty hard to extract.  (It's not super-lightweight, though.)

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
Ichoran: I don't think there's anything wrong with fusion here.  It's the natural way to write the operation.  Just because it's not with collections operations doesn't mean that the bare-metal way should be penalized.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayViewBenchmark.scala
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    v = ArrayView(fillArray(size))
+    vLo = ArrayView(fillArray(vLoSize))
+  }
+
+  @Benchmark
+  def sum (bh: Blackhole) = {
+    val ret : Long = v
+      .foldLeft(0L)(_+_)
```
Ichoran: No newline, please.  This makes it considerably harder, not easier, to see what's going on with `v`.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayViewBenchmark.scala
```diff
@@ -0,0 +1,131 @@
...
+  def sumOfSquaresEven (bh: Blackhole) = {
+    val ret : Long = v
+      .filter(x  => x % 2L == 0L)
+      .map(x => x * x)
+      .foldLeft(0L)(_+_)
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def mapsMegamorphic (bh: Blackhole) = {
```
Ichoran: How is this megamorphic?

Ichoran: This test isn't a good one, as it is a trivial candidate for fusion and might actually end up fused by the time the JIT compiler is done with it (and the compiler could do it too).  Unless it is actually a test for fusion, in which case the test name should be `def checkMapFusion`.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayViewBenchmark.scala
```diff
@@ -0,0 +1,131 @@
...
+      .map(x => x + 1)
+      .map(x => x + 1)
+      .map(x => x + 1)
+      .map(x => x + 1)
+      .foldLeft(0L)(_+_)
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def filtersMegamorphic (bh: Blackhole) = {
```
Ichoran: Same as with maps--you are testing fusion, and in a really silly way, too.  At least make it do some work e.g. with `(x => (x & 0xF) != 4)` and such.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayViewBenchmark.scala
```diff
@@ -0,0 +1,131 @@
...
+      .filter(x => x > 4)
+      .filter(x => x > 5)
+      .filter(x => x > 6)
+      .filter(x => x > 7)
+      .foldLeft(0L)(_+_)
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def maps (bh: Blackhole) = {
```
Ichoran: This also seems to contrived to be useful.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayViewBenchmark.scala
```diff
@@ -0,0 +1,131 @@
...
+      .map(function)
+      .map(function)
+      .foldLeft(0L)(_+_)
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def cart (bh: Blackhole) = {
+    val ret : Long = v
+      .flatMap(d => vLo.view.map (dp => dp * d))
```
Ichoran: vLo already is a view.  Why are we viewing it again?

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
DarkDimius: > If you're benchmarking primitives, it doesn't matter what the values are.

I disagree.
The collection API will box, and boxing Integers takes longer, as they have to check if its a small integer and should use a cached value. 
This generates a conditional jump which has its own profile. With small integers one of the branches is never executed.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
DarkDimius: >Just because it's not with collections operations doesn't mean that the bare-metal way should be penalized.

Are we intending to be benchmarking the best case here? If we are looking to get a realistic baseline, the collection implementation and baseline should do the same thing.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
biboudis: (initially @DarkDimius thought that this is a collection's benchmark, so I think he agrees with you now)

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
biboudis: It is filtering indeed but I agree with your points about its exploitable structure. I would like to check fusion of multiple filters. What if I extract the values and blackhole the variables?

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
biboudis: 👍  I'll update!

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayViewBenchmark.scala
biboudis: 👍

biboudis:
Thnks @Ichoran. I update the PR. I have most of the Java infrastructure ready here:

https://github.com/strymonas/java8-benchmarks/blob/master/src/main/java/benchmarks/S.java

will add them!

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayViewBenchmark.scala
biboudis: all are tests for fusion (this is testing `ArrayView`)

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
Ichoran: You can't extract the values; they're all the same. At least do a test that doesn't capture everything in the other test (e.g. `(v(i) & 0xF) == 0xC && (v(i) & 0x30) != 0x10`).

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
Ichoran: @DarkDimius - If the collection API will box primitives, and you want to know its performance in a non-boxing case, you should test the non-boxing case (e.g. use `String`).

The only case where you care about boxing is the mixed primitive/generic case, which can be important, but is usually not as important as the unmixed case(s).

biboudis:
@Ichoran I've updated! I think the filters benchmark doesn't produce statically-known, true results. WDYT?

benchmarks/timeJava8/src/main/java/streams/Java8Streams.java
```diff
@@ -0,0 +1,204 @@
...
+                .map(x -> x + (x & 0xE) + 0xFFBADD11)
+                .map(x -> x + (x & 0xF) + 0x4B1D)
+                .reduce(0L, (a, b) -> a + b);
+        bh.consume(ret);
+    }
+
+    @Benchmark
+    public void filters_boxed(Blackhole bh) {
+        Long ret = Stream.of(v_B)
+                .filter(x -> (x & 0x15) != 0x11)
```
Ichoran: This cuts out 1/8th of the list

benchmarks/timeJava8/src/main/java/streams/Java8Streams.java
```diff
@@ -0,0 +1,204 @@
...
+                .filter(x -> (x & 0x10) == 0x10)
```
Ichoran: This cuts out 1/2 of the list,leaving 3/8ths of the original

benchmarks/timeJava8/src/main/java/streams/Java8Streams.java
```diff
@@ -0,0 +1,204 @@
...
+                .filter(x -> (x & 0x30) != 0x10)
```
Ichoran: This cuts out half of what remains, leaving 3/16ths

benchmarks/timeJava8/src/main/java/streams/Java8Streams.java
```diff
@@ -0,0 +1,204 @@
...
+                .filter(x -> (x & 0x10) == 0x10)
```
Ichoran: This is a no-op (we already had it)

benchmarks/timeJava8/src/main/java/streams/Java8Streams.java
```diff
@@ -0,0 +1,204 @@
...
+                .filter(x -> (x & 0x15) != 0x10)
```
Ichoran: Now we're down to 1/8th of the list (only 0x14 and 0x15 survive).

benchmarks/timeJava8/src/main/java/streams/Java8Streams.java
```diff
@@ -0,0 +1,204 @@
...
+                .filter(x -> (x & 0x10) == 0x10)
```
Ichoran: No-op

benchmarks/timeJava8/src/main/java/streams/Java8Streams.java
```diff
@@ -0,0 +1,204 @@
...
+                .filter(x -> (x & 0x30) != 0x10)
```
Ichoran: No-op

benchmarks/timeJava8/src/main/java/streams/Java8Streams.java
```diff
@@ -0,0 +1,204 @@
...
+                .filter(x -> (x & 0x15) == 0x10)
```
Ichoran: Throws away everything remaining

benchmarks/timeJava8/src/main/java/streams/Java8Streams.java
```diff
@@ -0,0 +1,204 @@
...
+                .reduce(0L, (a, b) -> a + b);
+        bh.consume(ret);
```
Ichoran: This whole thing just discards the entire list.  You shouldn't make your benchmark be, "See if you can figure out that I actually threw away the entire list!" for a default filter benchmark.

biboudis: We can leave the first three combinators. 

```
  @Benchmark
  def maps (bh: Blackhole) = {
    val ret : Long = v
      .map(x => x + (x & 0xD) + 0xCAFED00D)
      .map(x => x + (x & 0xE) + 0xD15EA5E)
      .map(x => x + (x & 0xA) + 0xDABBAD00)
      .foldLeft(0L)(_+_)
    bh.consume(ret)
  }

  @Benchmark
  def filters (bh: Blackhole) = {
    val ret : Long = v
      .filter(x => (x & 0x15) != 0x11)
      .filter(x => (x & 0x10) == 0x10)
      .filter(x => (x & 0x20) != 0x10)
      .foldLeft(0L)(_+_)
    bh.consume(ret)
  }
```

And in the filter case maybe do AND with 0x20

DarkDimius: `(x & 0x20) != 0x10` is always true and easy to optimize away

biboudis: Of course. Will put 0x13!

biboudis: And maybe increase the second a bit.

biboudis:
```scala
.filter(x => (x & 0x13) != 0x11)
.filter(x => (x & 0x12) == 0x12)
.filter(x => (x & 0x11) != 0x10)
```

DarkDimius: I'd propose to change the scheme entirely, as it's too fragile.

```
@volatile var a1 = 237
@volatile var a2 = 432
@volatile var a3 = 435
@Benchmark
  def filters (bh: Blackhole) = {
    a1 = a1 + 1 // make finality analysis fail
    a2 = a2 + 1 
    a3 = a3 + 1 
    val (a1r, a2r, a3r) = (a1, a2, a3)
    val ret : Long = v
      .filter(x => x & a1r == a1r)
      .filter(x => x & a2r == a2r)
      .filter(x => x & a3r == a3r)
      .foldLeft(0L)(_+_)
    a1 = a1 - 1 // stabilize
    a2 = a2 - 1 
    a3 = a3 - 1 
    bh.consume(ret)
  }
```

DarkDimius: The idea is that you want to make optimizations fail. The code above would be good as long as the VM doesn't speculate on values(graal sometimes does).

Ichoran: @DarkDimius - This seems like overkill.  Value speculation is fine as long as it's representative of what can be accomplished in real-world scenarios.  Because of the variation in the input list, the benchmark can't use value speculation to avoid actually finding the correct answer.

The thing to avoid is making the different tests too dependent on each other so that fusion doesn't just avoid unnecessary intermediates but actually shows that the entire exercise is pointless; and to avoid adding steps that aren't measured because all the work is done in the first step.  As you wrote it here, the first step is effectively a Bloom filter: you reject all values less than 237, and thereafter succeed on only 1/64th of them.

It's better to choose some set of simple algorithms that behave predictably, like ((x*8783 >> 7) ^ x) & 0x3 != 0 and (x*1144 >> 7) & 0x3 != 0 and x & 0x3 != 0.

julienrf:
@Ichoran do approve the last changes?

Ichoran:
@julienrf - Sorry, haven't had a chance to check yet.  This week is inordinately busy.  I'll get to it by Monday, or someone else can have a look. @DarkDimius is pretty experienced with the relevant issues if he has time.

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
```diff
@@ -0,0 +1,148 @@
...
+    bh.consume(ret)
+  }
+
+  @Benchmark
+  def maps(bh: Blackhole)= {
+    var i=0
+    var ret=0L
+    while (i < v.length) {
+      ret += v(i) +
+        (v(i) & 0xD) + 0xCAFED00D +
```
Ichoran: This test is misleading since it looks like the additions will be separate.  The compiler ought to merge the constants--so either use multiplication between lines, or manually do the addition.  (The answer is `0xB2D9676B`.)

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
Ichoran: Were you going to update this?

Ichoran:
There are two minor issues outstanding but neither substantially interferes with this benchmark delivering instructive results.  So another modification would be best but if you don't have time @biboudis it looks good enough as is!  Thanks!

benchmarks/time/src/main/scala/strawman/collection/immutable/ArrayBaselineBenchmark.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  @Benchmark
+  def cart(bh: Blackhole)= {
+    var d, dp = 0
+    var ret = 0L
+    while (d < v.length) {
+      dp = 0
+      while (dp < vLo.length) {
+        ret += v(d) * vLo(dp)
+        dp += 1
```
Ichoran: (Comment isn't showing up for me on the overview, but this is still potentially hoistable and doesn't seem to have been updated.  It's a minor issue though.)

PR #63, opened 2017-04-25
Compile under dotty, V2

odersky:
Changes to compile latest master under dotty.

This compiles with the dotty compiler in https://github.com/lampepfl/dotty/pull/2240.

There's a known problem in parameter forwarding, so for the moment you need to compile with

 -Ysuppress-param-forwarding.

Also for the moment you need to compile with a 2.12 version of Either that starts as follows:

    sealed abstract class Either[+A, +B] extends Product1[A | B] with Serializable { ...

If you don't do that, type inference will yield the wrong type in TraverseTest.eitherSequence. We should check whether there's something we can do in the library design about this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ target/
 .DS_Store
 /tmp/
 /bin/
+/.cache-main
+/.cache-tests
--- a/src/main/scala/strawman/collection/BitSet.scala
+++ b/src/main/scala/strawman/collection/BitSet.scala
@@ -3,7 +3,7 @@ package collection
 
 import strawman.collection.BitSetLike.{LogWL, WordLength}
 
-import scala.{Array, Boolean, inline, Int, Long, Option, Ordering, Unit}
+import scala.{Array, Boolean, `inline`, Int, Long, Option, Ordering, Unit}
 import scala.Predef.{assert, intWrapper}
 
 /** Base type of bitsets */
@@ -141,7 +141,7 @@ trait BitSetMonoTransforms[+C <: BitSet with BitSetLike[C]]
     coll.fromBitMaskNoCopy(words)
   }
 
-  @inline final def ^ (other: BitSet): C = xor(other)
+  @`inline` final def ^ (other: BitSet): C = xor(other)
 
   /**
     * Builds a new bitset by applying a function to all elements of this bitset
--- a/src/main/scala/strawman/collection/BuildFrom.scala
+++ b/src/main/scala/strawman/collection/BuildFrom.scala
@@ -32,12 +32,12 @@ object BuildFrom {
     def fromIterable(from: C[A])(it: Iterable[E]): To = from.fromIterable(it)
   }
 
-  /** Build the source collection type from a ConstrainedPolyBuildable */
-  implicit def buildFromConstrainedPolyBuildable[Ev[_], CC[_], A, E : Ev]: BuildFrom[ConstrainedPolyBuildable[A, CC, Ev], E] { type To = CC[E] } = new BuildFrom[ConstrainedPolyBuildable[A, CC, Ev], E] {
+  /** Build the source collection type from a OrderedPolyBuildable */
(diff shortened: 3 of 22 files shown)


odersky:
I don't have the right setup to compile it under scalac. My local setup is still 2.11, which is needed for the dotty project. @julienrf can you make this compile under scalac? 

I also note that 2.11 calac crashes with stackoverflows when trying to compile the project, so it's not just the overloading fix we added.

julienrf:
@odersky I added a commit that fixes compilation issues with scalac 2.12. Note that I removed [this test line](https://github.com/scala/collection-strawman/pull/63/commits/9db1671feab136f9365def7f95385e1d60e07b91#diff-62fbc4823d7bdc006d3762035738b6e5L317).

julienrf:
@DarkDimius I’m not sure whether we want to support this syntax in dotty?

DarkDimius:
Does it mean that this change is unrelated to making it compile under 2.12 and should be in a different commit?

julienrf:
Actually that line did not compile on 2.12 either.

DarkDimius:
Because syntax is supported by current collections, I'd say we should maintain it. 
If we decide not to, we should add a conversion rule to scalafix.

odersky:
@DarkDimius That was something I tried locally. Was not aware that it does not compile under 2.12. We can drop it in that case.

DarkDimius:
@odersky, current Scala 2.11 collections do support this syntax and dropping it may make it harder for people to migrate. 

@olafurpg would it be possible to add a migration rule to scalafix for this?

julienrf:
Now I remember, @szeiger did these changes to support that use case: https://github.com/scala/scala/pull/5698

It has been merged in the 2.13.x branch only. Is it currently possible to use Scala 2.13 SNAPSHOTs? Anyway, that would be nice if the collection strawman was compatible with 2.12…

julienrf:
To be clear, the `{ case (x, y) => … }` syntax does not work in that case because we have two overloads of the method, one that takes a `((A, B)) => C` and one that takes a `(A, B) => C`.

olafurpg:
@DarkDimius Please correct me if I'm wrong, this rewrite would be triggered when `PartialFunction[TupleN, _]` is passed as a `FunctionN` argument? Example

```scala
def foo(f: (A, B) => C)

// before rewrite
foo { case (a, b) => c }
// after rewrite
foo((a, b) => c)
```
The rewrite should not be triggered if `foo` expects a `Function[TupleN, _]` argument

```scala
def foo(f: ((A, B)) => C)

// before rewrite
foo { case (a, b) => c }
// after rewrite
foo { case (a, b) => c } // unchanged
```

If so, I believe this is possible with the current scalafix api.

Does it impact the rewrite if `foo` is overloaded to support both `FunctionN` and Function1[TupleN, _]`?

olafurpg:
@DarkDimius I took a stab at implementing this rewrite and got a prototype working in 20 LOC https://github.com/scalacenter/scalafix/commit/4f51d3879a0bb41ce66d9fc72e44e80e4ca7cdf1 it does not trim away the whitespace around the curly braces, but that should be easy to accommodate or it could be handled by a formatter. As you can see in the unit test, the rewrite is not triggered when `foo` accepts a `Function1[TupleN, _]` parameter.

odersky:
Updates: There's a new PR against dotty #2331 which compiles the strawman without any special options, as soon as the fix for ProductN is in. So I think we can go ahead and merge if others agree.

szeiger:
IMHO both 2.12 and full Dotty compatibility are nice to have but our focus should be 2.13. 2.13.x should be close enough to 2.12 and still binary compatible from the compiler's PoV (not necessarily all parts of the library, but we're not using many library classes here) to make it usable there. Requiring minor changes to the strawman sources for better 2.12 or Dotty compatibility would also be fine to me. For example, in the case of the overloaded `map` call in the test, disabling it on 2.12 (if we want to build on 2.12 at all) is OK, removing it entirely is not (if we expect it to work in 2.13).

PR #66, opened 2017-05-01
Simplify PolyBuildable

julienrf:
I removed the unused type parameter in `PolyBuildable` and `ConstrainedPolyBuildable`.


odersky:
This is a welcome simplification. But since it conflicts with the changes from Constrained to Ordered maybe we should get that in first?

julienrf:
Sure, can I merge #63 now?

PR #67, opened 2017-05-01
Add mutable Map collections implementation

julienrf:
- [x] `mutable.HashMap`
- [x] `mutable.TreeMap`

### DIFF
--- a/src/main/scala/strawman/collection/Factories.scala
+++ b/src/main/scala/strawman/collection/Factories.scala
@@ -107,8 +107,7 @@ trait OrderedMapFactory[+CC[X, +Y] <: SortedMap[X, Y] with SortedMapOps[X, Y, CC
 
   def empty[K : Ordering, V]: CC[K, V]
 
-  def fromIterable[K : Ordering, V](it: Iterable[(K, V)]): CC[K, V] =
-    empty[K, V] ++ it
+  def fromIterable[K : Ordering, V](it: Iterable[(K, V)]): CC[K, V]
 
   def apply[K : Ordering, V](elems: (K, V)*): CC[K, V] =
     empty[K, V] ++ elems.toStrawman
--- a/src/main/scala/strawman/collection/Factories.scala
+++ b/src/main/scala/strawman/collection/Factories.scala
@@ -49,17 +49,16 @@ trait SpecificIterableFactory[-A, +C] extends FromSpecificIterable[A, C] {
 }
 
 /** Factory methods for collections of kind `* −> * -> *` */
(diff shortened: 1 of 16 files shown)


julienrf:
This PR fails to compile with dotty. I think that’s a problem on dotty side with inlining: https://travis-ci.org/scala/collection-strawman/builds/230277499#L2430

/cc @odersky @smarter @DarkDimius

smarter:
@julienrf Yes, please open an issue against dotty with a minimized testcase. Meanwhile, you could try removing the inline annotations and see if that helps.

odersky:
The compilation problems should be fixed in https://github.com/lampepfl/dotty/pull/2408

src/main/scala/strawman/collection/Factories.scala
```diff
@@ -49,22 +49,26 @@ trait SpecificIterableFactory[-A, +C] extends FromSpecificIterable[A, C] {
 }
 
 /** Factory methods for collections of kind `* −> * -> *` */
-trait MapFactory[+CC[X, Y] <: Map[X, Y] with MapOps[X, Y, CC, _]] {
+trait MapFactory[+CC[X, Y]] {
 
   def empty[K, V]: CC[K, V]
```
julienrf: I think we really have to be careful not to introduce such cycles. Currently `fromIterable` implementations often rely on `empty`. Also, very often for concrete collection types we want to override this `empty` factory method for performance. So, I’m not sure it is helpful to have such a default implementation.

src/main/scala/strawman/collection/immutable/HashMap.scala
```diff
@@ -104,6 +100,12 @@ object HashMap extends MapFactory[HashMap] {
 
   def empty[K, V]: HashMap[K, V] = EmptyHashMap.asInstanceOf[HashMap[K, V]]
 
+  def fromIterable[K, V](it: collection.Iterable[(K, V)]): HashMap[K, V] =
+    it match {
+      case hm: HashMap[K, V] => hm
+      case _ => empty ++ it
```
julienrf: I didn’t measure but builders for immutable collections do nothing more than creating an empty collection and then adding the elements one by one. So, that seems equivalent. I would expect not using builders could only improve performance because that’s one abstraction less.

szeiger: For an immutable array, for example, relying on `++` requires this to be optimized for corner cases like the LHS being empty. Maybe that's something we'd want to have anyway but when using a Builder you already know this precondition.

PR #69, opened 2017-05-02
Fix arity of the function parameter in the map method of Map collections

### LINKED ISSUE #68: Arity of map method of Map collection type
~~~ scala
val kvs: Map[Int, String] = null

// This one works fine
val kvs1 = kvs.map((k, v) => (v, k))
val kvs2: Map[String, Int] = kvs1

// This one does not
val kvs3 = for ((k, v) <- kvs) yield (v, k)
val kvs4: Map[String, Int] = kvs3
//                           ^
// type mismatch;
//   found   : strawman.collection.Iterable[(String, Int)]
//   required: strawman.collection.Map[String,Int]
~~~

This is because `Map[K, V]` extends `Iterable[(K, V)]`, and thus inherits a `map[B](f: ((K, V)) => B): Iterable[B]` method, which is used by the `for` expression.

`Map` introduces an overload of `map` with the following signature: `map[K2, V2](f: (K, V) => (K2, V2)): Map[K2, V2]`, but the fact that this overload takes a binary function as parameter rather than a unary function (whose argument type is a tuple) makes it impossible to use it in `for` expressions.

julienrf:
Tries to fix #68.

But I get the following error when I try compiling with scalac (cc @odersky ):

~~~
[error] collection-strawman/src/main/scala/strawman/collection/Map.scala:84: name clash between defined and inherited member:
[error] def map[B](f: ((K, V)) => B): strawman.collection.Iterable[B] in trait IterablePolyTransforms and
[error] def map[K2, V2](f: ((K, V)) => (K2, V2)): C[K2,V2] at line 84
[error] have same type after erasure: (f: Function1)Object
[error]   def map[K2, V2](f: ((K, V)) => (K2, V2)): C[K2, V2] = mapFromIterable(View.Map(coll, f))
[error]       ^
[error] collection-strawman/src/main/scala/strawman/collection/Map.scala:86: name clash between defined and inherited member:
[error] def flatMap[B](f: ((K, V)) => strawman.collection.IterableOnce[B]): strawman.collection.Iterable[B] in trait IterablePolyTransforms and
[error] def flatMap[K2, V2](f: ((K, V)) => strawman.collection.IterableOnce[(K2, V2)]): C[K2,V2] at line 86
[error] have same type after erasure: (f: Function1)Object
[error]   def flatMap[K2, V2](f: ((K, V)) => IterableOnce[(K2, V2)]): C[K2, V2] = mapFromIterable(View.FlatMap(coll, f))
[error]       ^
~~~


odersky:
Maybe try without `mapPolyTransforms`? Define it directly in `Map` and you should be OK.

odersky:
Maybe change the definition of `MapPolyTransforms` from

   trait MapPolyTransforms[K, +V, +C[X, Y]]

to

    trait MapPolyTransforms[K, +V, +C[X, Y] <: Map[X, Y]]

Have you tried that? (The intention is that this would then be erased to Map instead of Object).

julienrf:
Yes, nice trick!

julienrf:
The trick you proposed solved the erasure error but now we have troubles at use site: we can not use the partial function syntax to deconstruct the tuple (see the tests).

Maybe the following patch is supposed to help? https://github.com/scala/scala/pull/5698 (cc @szeiger )

PR #75, opened 2017-05-04
Try simplify

odersky:


Ichoran:
I haven't had time to go through in depth, but on a cursory reading, this looks great!  I had many of the same simplifications in mind.  I think this makes it a lot clearer, save possibly for "MapMappings" which is a bit of a confusing twist since it uses `Map` twice in a row with completely different meanings.  I was going to use `ToAnyElem` instead of `Mappings`.

src/main/scala/strawman/collection/Set.scala
```diff
@@ -52,67 +42,38 @@ trait SetLike[A, +C[X] <: Set[X]]
...
+    */
+  def empty: C
 }
 
 object Set extends IterableFactory[Set] {
+  def newBuilder[A]: mutable.Builder[A, Set[A]] = ??? // immutable.Set.newBuilder
+  def empty[A <: Any]: Set[A] = ??? // immutable.Set.empty
 
-  def newBuilder[A]: mutable.Builder[A, Set[A]] = immutable.Set.newBuilder
-  def empty[A <: Any]: Set[A] = immutable.Set.empty
```
julienrf: Why did you comment that?

julienrf:
I just added a commit to fix the missing imports, for scalac. However, scalac reports *lots* of type errors. I’m not sure this is expected that dottyc happily compiles the code.

julienrf:
You can see the error messages here: https://travis-ci.org/scala/collection-strawman/builds/229042350#L595

odersky:
@ichoran I did another pass, and managed now to merge Mappings with Ops traits so that we only have one base trait instead of 4.

Issue #65, opened 2017-04-30
Have some sort of HasBuilder type class ?


nrinaudo:
Is there any chance to have some sort of type class that would provide features similar to:

```scala
trait HasBuilder[F[_]] {
  def newBuilder[A]: Builder[A, F[A]]
}
```

With default instances for all applicable collection types?

I know of at least one concrete use case: type class based decoding libraries. Let's take the example of a very simple CSV decoding library. One would write decoding type class as follows:

```scala
// Decodes a single CSV cell
trait CellDecoder[A] {
  def decode(cell: String): Try[A]
}

// Decode an entire CSV row
trait RowDecoder[A] {
  def decode(row: Seq[String]): Try[A]
}
```

With `CanBuildFrom`, it's easy to provide an instance of `RowDecoder` for all collection types:

```scala
implicit def cbfDecoder[A, F[_]](implicit da: CellDecoder[A], cbf: CanBuildFrom[Nothing, A, F[A]]): RowDecoder[F[A]] = new RowDecoder[F[A]] {
  override def decode(row: Seq[String]): (row.foldLeft(Success(cbf.apply())) { (acc, cell) => 
    for {
      builder <- acc
      a <- da.decode(cell)
    } yield builder += a).map(_.result())
}
```

I know of at least three libraries that use this mechanism:

* [circe](https://github.com/circe/circe/blob/fb4ef2dcd145682b32f1e1f27e93d773b6f400bf/modules/core/shared/src/main/scala/io/circe/SeqDecoder.scala#L6)
* [argonaut](https://github.com/argonaut-io/argonaut/blob/48b052a13cf5b86f081034e820144338adeb88d9/argonaut/shared/src/main/scala/argonaut/DecodeJson.scala#L186)
* [kantan.csv](https://github.com/nrinaudo/kantan.csv/blob/4d73b3698a1020288aec8516e7eafc8204090698/core/src/main/scala/kantan/csv/RowDecoder.scala#L77)

I understand `CanBuildFrom` is going to disappear, and no mechanism is planned to emulate the above behaviour. The `HasBuilder` type class I'm suggesting would replace it, and is vastly simpler (and less tied to the core implementation of the collection API) than CBF.

It would ideally be part of the standard collection API, although I can't think of anything that would prevent third party libraries from providing that feature externally.

nrinaudo:
Oh, and if this or some shape of this were to be made part of the final design, it'd be _amazing_ if it could be `Serializable` :)

[...]

Issue #34, opened 2017-02-23
Recommended way to produce collections


julienrf:
What’s the recommended way to produce collections?

The default collections are immutable collections. When we [present](https://www.scala-exercises.org/scala_tutorial/standard_library) how to use them we mention the `::` operator (for `List`) to construct a new collection. Constructing a `List` containing the n first fibonacci numbers would then look like this:

~~~ scala
def fibonacci(n: Int): List[Int] = {
  def loop(remaining: Int, prev: Int, prevPrev: Int, result: List[Int]): List[Int] =
    if (remaining > 0) {
      val value = prev + prevPrev
      loop(remaining - 1, value, prev, value :: result)
    } else result
  loop(n - 2, 1, 0, List(1, 0)).reverse
}
~~~

But then one might ask: would it be possible to construct the sequence in the right order from the beginning?

One way to do that is to use `Builder`s:

~~~ scala
def fibonacci(n: Int): List[Int] = {
  val builder = List.newBuilder[Int]
  builder += 0
  builder += 1
  def loop(remaining: Int, prev: Int, prevPrev: Int): Unit = {
    if (remaining > 0) {
      val value = prev + prevPrev
      builder += value
      loop(remaining - 1, value, prev)
    } else ()
  }
  loop(n - 2, 1, 0)
  builder.result()
}
~~~

But then mixing immutable and mutable programming style feels a bit awkward… We could go all mutable but then that would contrast with the general advice of going immutable first.

We can consider several alternative solutions, like having an immutable `Tsil` that would append elements instead of prepending them, but then the resulting collection would require precautions (e.g. we should avoid to use head/tail decomposition on it).

Otherwise, we could use a collection implementation that provides good performance for both `+:` and `:+`, like `Vector`. But that seems perfectible…

I think that it would be great to provide a general API to efficiently construct all kind of collections. The user-facing API would use immutable style, but the underlying implementation would use specific mutable collections (e.g. `ListBuffer` to build a `List`, etc.).

It seems that what I am looking for are “unfolds”, but I’m wondering if people tried other alternatives or have any advice to address this problem? Does someone know a nice implementation of unfolds? (I mean, a lot of variants could make sense: should we discard the final state or not? should we support the production of several elements in a single step? conversely, should we support not producing an element yet without stopping the process? etc.)

odersky:
I usually use a `ListBuffer` as an accumulator, like this:
   
    def fibonacci(n: Int): List[Int] = {
[...]

PR #76, opened 2017-05-05
Merge ops, mappings, and like traits

odersky:
Merge ops, mappings, and like traits into an Ops trait with 3 parameters.

Also, drop ImmutableBuilders, we don't seem to need them.


DarkDimius:
Glad this idea worked!

julienrf:
There is an issue because `ArrayOps` and `StringOps` can *not* implement the self-type they pretend to implement.

julienrf:
To give you more information, the heart of the issue is here:

~~~
[error] /home/travis/build/scala/collection-strawman/src/main/scala/strawman/collection/View.scala:136: illegal inheritance;
[error]  trait IndexedView inherits different type instances of trait IterableOps:
[error] strawman.collection.IterableOps[A,strawman.collection.Seq,strawman.collection.Seq[A]] and strawman.collection.IterableOps[A,strawman.collection.View,strawman.collection.View[A]]
[error] trait IndexedView[+A] extends View[A] with Seq[A] { self =>
[error]       ^
~~~

The problem is that `Seq[A]` inherits `IterableOps[A, Seq, Seq[A]]`, whereas `View[A]` inherits `IterableOps[A, View, View[A]]` but `Seq` and `View` are unrelated types. The fix should be to make `IndexedView` extend `SeqLike[A, IndexedView, IndexedView[A]]` but then we would have to re-implement `fromIterable` and `fromSpecificIterable` because we wouldn’t anymore be able to reuse implementations inherited from `View` (because they return a `View` instead of an `IndexedView`. But then even if we re-implement them that means implementing the following:

~~~ scala
def fromIterable[B](c: Iterable[B]): IndexedView[B]
~~~

Which is not possible when the given `Iterable[B]` does not itself extend `Seq[B]`.

odersky:
@julienrf Good analysis, thanks! I fixed it by not having IndexedView inherit from Seq.

julienrf:
I added one commit that essentially fixes some `C` and `CC` types in the hierarchy, moves the `fromIterable` method implementation to the leaves, adds explicit type annotations to the `xs.map { case (k, v) => … }` tests and comments out the `traverse` tests.

szeiger:
So what about TraverseTest? This is hardly a problem we can ignore. We need these methods in the standard library and I'm pretty sure they will also be common in third-party code.

julienrf:
@szeiger I agree with you that we should support these operations. And also, I think we should even support the use case described in #65. On the other hand, the previous solution was maybe too complex. We should try to make it simpler.

szeiger:
The essence of this simplification here seems to be the removal of all the work I did on `BuildFrom` (which went through several iterations to come up with a simple design). This includes single-method extensions (like `Future.traverse`, a breakOut-like pattern, and first-class support for strict building. We're taking a giant step back by removing it instead of trying to simplify it further.

julienrf:
I think we can re-introduce a simplified `BuildFrom` thing on top of which things like `Traverse`, generic programming things (like in #65) and unfolds (like in #34) could be implemented externally.

Issue #44, opened 2017-03-08
Extension Methods without CanBuildFrom?


dlwh:
(Maybe I'm late to the party on this. I've been a bit checked out from mailing lists and such. Sorry if so!)

I'm curious about the future of ergonomic generic extension methods in the new collections library, at least if `CanBuildFrom` is removed entirely.

For instance, in our codebase, we've introduced a method `flatCollect`, which we frequently find useful. (I noticed `collect`'s missing, I assume that's just because it's relatively uninteresting, in terms of design.) It has the obvious meaning, with signature:

```scala
def flatCollect[Result, B](f: PartialFunction[A, TraversableOnce[B]])
                              (implicit cbf: CanBuildFrom[Coll, B, Result]): Result 
```

We also have `toMultiMap`, which is an efficient implementation of `xs.groupBy(_.a).eagerMapValues(_.map(_.b))` (as mentioned in #42 but using our own `eagerMapValues` because of frequent unintended consequences of the lazy variant).

```scala
implicit class ScEnrichColl[Coll <: Traversable[_], A, B](val __this: Coll)(implicit view: Coll <:< Traversable[(A, B)]) {
  def toMultiMap[Result](implicit cbf: CanBuildFrom[Coll, B, Result]): Map[A, Result] = ???
}
```

Maybe I'm being insufficiently imaginative, but it's hard for me to see how to implement them in the new approach where they maintain the same level of ergonomics as before. It's possible it's worth the reduction in complexity, but it makes me somewhat sad.

One answer is "use views and use `to`/`FromIterable`" at the end, but I don't think that handles the `toMultiMap` case particularly nicely. Another is to have `CBF`, but only for these cases. (This can be largely done via an external library, possibly.) Is there another?

dlwh:
Basically, what's sad about this is that you lose the ability to extract methods for common patterns like `toMultiMap`, without giving up the more precise types. You can do toMultiMap(_: FromIterable[B]), of course, but:

1. the API is a bit less fluent (imho), and
2. that implementation of toMultiMap I think *has* to make unnecessary intermediates that would not be required using CBF.

julienrf:
In the current state, this kind of use of `CanBuildFrom` would be replaced by an implicit `Builder` factory, as in [this example](https://github.com/scala/collection-strawman/blob/d3b1d42673f2914bebfe75a68c7e7ef15bf5bfc3/src/test/scala/strawman/collection/test/TraverseTest.scala#L15-L19).

[...]

PR #80, opened 2017-05-12
Introduce a CanBuild typeclass witnessing the ability to produce a Builder

julienrf:
This PR addresses #65, #44, #34 and the comments made after we merged #76.

I tried to add as few things as possible to the core of the collections and checked that the problems described in the aforementioned issues could still be fixed.

In essence, all collection factories now provide an implicit `CanBuild[A, C]` instance, making it possible to get an imperative builder yielding a collection `C` containing elements of type `A`.

### DIFF
--- a/src/main/scala/strawman/collection/Factories.scala
+++ b/src/main/scala/strawman/collection/Factories.scala
@@ -12,6 +12,8 @@ trait FromIterable[+CC[_]] extends Any {
 }
 
 object FromIterable {
+  import scala.language.implicitConversions
+
   implicit def toSpecific[A, CC[_]](fi: FromIterable[CC]): FromSpecificIterable[A, CC[A]] =
     new FromSpecificIterable[A, CC[A]] {
       def fromSpecificIterable(it: Iterable[A]): CC[A] = fi.fromIterable[A](it)
@@ -22,25 +24,45 @@ trait FromSpecificIterable[-A, +C] extends Any {
   def fromSpecificIterable(it: Iterable[A]): C
 }
 
+/**
+  * Builder factory
+  * @tparam A Element type (e.g. `Int`)
+  * @tparam C Collection type (e.g. `List[Int]`)
+  */
+trait CanBuild[-A, C] {
+  /** Creates a builder */
+  def newBuilder(): Builder[A, C]
+}
+
 /** Base trait for companion objects of unconstrained collection types */
-trait IterableFactory[+CC[_]] extends FromIterable[CC] {
+trait IterableFactory[CC[_]] extends FromIterable[CC] {
   def empty[A]: CC[A]
 
   def apply[A](xs: A*): CC[A] = fromIterable(View.Elems(xs: _*))
 
   def fill[A](n: Int)(elem: => A): CC[A] = fromIterable(View.Fill(n)(elem))
+
+  def newBuilder[A](): Builder[A, CC[A]]
+
+  implicit def canBuild[A]: CanBuild[A, CC[A]] = () => newBuilder[A]()
+
 }
 
-trait SpecificIterableFactory[-A, +C] extends FromSpecificIterable[A, C] {
+trait SpecificIterableFactory[-A, C] extends FromSpecificIterable[A, C] {
   def empty: C
 
   def apply(xs: A*): C = fromSpecificIterable(View.Elems(xs: _*))
 
   def fill(n: Int)(elem: => A): C = fromSpecificIterable(View.Fill(n)(elem))
+
+  def newBuilder(): Builder[A, C]
+
+  implicit def canBuild: CanBuild[A, C] = () => newBuilder()
+
 }
 
 /** Factory methods for collections of kind `* −> * -> *` */
-trait MapFactory[+CC[X, Y] <: Map[X, Y] with MapOps[X, Y, CC, _]] {
+trait MapFactory[CC[X, Y] <: Map[X, Y] with MapOps[X, Y, CC, _]] {
 
   def empty[K, V]: CC[K, V]
   def fromIterable[K, V](it: Iterable[(K, V)]): CC[K, V] =
@@ -48,9 +70,16 @@ trait MapFactory[+CC[X, Y] <: Map[X, Y] with MapOps[X, Y, CC, _]] {
 
   def apply[K, V](elems: (K, V)*): CC[K, V] =
     empty[K, V] ++ elems.toStrawman
+
+  def newBuilder[K, V](): Builder[(K, V), CC[K, V]]
+
+  implicit def canBuild[K, V]: CanBuild[(K, V), CC[K, V]] = () => newBuilder[K, V]()
+
 }
 
 object MapFactory {
+  import scala.language.implicitConversions
+
   implicit def toSpecific[K, V, CC[X, Y] <: Map[X, Y] with MapOps[X, Y, CC, _]]
       (fi: MapFactory[CC]): FromSpecificIterable[(K, V), CC[K, V]] =
     new FromSpecificIterable[(K, V), CC[K, V]] {
@@ -63,6 +92,8 @@ trait OrderedFromIterable[+CC[_]] extends Any {
 }
 
 object OrderedFromIterable {
+  import scala.language.implicitConversions
+
   implicit def toSpecific[A: Ordering, CC[_]](fi: OrderedFromIterable[CC]): FromSpecificIterable[A, CC[A]] =
     new FromSpecificIterable[A, CC[A]] {
       def fromSpecificIterable(it: Iterable[A]): CC[A] = fi.orderedFromIterable[A](it)
@@ -70,17 +101,22 @@ object OrderedFromIterable {
 }
 
 /** Base trait for companion objects of collections that require an implicit evidence */
-trait OrderedSetFactory[+CC[_]] extends OrderedFromIterable[CC] {
+trait OrderedIterableFactory[CC[_]] extends OrderedFromIterable[CC] {
 
   def empty[A : Ordering]: CC[A]
 
   def apply[A : Ordering](xs: A*): CC[A] = orderedFromIterable(View.Elems(xs: _*))
 
   def fill[A : Ordering](n: Int)(elem: => A): CC[A] = orderedFromIterable(View.Fill(n)(elem))
+
+  def newBuilder[A : Ordering](): Builder[A, CC[A]]
+
+  implicit def canBuild[A : Ordering]: CanBuild[A, CC[A]] = () => newBuilder[A]()
+
 }
 
 /** Factory methods for collections of kind `* −> * -> *` which require an implicit evidence value for the key type */
-trait OrderedMapFactory[+CC[X, +Y] <: SortedMap[X, Y] with SortedMapOps[X, Y, CC, _]] {
+trait OrderedMapFactory[CC[X, Y] <: SortedMap[X, Y] with SortedMapOps[X, Y, CC, _]] {
 
   def empty[K : Ordering, V]: CC[K, V]
 
@@ -89,4 +125,9 @@ trait OrderedMapFactory[+CC[X, +Y] <: SortedMap[X, Y] with SortedMapOps[X, Y, CC
 
   def apply[K : Ordering, V](elems: (K, V)*): CC[K, V] =
     empty[K, V] ++ elems.toStrawman
+
+  def newBuilder[K : Ordering, V](): Builder[(K, V), CC[K, V]]
+
+  implicit def canBuild[K : Ordering, V]: CanBuild[(K, V), CC[K, V]] = () => newBuilder[K, V]()
+
 }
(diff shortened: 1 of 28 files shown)


src/main/scala/strawman/collection/Factories.scala
```diff
@@ -22,35 +24,62 @@ trait FromSpecificIterable[-A, +C] extends Any {
...
+trait IterableFactory[CC[_]] extends FromIterable[CC] {
   def empty[A]: CC[A]
 
   def apply[A](xs: A*): CC[A] = fromIterable(View.Elems(xs: _*))
 
   def fill[A](n: Int)(elem: => A): CC[A] = fromIterable(View.Fill(n)(elem))
+
+  def newBuilder[A](): Builder[A, CC[A]]
+
+  implicit def canBuild[A]: CanBuild[A, CC[A]] = () => newBuilder[A]()
```
julienrf: Our four flavors of factories (iterable, ordered iterable, map and ordered map) now have a new abstract method `newBuilder` and an implemented implicit member providing a `CanBuild` instance.

Note that the `newBuilder` method is already existing in the current collections, so that’s not really a new thing…

Also note that I added parenthesis to the method, to emphasize the fact that it returns different values each time it is called.

src/main/scala/strawman/collection/mutable/ImmutableBuilder.scala
```diff
@@ -0,0 +1,19 @@
...
+abstract class ImmutableBuilder[-Elem, Coll](empty: Coll)
+  extends ReusableBuilder[Elem, Coll] {
+
+  protected var elems: Coll = empty
+
+  def clear(): Unit = { elems = empty }
+
+  def result(): Coll = elems
+
+}
```
julienrf: I also re-introduced this class which is useful to define builders for immutable collections.

src/test/scala/strawman/collection/test/GenericTest.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  @Test
+  def genericTest: Unit = {
+    assert(implicitly[Parse[immutable.List[Int]]].parse("1|2|3").contains(1 :: 2 :: 3 :: immutable.Nil))
+
+    // TODO wrap with assert when HashMap’s equality is correctly implemented
+    implicitly[Parse[immutable.HashMap[Int, Int]]].parse("1-2|3-4").contains(immutable.HashMap((1, 2), (3, 4)))
+  }
+
+}
```
julienrf: This test shows that `CanBuild` gives enough power to address #65.

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -1,77 +1,139 @@
...
+  * @tparam A Element type
+  */
+trait BuildFrom[-From, -A] {
+
+  /** Collection type to build */
+  type To
+
+  def newBuilder(): Builder[A, To]
+
+}
```
julienrf: I re-introduced the `BuildFrom` type initially written by @szeiger. This type is used to infer the type of a collection to build (`To`) based on the type of an existing collection (`From`).

I think this use case is very specific to `Traverse` where we want to turn a `F[G[?]]` into a `G[F[?]]`, so I think it makes sense to not include it in the core of the collections (well, it will be in the standard library anyway if we want to keep `Future.sequence`).

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -1,77 +1,139 @@
...
+      case (Some(builder), Some(a)) => Some(builder += a)
+      case _ => None
+    }.map(_.result)
+
+  def eitherSequence[CC[X] <: Iterable[X], A, B](xs: CC[Either[A, B]])(implicit bf: BuildFrom[CC[Either[A, B]], B]): Either[A, bf.To] =
+    xs.foldLeft[Either[A, Builder[B, bf.To]]](Right(bf.newBuilder())) {
+      case (Right(builder), Right(b)) => Right(builder += b)
+      case (Left(a)       ,        _) => Left(a)
+      case (_             ,  Left(a)) => Left(a)
+    }.right.map(_.result)
```
julienrf: I just uncommented these methods previously written by @szeiger.

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -1,77 +1,139 @@
...
+    * Extracts the unary type constructor of `From` and applies it to `A`
+    * to build the `To` type.
+    */
+  implicit def unaryTC[CC[_], A](implicit
+    cb: CanBuild[A, CC[A]]
+  ): BuildFrom.Aux[CC[_], A, CC[A]] =
+    new BuildFrom[CC[_], A] {
+      type To = CC[A]
+      def newBuilder(): Builder[A, CC[A]] = cb.newBuilder()
+    }
```
julienrf: This method provides a `BuildFrom` based on an available `CanBuild` instance.

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -1,77 +1,139 @@
...
-//    val xs3 = mutable.ListBuffer(Right("foo"), Left(0), Right("bar"))
-//    val xs3t: mutable.ListBuffer[Either[Int, String]] = xs3
-//    val e1 = eitherSequence(xs3)
-//    val e1t: Either[Int, mutable.ListBuffer[String]] = e1
-//  }
-//}
+  def optionSequenceTest: Unit = {
+    val xs1 = immutable.List(Some(1), None, Some(2))
+    val o1 = optionSequence(xs1)
+    val o1t: Option[immutable.List[Int]] = o1
```
julienrf: By default, the `BuildFrom` instance implicitly available produces a builder of the same collection type.

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -1,77 +1,139 @@
...
+
+    val xs2 = immutable.TreeSet(Some("foo"), Some("bar"), None)
+    val o2 = optionSequence(xs2)
+    val o2t: Option[immutable.TreeSet[String]] = o2
+
+    // Breakout-like use case from https://github.com/scala/scala/pull/5233:
+    val xs4 = immutable.List[Option[(Int, String)]](Some((1 -> "a")), Some((2 -> "b")))
+    val o4 = optionSequence(xs4)(BuildFrom.factory(immutable.TreeMap))
+    val o4t: Option[immutable.TreeMap[Int, String]] = o4
```
julienrf: But it’s possible to explicitly provide a `BuildFrom` instance derived from a collection factory (in such a case, we just ignore the `From` collection, which is a `List[Option[(Int, String)]]` in this example, and use the supplied factory to build the target collection).

src/test/scala/strawman/collection/test/UnfoldTest.scala
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.Test
+
+/**
+  * Builds a collection of type `C`.
+  *
+  * @tparam C Type of the collection to produce (e.g. `List[Int]`, `TreeMap[Int, String]`, etc.)
+  */
+trait Unfolder[C] {
+
+  val bt: BuildTo[C]
```
julienrf: These tests check that it is possible to implement generic “unfold” operations without touching the core of the collections.

We could use the same approach as in the `BuildFrom.factory` methods but that requires to write 4 methods, corresponding to the 4 variants of collection factories ((ordered vs unordered) × (unary vs binary type constructor)). Here I chose to explore a way to define one single `unfold` method that could be used to produce lists, maps and ordered collections as well. I’m relying on an auxiliary type named `BuildTo`, which plays a similar role as `BuildFrom`: provide builders in a type-directed way. In the case of `BuildTo` the goal is to get a `Builder[Int, List[Int]]` from a `BuildTo[List[Int]]`, or a `Builder[(Int, String), TreeMap[Int, String]]` from a `BuildTo[TreeMap[Int, String]]`.

src/test/scala/strawman/collection/test/UnfoldTest.scala
```diff
@@ -0,0 +1,98 @@
...
+      type Elem = A
+      def newBuilder(): Builder[A, CC[A]] = cb.newBuilder()
+    }
+
+  /** Provides a `BuildTo` based on an available `CanBuild` for a binary collection type constructor */
+  implicit def binaryTC[CC[_, _], A, B](implicit cb: CanBuild[(A, B), CC[A, B]]): BuildTo[CC[A, B]] { type Elem = (A, B) } =
+    new BuildTo[CC[A, B]] {
+      type Elem = (A, B)
+      def newBuilder(): Builder[(A, B), CC[A, B]] = cb.newBuilder()
+    }
```
julienrf: We need these two rules (`unaryTC` and `binaryTC`) to extract the element types and collection type constructors in the cases of unary type constructors and binary type constructors. These rules rely on the available implicit `CanBuild` instances, which are provided by the core.

src/test/scala/strawman/collection/test/UnfoldTest.scala
```diff
@@ -0,0 +1,98 @@
...
+
+}
+
+
+class UnfoldTest {
+
+  @Test
+  def unfoldTest: Unit = {
+    implicitly[BuildTo[immutable.List[Int]]]
+    implicitly[BuildTo[immutable.TreeMap[Int, String]]]
```
julienrf: We can then request instances of `BuildTo` of different collection types.

julienrf:
I just pushed more test cases drawn from #44.

The tests do not compile with Dotty. I think that’s because of `Either` extending `Product with Serializable`? (you can see the error message [here](https://travis-ci.org/scala/collection-strawman/builds/231684212#L2453))

src/main/scala/strawman/collection/Factories.scala
odersky: I think this is problematic. Not all collections admit a builder.E.g. LazyList does not. Also, I am 
not happy that every factory now has two ways to construct things, `fromIterable` and `newBuilder`. What argument is there that we are not getting back into the complexity we just struggled to get out of?

odersky:
I am skeptical for reasons of complexity. If we decide to do this I think it would be better to put the functionality in its own class, instead of adding it to all collection factories.

odersky:
One alternative suggestion would be to try to make `fromIterable` public in `Iterable`, and have `IterableOps` extend `FromIterable` again. This would mean that given an iterables `xs` and `ys`, we could convert `ys` to an iterable of the same class as `xs` simply by writing: `ys.to(xs)`. 

If we have that we can always build with something general, like an `ArrayBuffer` and then convert to the target collection type. It's an added intermediate collection, but maybe that does not matter in the grand scheme of things?

julienrf:
> I am skeptical for reasons of complexity. If we decide to do this I think it would be better to put the functionality in its own class, instead of adding it to all collection factories.

Then the drawback is that support for any new collection type will have to be added explicitly, whereas with this PR factories force collection implementations to support the `CanBuild` thing.

> One alternative suggestion would be to try to make `fromIterable` public in `Iterable`, and have `IterableOps` extend `FromIterable` again. This would mean that given iterables `xs` and `ys`, we could convert `ys` to an iterable of the same class as `xs` simply by writing `ys.to(xs)`. 
> 
> If we have that we can always build with something general, like an `ArrayBuffer` and then convert to the target collection type. It's an added intermediate collection, but maybe that does not matter in the grand scheme of things?

I think that’s more or less what @szeiger initially did. The drawback of this solution was that you had to provide an initial collection instance (the `xs` in your example, IIUC) in order to build a new collection, whereas with this PR you can build a collection out of nothing.

src/main/scala/strawman/collection/Factories.scala
julienrf: `newBuilder` is more general than `fromIterable`.

That being said, I agree that it makes no sense for `LazyList`…

julienrf: `fromIterable` makes not more sense on `LazyList`, does it?

Also, we do use `Builder` in some other places (like in the `Buildable` trait), so anyway we are going to keep it.

szeiger: Some operations are inherently strict. This includes basic operations like grouping and sorting that make sense for all sequence types. I think it is generally desirable that they use the original collection type rather than always building a fixed type. Consequently we have to be able to build even a `LazyList` in a strict way. This requires copying the elements at some points but the same is true for other immutable collections and their builders. The result could still be a "proper" `LazyList`, albeit one whose elements have already been evaluated.

szeiger:
No, only the implicit instances (for automatically building an instance of the source collection again) in my design required a source collection. The implicit *conversions* from the factory objects did not use the source object (only requiring `Any` as the source type). That way you cannot ask to build a collection of a specific *type* but you can ask to build a collection from its *factory*.

src/main/scala/strawman/collection/Factories.scala
szeiger: (That being said, `fromIterable` does make more sense on `LazyList` because it can be implemented lazily, consuming an `Iterator` of the `Iterable` element by element on demand.)

PR #83, opened 2017-05-15
Harmonize factories

julienrf:
- Removes the `FromIterable` trait, which was only inherited from `IterableFactory`, and harmonizes the names used in the `Factories.scala` file.
- Systematically uses delegate factories for abstract collections.
- Factors out the implementation of `fromIterable` for mutable collections (into a `Growable.fromIterable` method).

### DIFF
--- a/src/main/scala/strawman/collection/mutable/HashMap.scala
+++ b/src/main/scala/strawman/collection/mutable/HashMap.scala
@@ -94,11 +94,7 @@ object HashMap extends MapFactory[HashMap] {
 
   def empty[K, V]: HashMap[K, V] = new HashMap[K, V]
 
-  def fromIterable[K, V](it: collection.Iterable[(K, V)]): HashMap[K, V] =
-    it match {
-      case hm: HashMap[K, V] => hm
-      case _ => empty ++= it
-    }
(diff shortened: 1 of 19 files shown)


src/main/scala/strawman/collection/mutable/HashMap.scala
```diff
@@ -94,11 +94,7 @@ object HashMap extends MapFactory[HashMap] {
 
   def empty[K, V]: HashMap[K, V] = new HashMap[K, V]
 
-  def fromIterable[K, V](it: collection.Iterable[(K, V)]): HashMap[K, V] =
-    it match {
-      case hm: HashMap[K, V] => hm
-      case _ => empty ++= it
-    }
+  def fromIterable[K, V](it: collection.Iterable[(K, V)]): HashMap[K, V] = Growable.fromIterable(empty[K, V], it)
```
julienrf: Note that the type test (the `case hm: HashMap[K, V] => hm` line above) has been removed. This test is needed to prevent infinite recursion in the case of immutable collections because the implementation of `fromIterable` might rely on `fromIterable`. I don’t think this can be a problem for mutable collections.

odersky:
LGTM. Nice simplification!

szeiger:
Yes, `FromIterable` was already looking rather unnecessary in my work on `BuildFrom`. Merging it into the factories (and keeping those separate from the collection types) makes sense.

PR #85, opened 2017-05-16
Expose factories

odersky:
Every collection now has a `iterableFactory` method that returns the `IterableFactory` associated with the collection type. `fromIterable` now is defined only once in terms of `IterableFactory`. 

This lets us revive fully generic forms of `TraverseTest`.

We could go further and have a subset of Factories that have `newBuilder` methods. (Lazy collections should not offer such a method). We could then avoid using an intermediate collection in traverse test for the building, and instead use the builder of the target collection directly.

### DIFF
--- a/src/main/scala/strawman/collection/Factories.scala
+++ b/src/main/scala/strawman/collection/Factories.scala
@@ -74,54 +74,54 @@ object MapFactory {
 }
 
 /** Base trait for companion objects of collections that require an implicit evidence */
-trait OrderedIterableFactory[+CC[_]] {
+trait SortedIterableFactory[+CC[_]] {
 
-  def orderedFromIterable[E : Ordering](it: Iterable[E]): CC[E]
+  def sortedFromIterable[E : Ordering](it: Iterable[E]): CC[E]
 
   def empty[A : Ordering]: CC[A]
 
-  def apply[A : Ordering](xs: A*): CC[A] = orderedFromIterable(View.Elems(xs: _*))
+  def apply[A : Ordering](xs: A*): CC[A] = sortedFromIterable(View.Elems(xs: _*))
 
-  def fill[A : Ordering](n: Int)(elem: => A): CC[A] = orderedFromIterable(View.Fill(n)(elem))
+  def fill[A : Ordering](n: Int)(elem: => A): CC[A] = sortedFromIterable(View.Fill(n)(elem))
 }
 
-object OrderedIterableFactory {
+object SortedIterableFactory {
   import scala.language.implicitConversions
 
-  implicit def toSpecific[A: Ordering, CC[_]](factory: OrderedIterableFactory[CC]): FromSpecificIterable[A, CC[A]] =
+  implicit def toSpecific[A: Ordering, CC[_]](factory: SortedIterableFactory[CC]): FromSpecificIterable[A, CC[A]] =
     new FromSpecificIterable[A, CC[A]] {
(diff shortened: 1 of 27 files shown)


odersky:
Need to get dotty #2435 in, then it should build OK.

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -29,8 +29,12 @@ trait Iterable[+A] extends IterableOnce[A] with IterableOps[A, Iterable, Iterabl
 trait IterableOps[+A, +CC[X], +C] extends Any {
 
   protected def coll: Iterable[A]
+
   protected[this] def fromSpecificIterable(coll: Iterable[A]): C
-  protected[this] def fromIterable[E](it: Iterable[E]): CC[E]
+
+  def iterableFactory: IterableFactory[CC]
```
szeiger: This is a tempting design that I considered on several occasions but always backed out because that factory is limited to building unconstrained unary types. It looks promising the way it's used here. I'll have to take a closer look for potential problems I ran into when I tried it.

julienrf: Yes, I’d also like to stress that the solution proposed in this PR is very simple but does not support the use cases described in #65 and might require substantial changes to support https://github.com/scala/scala/pull/5233.

julienrf:
I rebased this PR and introduced a subtype of `IterableFactory` named `IterableFactoryWithBuilder`. This factory type is intended to be implemented by any strict collection that supports a `newBuilder` operation.

This new type of factory makes it possible to use a builder that targets the right collection type, in `TraverseTest` (instead of using an intermediate `ArrayBuffer`, which is the default behavior).

src/main/scala/strawman/collection/Factories.scala
```diff
@@ -26,6 +28,10 @@ trait IterableFactory[+CC[_]] {
 
 }
 
+trait IterableFactoryWithBuilder[+CC[_]] extends IterableFactory[CC] {
+  def newBuilder[A](): Builder[A, CC[A]]
+}
```
julienrf: As explained in the document at the bottom of the diff, it would be useful to have more power and also add the following method:

~~~ scala
implicit def canBuild[A]: CanBuild[A, CC[A]] = () => newBuilder[A]
~~~

Where `CanBuild` is the type defined in #80. If we add this line and if we also add `XxxWithBuilder` factories for the other kind of factories (`SortedIterable`, `Map` and `SortedMap`), we would be able to implement the tests given in #80.

julienrf: I implemented this suggestion in #91.

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -1,77 +1,53 @@
...
+  def optionSequence[CC[X] <: Iterable[X] with IterableOps[X, CC, _], A](xs: CC[Option[A]]): Option[CC[A]] = {
+    def folder[F[X] <: Growable[X]]: (Option[F[A]], Option[A]) => Option[F[A]] = { (bo, xo) =>
+      (bo, xo) match {
+        case (Some(builder), Some(a)) => Some(builder += a)
+        case _ => None
+      }
+    }
+    val factory = xs.iterableFactory
+    factory match {
+      case iterableBuilder: IterableFactoryWithBuilder[CC] =>
```
szeiger: The type of `IterableFactoryWithBuilder` is rather unfortunate. Using a type constructor argument prevents you from writing a pattern match that doesn't generate a warning.

PR #87, opened 2017-05-22
Support for immutable arrays

szeiger:
Based on https://github.com/scala/collection-strawman/pull/31, updated
for the new design.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ libraryDependencies += "ch.epfl.scala" %% "collection-strawman" % "0.2.0-SNAPSHO
 - [x] `ArrayOps`
 - [x] `StringOps`
 - [x] `ArrayBuffer`
+- [x] `ImmutableArray` (new)
 - [x] `ListBuffer`
 - [ ] `UnrolledBuffer`
 - [ ] `LinkedList`
--- a/benchmarks/memory/src/main/scala/strawman/collection/MemoryFootprint.scala
+++ b/benchmarks/memory/src/main/scala/strawman/collection/MemoryFootprint.scala
@@ -41,11 +41,13 @@ object MemoryFootprint extends App {
       "List"          -> benchmark(List.fill(_)(obj)),
       "LazyList"      -> benchmark(LazyList.fill(_)(obj)),
       "scala.HashSet" -> benchmark(n => scala.collection.immutable.HashSet((1 to n).map(_.toString): _*)),
-      "HashSet"       -> benchmark(n => collection.immutable.HashSet((1 to n).map(_.toString): _*)),
+      "HashSet"       -> benchmark(n => strawman.collection.immutable.HashSet((1 to n).map(_.toString): _*)),
       "scala.TreeSet" -> benchmark(n => scala.collection.immutable.TreeSet((1 to n).map(_.toString): _*)),
-      "TreeSet"       -> benchmark(n => collection.immutable.TreeSet((1 to n).map(_.toString): _*)),
+      "TreeSet"       -> benchmark(n => strawman.collection.immutable.TreeSet((1 to n).map(_.toString): _*)),
       "ArrayBuffer"   -> benchmark(ArrayBuffer.fill(_)(obj)),
-      "ListBuffer"    -> benchmark(ListBuffer.fill(_)(obj))
+      "ListBuffer"    -> benchmark(ListBuffer.fill(_)(obj)),
+      "immutable.Array" -> benchmark(strawman.collection.immutable.ImmutableArray.fill(_)(obj)),
+      "immutable.Array (primitive)" -> benchmark(strawman.collection.immutable.ImmutableArray.fill(_)(123))
     )
 
   // We use a format similar to the one used by JMH so that
--- a/src/main/scala/strawman/collection/Seq.scala
+++ b/src/main/scala/strawman/collection/Seq.scala
@@ -67,10 +67,10 @@ trait SeqOps[+A, +CC[X], +C] extends Any
 }
 
 /** Base trait for indexed Seq operations */
-trait IndexedSeqOps[+A, +CC[X] <: IndexedSeq[X], +C] extends Any with SeqOps[A, CC, C] {
+trait IndexedSeqOps[+A, +CC[X] <: IndexedSeq[X], +C] extends Any with SeqOps[A, CC, C] { self =>
   override def view: IndexedView[A] = new IndexedView[A] {
-    def length: Int = coll.length
-    def apply(i: Int): A = coll(i)
+    def length: Int = self.length
+    def apply(i: Int): A = self(i)
   }
   override def reverse: C = fromSpecificIterable(view.reverse)
 }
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -36,6 +36,7 @@ libraryDependencies += "ch.epfl.scala" %% "collection-strawman" % "0.2.0-SNAPSHO
 - [x] `ArrayOps`
 - [x] `StringOps`
 - [x] `ArrayBuffer`
+- [x] `ImmutableArray` (new)
```
szeiger: I renamed `immutable.Array` to `immutable.ImmutableArray`. The redundancy seems less worrisome to me than reusing the name `Array`, which is in scope everywhere as `scala.Array`. WDYT?

src/main/scala/strawman/collection/Seq.scala
```diff
@@ -67,10 +67,10 @@ trait SeqOps[+A, +CC[X], +C] extends Any
 }
 
 /** Base trait for indexed Seq operations */
-trait IndexedSeqOps[+A, +CC[X] <: IndexedSeq[X], +C] extends Any with SeqOps[A, CC, C] {
+trait IndexedSeqOps[+A, +CC[X] <: IndexedSeq[X], +C] extends Any with SeqOps[A, CC, C] { self =>
   override def view: IndexedView[A] = new IndexedView[A] {
-    def length: Int = coll.length
-    def apply(i: Int): A = coll(i)
+    def length: Int = self.length
+    def apply(i: Int): A = self(i)
```
szeiger: Non-`Iterable` collections usually implement `coll` through a `View`, so the old default implementation here ran into an endless recursion.

benchmarks/memory/src/main/scala/strawman/collection/MemoryFootprint.scala
```diff
@@ -41,11 +41,13 @@ object MemoryFootprint extends App {
       "List"          -> benchmark(List.fill(_)(obj)),
       "LazyList"      -> benchmark(LazyList.fill(_)(obj)),
       "scala.HashSet" -> benchmark(n => scala.collection.immutable.HashSet((1 to n).map(_.toString): _*)),
-      "HashSet"       -> benchmark(n => collection.immutable.HashSet((1 to n).map(_.toString): _*)),
+      "HashSet"       -> benchmark(n => strawman.collection.immutable.HashSet((1 to n).map(_.toString): _*)),
       "scala.TreeSet" -> benchmark(n => scala.collection.immutable.TreeSet((1 to n).map(_.toString): _*)),
-      "TreeSet"       -> benchmark(n => collection.immutable.TreeSet((1 to n).map(_.toString): _*)),
+      "TreeSet"       -> benchmark(n => strawman.collection.immutable.TreeSet((1 to n).map(_.toString): _*)),
```
szeiger: This didn't compile in master. We should compiler benchmarks as part of PR validation to prevent these regressions in the future.

src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -0,0 +1,102 @@
...
+  override def take(n: Int): ImmutableArray[A] = ImmutableArray.tabulate(n)(apply)
+
+  override def drop(n: Int): ImmutableArray[A] = ImmutableArray.tabulate((length - n) max 0)(i => apply(n + i))
+
+  override def tail: ImmutableArray[A] =
+    if (length > 0) {
+      val dest = scala.Array.ofDim[Any](length - 1)
+      java.lang.System.arraycopy(elements, 1, dest, 0, length - 1)
+      new ImmutableArray(dest)
+    } else Nil.tail
```
odd: This fails on Dotty with: 
```
[E007] Type Mismatch Error: /Users/odd/DropboxTraveas/Code/collection-strawman/src/main/scala/strawman/collection/immutable/ImmutableArray.scala:77:15
[error] 77 |    } else Nil.tail
[error]    |           ^^^^^^^^
[error]    |           found:    strawman.collection.immutable.List[scala.Nothing]
[error]    |           required: strawman.collection.immutable.ImmutableArray[A]
```

Replacing with: 
```scala
    } else ???
```
 (à la immutable.List) avoids the error.

README.md
SethTisue: agree

src/main/scala/strawman/collection/immutable/ImmutableArray.scala
julienrf: We should explicitly override the return type of `tail` on `Nil` to be `Nothing` rather than the inherited `List[A]`.

src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -0,0 +1,102 @@
...
+  override def take(n: Int): ImmutableArray[A] = ImmutableArray.tabulate(n)(apply)
+
+  override def drop(n: Int): ImmutableArray[A] = ImmutableArray.tabulate((length - n) max 0)(i => apply(n + i))
+
+  override def tail: ImmutableArray[A] =
+    if (length > 0) {
+      val dest = scala.Array.ofDim[Any](length - 1)
+      java.lang.System.arraycopy(elements, 1, dest, 0, length - 1)
+      new ImmutableArray(dest)
+    } else ???
```
julienrf: You should throw an `UnsuportedOperationException` here.

szeiger: That goes for other cases, too. I left it at `???` for now because we're doing the same in `List`. We need to check for all occurrences and replace them with the correct exception.

src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -0,0 +1,102 @@
...
+  override def map[B](f: A => B): ImmutableArray[B] = ImmutableArray.tabulate(length)(i => f(apply(i)))
+
+  override def flatMap[B](f: A => IterableOnce[B]): ImmutableArray[B] = ImmutableArray.fromIterable(View.FlatMap(coll, f))
+
+  def :+ [B >: A](elem: B): ImmutableArray[B] = {
+    val dest = scala.Array.ofDim[Any](length + 1)
+    java.lang.System.arraycopy(elements, 0, dest, 0, length)
+    dest(length) = elem.asInstanceOf[Any]
+    new ImmutableArray(dest)
+  }
```
szeiger: Not copying every time would require a more elaborate copy-on-write implementation with implications for multi-threaded sharing. It would also lead to unpredictable performance in many cases and prevent use cases (like for varargs) where we need access to the underlying array without copying.

Such a data structure may be useful but it shouldn't be part of a simple `ImmutableArray`.

odd: That's precisely the use case Spandex (#52) is designed for. It also has a method that wraps an existing array without copying (its package private since it is unsafe if the array is later modified from the outside) so it should support usage for varargs. The performance is only unpredictable in so far as an append or prepend operation will result in an array copy only when the array is full on the affected side or if the targeted array slot is already occupied (by a non eq value/reference).

PR #88, opened 2017-05-22
Steque 2

edmundnoble:
This version of the Steque PR removes `OfSeq` and adds benchmarks equivalent to the ones provided by `ListBenchmark`. To see the benefits clearly, there needs to be a variation of the `concat`, `snoc` and `cons` benchmarks which actually tests the `foldLeft` performance, because `snoc`, `cons` and `concat` are amortized O(1) and their cost is paid off during `fold`ing. I would be very interested in the performance difference between building and folding a `Steque` and building and folding a `Spandex`, especially using different building strategies. `Steque` also produces less garbage than `Spandex` and is possible to use multiple times without forfeiting the performance benefits of `Spandex`'s guarded mutation.

Edit: I am also proposing that this be considered for the default collection in Scala 2.13.

### DIFF
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/StequeBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/StequeBenchmark.scala
@@ -1,0 +1,90 @@
+package strawman.collection.immutable
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+
+import scala.{Any, AnyRef, Int, Long, Unit}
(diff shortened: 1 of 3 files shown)


julienrf:
Hi @edmundnoble, in the last PR you told that the performance of this datastructure was not better than `InceptionQ` (https://github.com/scala/collection-strawman/pull/51#issuecomment-294622903). Can you elaborate on that point? Has anything changed in the meantime?

edmundnoble:
Yes, InceptionQ is a double-ended queue variation of Steque but I haven't seen a version with a tail-recursive fold, which is severely limiting.

SethTisue:

PR #91, opened 2017-05-24
Introduce CanBuild on top of #85

julienrf:


julienrf:
I’m still not sure that’s the best solution. I’m going to experiment with another solution requiring users to explicitly pass the `SpecificIterableFactory` (as we do with the `.to` method)

julienrf:
I gave a try to the solution explained in the above comment but I’m not super happy with it. In short, I still think that providing implicit `CanBuild` instances in factories (as it is done in this PR) would provide the best developer experience.

You can see the diff here: https://github.com/scala/collection-strawman/compare/expose-factories...expose-factories-bis

The problem I wanted to solve was the ability to have `optionSequence` return sorted collections or map collections (in the `expose-factories` branch this is not possible). By taking the `FromSpecificIterable` as an explicit parameter users can supply whatever factory object they want (e.g. `TreeSet`, `HashMap`, etc.), however some problems still remain:
- because of the way type inference works this parameter has to be the last one, which is not always convenient ;
- the optimization that uses the “specific” builder, when available, instead of an intermediate `ArrayBuffer`, requires a type cast ;
- in practice users would need to versions of the methods, one with the explicit factories and one based on `iterableFactory`.

/cc @szeiger @odersky

PR #94, opened 2017-05-31
Add Scala 2.13.0-M1 to scalaCrossVersions.

odd:
First work on #93 (memoryBenchmark does not compile however since no 2.13 artifact exists for the jawn dependency).


julienrf:
Thanks @odd!

SethTisue:
this is from the Scala open source spree, Copenhagen 2017

Issue #95, opened 2017-05-31
Add missing operations


odd:
Add methods from the existing collections library that are missing in strawman-collections.

PR #96, opened 2017-05-31
Add missing operations

odd:
Added init, takeRight and dropRight #95.


src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -206,6 +206,12 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
     drop(1)
   }
 
+  /** The initial part of the collection without its last element. */
+  def init: C = {
+    if (coll.isEmpty) throw new UnsupportedOperationException
+    take(size - 1)
```
Ichoran: Without `dropRight`, this is the canonical implementation.  However, with `dropRight`, I think this should be `dropRight(1)`.  Maybe leave a comment here to prompt revisiting this once we have `dropRight`?

Ichoran:
Tests would be lovely!

SethTisue:
this is from the Scala open source spree, Copenhagen 2017

src/main/scala/strawman/collection/Iterable.scala
julienrf: @odd Could you also introduce `dropRight` (implemented in terms of `take`), as suggested by @Ichoran?

Also, could you check that all the collection types that can provide a more efficient implementation of `dropRight` or `init` do override these operations?

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +119,27 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+    new Iterator[A]() {
+      override def hasNext: Boolean = m > 0
+      override def next(): A = {
+        val value = buffer(i % l).asInstanceOf[A]
+        i = (i + 1) % l
+        m -= 1
+        value
+      }
+    }
+  }
```
julienrf: You don’t need an intermediate copy if you create two iterators: see how it is implemented in the [current collections](https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/IterableLike.scala#L209-L227).

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -127,6 +149,25 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+      override def hasNext: Boolean = m > 0 && self.hasNext
+      override def next(): A = {
+        val value = buffer(i % m).asInstanceOf[A]
+        i = (i + 1) % m
+        buffer.update(i, self.next())
+        value
+      }
+    }
+  }
+
```
julienrf: A similar trick can be applied to implement `dropRight`

src/main/scala/strawman/collection/Iterator.scala
odd: But that implementation is in `Iterable` where you can summon a new iterator for the collection in question (which means that the collection will be traversed twice). This implementation is in `Iterator` so the same double-iterator method will not work (since iterators are only traversable once). Instead a buffer of size `n` for `takeRight(n)` and `dropRight(n)` is used. 

Should I change the implementations of `takeRight`/`dropRight` in `Iterable` to use the double-iteration method from the current collections (instead of delegating to `View.TakeRight`/`View.DropRight`, which in turn delegates to this implementation in `Iterator`)?

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +119,27 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
       else Iterator.empty.next()
   }
 
+  def takeRight(n: Int): Iterator[A] = {
+    val buffer = new Array[Any](n)
+    var i = 0
+    var m = 0
+    while (n > 0 && self.hasNext) {
```
Ichoran: You don't need n > 0 here.

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +119,27 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+      buffer.update(i, self.next())
```
Ichoran: Use the standard `buffer(i) = self.next` notation?

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +119,27 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+      i = (i + 1) % n
```
Ichoran: Ow, modulus every iteration?  A branch is usually considerably faster.  `if (i >= n) 0 else i+1`

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +119,27 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+      if (m < n) m += 1
+    }
+    val l = m
+    new Iterator[A]() {
+      override def hasNext: Boolean = m > 0
+      override def next(): A = {
+        val value = buffer(i % l).asInstanceOf[A]
```
Ichoran: Again, a simple if-statement is better than the modulus

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -127,6 +149,25 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+      buffer.update(i, self.next())
+      i += 1
+    }
+    val m = i
+    new Iterator[A]() {
+      override def hasNext: Boolean = m > 0 && self.hasNext
+      override def next(): A = {
+        val value = buffer(i % m).asInstanceOf[A]
+        i = (i + 1) % m
+        buffer.update(i, self.next())
```
Ichoran: Same comments about style as in `takeRight`

src/main/scala/strawman/collection/Iterator.scala
Ichoran: I'm not sure whether the buffering approach should be avoided in `Iterable` but certainly it should in classes which have an efficient way to access the last `n` elements.

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +119,27 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
       else Iterator.empty.next()
   }
 
+  def takeRight(n: Int): Iterator[A] = {
+    val buffer = new Array[Any](n)
```
Ichoran: `takeRight(Int.MaxValue)` should be a no-op, not consume all available memory.  For less extreme things, it's okay to have some overhead if you can't know the length, but eagerly allocating a huge chunk is not so good.  There should be a reasonable default size and then it should grow until it reaches the full requested size.

src/main/scala/strawman/collection/Iterator.scala
julienrf: Oh yes, good point! Then I think your approach sounds good. We should probably override `drop`, `take`, `takeRight` and `dropRight` in `IndexedSeq` for performance. What do others think?

src/main/scala/strawman/collection/Seq.scala
```diff
@@ -72,6 +72,14 @@ trait IndexedSeqOps[+A, +CC[X] <: IndexedSeq[X], +C] extends Any with SeqOps[A,
     def length: Int = self.length
     def apply(i: Int): A = self(i)
   }
+
+  /** A collection containing the last `n` elements of this collection. */
+  override def takeRight(n: Int): C = fromSpecificIterable(view.takeRight(n))
```
julienrf: This override has exactly the same implementation as the parent, or am I missing something?

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +119,32 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
       else Iterator.empty.next()
   }
 
+  def takeRight(n: Int): Iterator[A] = {
```
julienrf: Would you mind adding a small comment describing your algorithm?

src/main/scala/strawman/collection/View.scala
```diff
@@ -167,8 +185,22 @@ object IndexedView {
     def apply(i: Int) = underlying.apply(i)
   }
 
+  class TakeRight[A](underlying: IndexedView[A], n: Int)
+  extends View.TakeRight(underlying, n) with IndexedView[A] {
+    override def iterator() = super.iterator() // needed to avoid "conflicting overrides" error
+    def length = underlying.length min normN
+    def apply(i: Int) = underlying.apply(i)
+  }
```
julienrf: I don’t understand why you need to extend `View.TakeRight` here. Also, I’m sure this implementation is correct.

src/main/scala/strawman/collection/Seq.scala
odd: The implementation in the sub class will use the view implementation returned by ```view``` which is tailored for indexed views (instead of the inherited implementation which will use the one for plain iterables). I simply made it analogous to the existing reverse method.

src/main/scala/strawman/collection/View.scala
odd: It is to provide indexed views with a more optimized implementation than the one in ```Iterator``` (which is used in the implementation inherited from ```View.TakeRight```).

src/main/scala/strawman/collection/Iterator.scala
odd: I have added comments to the ```takeRight``` and ```dropRight``` methods in ```Iterator``` describing the steps taken.

src/main/scala/strawman/collection/Seq.scala
julienrf: I would expect that the fact that the `view` is overriden to be an `IndexedView` would be enough to get the optimized implementation of `takeRight`. Why do we also have to override `takeRight`?

src/main/scala/strawman/collection/View.scala
julienrf: I understand that we want `IndexedViews` to implement an optimized version of `TakeRight` but I don’t get why this implementation would have to extend `View.TakeRight`.

src/main/scala/strawman/collection/Seq.scala
odd: Hmm, the implementation inherited from `IterableOps` is this: 
```
  def dropRight(n: Int): C = fromSpecificIterable(View.DropRight(coll, n))
```
which does not seem to make use of the `view` method but instead delegate to `View` directly. If that was changed to:
```
  def dropRight(n: Int): C = fromSpecificIterable(view.dropRight(n))
```  
it would not have to be overridden in `IndexedSeqOps` any more to make use of the overridden `view` method, but I might be misunderstanding the intended inheritance design (if I'm not, all similar methods in `IterableOps` should probably switch from using `View` to `view` to allow sub traits to specialize the view type).

src/main/scala/strawman/collection/View.scala
odd: Do you mean that it should extend `View.Take` instead? Perhaps it should, I made it analogous to the way `Take` and `Drop` were implemented as extending a tailored view sub class in the `View` companion.

src/main/scala/strawman/collection/Seq.scala
julienrf: Ah thanks, I missed that difference between both implementations. Indeed imho that would make sense to use `view.xxx` everywhere in place of `View.Xxx`. That might have an impact on performance, since that’s one more indirection level but I wonder if the JVM can always inline that.

src/main/scala/strawman/collection/View.scala
julienrf: Well, are we really reusing something from `View.TakeRight`? I think we sould just inherit from `IndexedView`.

src/main/scala/strawman/collection/Seq.scala
odd: Should I change so that the `view` method is used for all the operations? If so, in this PR or a new PR?

julienrf: We could do that in a separate PR!

odd: My proposed change won't work since it leads to infinite loops via `View` and `IterableOps` calling each other. The call in `IterableOps` need to call a method that is not delegating back to the implementation inherited from `IterableOps` (which the methods in `View` do, but the ones in the operation specific `View` subclass do not, since they call the same operation on the underlying collection's iterator instead).

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +119,37 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+      var index = 0
+      // First iterate over all elements while keeping track of the last n
+      while (self.hasNext) {
+        if (index >= buffer.length) buffer += self.next()
+        else buffer(index) = self.next()
+        index = if ((index + 1) >= n) 0 else index + 1
+        if (count < n) count += 1
+      }
+      // Adjust the starting index if needed
+      index = if (index >= buffer.length) 0 else index
```
szeiger: I'd put this at the beginning of `hasNext`

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +119,37 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+      // Return an iterator for the elements in the buffer starting from index
+      new Iterator[A]() {
+        override def hasNext: Boolean = count > 0
+        override def next(): A = {
+          val value = buffer(index)
+          index = if (index + 1 >= buffer.length) 0 else index + 1
```
szeiger: ...then you wouldn't need another copy of it here and could simply do `index += 1`

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +119,37 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
       else Iterator.empty.next()
   }
 
+  /*
+   * Implemented by means of a buffer to keep track of the last n elements during iteration.
+   */
+  def takeRight(n: Int): Iterator[A] = {
+    if (n == 0) Iterator.empty
+    else {
+      val buffer = ArrayBuffer[A]()
```
szeiger: It may be beneficial to set an initial size for the buffer, at least when `n` is less than the default buffer size, possibly also when it is within a reasonable size (e.g. default buffer size is 16, AFAIR, but we could pre-allocate for `n` up to 64 or so)

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -127,6 +159,34 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
     this
   }
 
+  /*
+   * Implemented by means of a buffer to keep the last n elements from being returned during iteration.
+   */
+  def dropRight(n: Int): Iterator[A] = {
```
szeiger: Should we have this kind of operation that requires buffering at all on `Iterator`? And if so, what is the expectation when the buffering happens? Upon construction of the `Iterator` as implemented here, or lazily on the first access?

julienrf: I think that if we want to avoid buffering we have to perform two traversals. I’m not sure which solution is the best in terms of execution times.

Ichoran: You cannot in general re-traverse an iterator, so you can't assume you can "perform two traversals".  Furthermore, even if you can, depending on what it is, it can be very expensive to traverse a second time (e.g. if it's doing some sort of database lookup).  So in the absence of other type information, it has to be buffered.
The problem with making it not on `Iterator` at all, @szeiger, is that you want to make `takeRight` and `dropRight` both be present or not together, and `takeRight` can't be done efficiently on an iterator with anything but a bare array.  But you can't convert to an array first because that isn't lazy.  So you're really stuck: the algorithm needs to be tuned to `Iterator` to be efficient, and there's no other place to put it unless we add a subclass like `BufferedIterator` to isolate the methods that use non-lazy buffering.

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -127,6 +159,34 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+      // First fill the buffer with the first n elements (or fewer if the n is greater than the iterator length)
+      while (index < n && self.hasNext) {
+        buffer += self.next()
+        index += 1
+      }
+      index = 0
+      // Return an iterator that returns already buffered elements as it buffers new ones (using a buffer of most n elements)
+      new Iterator[A]() {
+        // End when the iterator is exhausted (without having returned the elements currently in the buffer since those are the ones to drop)
+        override def hasNext: Boolean = self.hasNext
```
Ichoran: Why is there an override here?  `Iterator` leaves the method abstract.

src/main/scala/strawman/collection/Iterator.scala
julienrf: @Ichoran We could implement `takeRight` and `dropRight` in `Iterable`, where we could get two `Iterator`s. But I’m not sure we can, in general, make the assumption that it’s better to buffer rather than performing two traversals, nor the inverse…

I’m happy with the buffering solution.

szeiger: I wouldn't ever want to do two traversals. We should either do buffering or not have the operations at all (like in the 2.12 collections). If we go with buffering I'd opt for a lazy implementation that doesn't buffer before the first call to `hasNext` or `next` of the resulting iterator.

julienrf: Unless I’m wrong in 2.12 we do have `takeRight` on `Iterable`: https://github.com/scala/scala/blob/v2.12.2/src/library/scala/collection/IterableLike.scala#L209-L227
(And it performs two traversals)

DarkDimius: @julienrf, your link is into IterableLike.

```
Welcome to Scala 2.12.2 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_45).
Type in expressions for evaluation. Or try :help.
scala> Array(1,2,3).iterator
res1: Iterator[Int] = non-empty iterator

scala> res1.takeRight(2)
<console>:13: error: value takeRight is not a member of Iterator[Int]
       res1.takeRight(2)
            ^
```

julienrf: Yes, that’s what I meant: we provide this operation on `Iterable` at the cost of two traversals.

odd:
I'll start work on moving the buffer initialization logic inside `hasNext`/`next` as suggested by @szeiger. Might take a couple of days though as I'm off to the sun for a week :sunglasses:

julienrf:
Wait, I’m not sure we should have a such a lazy semantic on strict collections.

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +120,44 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
 
+  /*
+   * Implemented by means of a buffer to keep track of the last n elements during iteration.
+   */
+  def takeRight(n: Int): Iterator[A] = {
+    if (n == 0) Iterator.empty
+    else {
+      // Return an iterator that iterates over the elements via a buffer
+      new Iterator[A]() {
+        private[this] var index = 0
```
julienrf: I’m curious about the `private[this]`: since we are in an anonymous class, everything is going to be private, right? Or maybe it has an impact on optimizations?

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +120,44 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+            if (index >= buf.length) buf += self.next()
+            else buf(index) = self.next()
+            index = if ((index + 1) >= n) 0 else index + 1
+            if (count < n) count += 1
+          }
+          // Adjust the starting index if needed
+          index = if (index >= buf.length) 0 else index
+          buf
+        }
+        override def hasNext: Boolean = {
```
julienrf: This is just a minor comment on style: I prefer not to have `override` explicitly written unless I really override the member. In this case, we just implement an abstract member, we don’t override it.

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -118,6 +120,44 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
       else Iterator.empty.next()
   }
 
+  /*
+   * Implemented by means of a buffer to keep track of the last n elements during iteration.
+   */
+  def takeRight(n: Int): Iterator[A] = {
+    if (n == 0) Iterator.empty
```
julienrf: The current collections support negative values for `n`. We should change this test to `if (n <= 0)` to support negative values as well.

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -127,6 +167,40 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+          while (index < n && self.hasNext) {
+            buf += self.next()
+            index += 1
+          }
+          index = 0
+          buf
+        }
+
+        override def hasNext: Boolean = {
+          // Force initialization of buffer and don't stop unitl the iterator is exhausted (without having returned the elements currently in the buffer since those are the ones to drop)
```
julienrf: Typo: “until”

src/main/scala/strawman/collection/Iterator.scala
odd: I only added `private|this]` to make it obvious that the fields are only used locally, but they could most likely be dropped without affecting performance (the optimizer will probably be able to deduce that the fields are local anyway).

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -127,6 +167,40 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
     this
   }
 
+  /*
+   * Implemented by means of a buffer to keep the last n elements from being returned during iteration.
+   */
+  def dropRight(n: Int): Iterator[A] = {
+    if (n == 0) self
```
julienrf: Here too that should be `n <= 0`

julienrf:
Thanks @odd!

PR #97, opened 2017-06-02
Retrofit `BuildFrom` into the new design

szeiger:
Extension of #85, work in progress

### DIFF
--- a/src/main/scala/strawman/collection/Factories.scala
+++ b/src/main/scala/strawman/collection/Factories.scala
@@ -20,16 +20,30 @@ trait BuildFrom[-From, -A, +C] extends Any {
   def fromSpecificIterable(from: From)(it: Iterable[A]): C
 }
 
-object BuildFrom {
-  /** Build the source collection type from an IterableOps */
-  implicit def buildFromIterableOps[C[X] <: IterableOps[X, C, _], A, E]: BuildFrom[C[A], E, C[E]] = new BuildFrom[C[A], E, C[E]] {
+object BuildFrom extends BuildFromLowPriority {
+  /** Build the source collection type from a MapOps */
+  implicit def buildFromMapOps[CC[K, V] <: Map[K, V] with MapOps[K, V, CC, _], A, B, E, F]: BuildFrom[CC[A, B], (E, F), CC[E, F]] = new BuildFrom[CC[A, B], (E, F), CC[E, F]] {
     //TODO: Reuse a prototype instance
-    def fromSpecificIterable(from: C[A])(it: Iterable[E]): C[E] = from.iterableFactory.fromIterable(it)
+    def fromSpecificIterable(from: CC[A, B])(it: Iterable[(E, F)]): CC[E, F] = from.mapFactory.fromIterable(it)
+  }
+
+  /** Build the source collection type from a SortedMapOps */
+  implicit def buildFromSortedMapOps[CC[K, V] <: SortedMap[K, V] with SortedMapOps[K, V, CC, _], A, B, E : Ordering, F]: BuildFrom[CC[A, B], (E, F), CC[E, F]] = new BuildFrom[CC[A, B], (E, F), CC[E, F]] {
+    //TODO: Reuse a prototype instance
+    def fromSpecificIterable(from: CC[A, B])(it: Iterable[(E, F)]): CC[E, F] = from.sortedMapFactory.fromSpecificIterable(it)
+  }
+}
+
+trait BuildFromLowPriority {
+  /** Build the source collection type from an IterableOps */
+  implicit def buildFromIterableOps[CC[X] <: IterableOps[X, CC, _], A, E]: BuildFrom[CC[A], E, CC[E]] = new BuildFrom[CC[A], E, CC[E]] {
+    //TODO: Reuse a prototype instance
+    def fromSpecificIterable(from: CC[A])(it: Iterable[E]): CC[E] = from.iterableFactory.fromIterable(it)
   }
 
   /** Build the source collection type from an Iterable with SortedOps */
-  implicit def buildFromSortedOps[C[X] <: Iterable[X] with SortedOps[X, C[X], C], A, E : Ordering]: BuildFrom[C[A], E, C[E]] = new BuildFrom[C[A], E, C[E]] {
-    def fromSpecificIterable(from: C[A])(it: Iterable[E]): C[E] = from.sortedIterableFactory.fromSpecificIterable(it)
+  implicit def buildFromSortedOps[CC[X] <: Iterable[X] with SortedOps[X, CC[X], CC], A, E : Ordering]: BuildFrom[CC[A], E, CC[E]] = new BuildFrom[CC[A], E, CC[E]] {
+    def fromSpecificIterable(from: CC[A])(it: Iterable[E]): CC[E] = from.sortedIterableFactory.fromSpecificIterable(it)
   }
 }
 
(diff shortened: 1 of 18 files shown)


szeiger:
`BuildFrom` is like `FromSpecificIterable` but with an extra `From`
argument, like in the final version of #45. `FromSpecificIterable`
existed conceptually in that version as `BuildFrom[Any, …]` but didn’t
have a separate type.

This new version has separate abstractions for buildable (strict)
collection types in the form of `StrictBuildFrom` and
`FromSpecificIterableWithBuilder`. Since we can get a surrogate builder
(through one of the new `Builder.from` methods) for any lazy collection
and we can restrict code to work only with strict collections via the
`Buildable` trait, this is not strictly necessary. The only reason for
separating the `Builder` abstractions is to avoid exposing them in
`FromSpecificIterable`. Even though everything can be built in a strict
way, these abstractions sit on top of the lazy ones, not below them.

szeiger:
@odersky @julienrf  I just updated this with the finished version of the design we discussed at ScalaDays.

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -3,51 +3,95 @@ package strawman
...
+      case _ => None
+    }.map(_.result)
+  def optionSequence1[C[X] <: Iterable[X] with SortedOps[X, C[X], C], A : Ordering](xs: C[Option[A]]): Option[C[A]] =
+    xs.foldLeft[Option[Builder[A, C[A]]]](Some(Builder.from[A, C](xs.sortedIterableFactory))) {
+      case (Some(builder), Some(a)) => Some(builder += a)
+      case _ => None
+    }.map(_.result)
+
+  // ...or use BuildFrom to abstract over both and also allow building arbitrary collection types
+  def optionSequence2[CC[X] <: Iterable[X], A, To](xs: CC[Option[A]])(implicit bf: BuildFrom[CC[Option[A]], A, To]): Option[To] =
```
szeiger: This should work for the tests below if you ask for a `StrictBuildFrom` instead of a `BuildFrom` but it doesn't because `Set` is not `Buildable` at the moment. I'm currently working on changing that but I ran into some complications, so I wanted to get this PR out first. Making Sets buildable is a separate concern.

src/main/scala/strawman/collection/SortedOps.scala
```diff
@@ -3,10 +3,12 @@ package strawman.collection
 import scala.{Ordering, Option, Some}
 
 /** Base trait for sorted collections */
-trait SortedOps[A, +C] {
+trait SortedOps[A, +C, +CC[_]] {
 
   implicit def ordering: Ordering[A]
 
+  def sortedIterableFactory: SortedIterableFactory[CC]
```
szeiger: This PR also adds the another factor kinds. So far every collection exposes an `IterableFactory`. Now in addition to that every sorted collection has a `SortedIterableFactory`. We'll probably want to add `MapFactory` and `SortedMapFactory` types to the appropriate collection types as well.

src/main/scala/strawman/collection/Factories.scala
```diff
@@ -1,20 +1,76 @@
...
-trait FromSpecificIterable[-A, +C] extends Any {
+trait FromSpecificIterable[-A, +C] extends Any with BuildFrom[Any, A, C] {
+  def fromSpecificIterable(from: Any)(it: Iterable[A]): C = fromSpecificIterable(it)
   def fromSpecificIterable(it: Iterable[A]): C
 }
 
+/** A more specific `FromSpecificIterable` for strict collection types which can provide a `Builder`. */
+trait FromSpecificIterableWithBuilder[-A, +C] extends Any with FromSpecificIterable[A, C] with StrictBuildFrom[Any, A, C] {
+  def newBuilder(from: Any): Builder[A, C] = newBuilder
+  def newBuilder: Builder[A, C]
```
julienrf: I prefer that we add parens to emphasize the fact that each created builder has its own identity: `def newBuilder(): Builder[A, C]`.

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -253,6 +253,10 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
 }
 
 /** Base trait for strict collections that can be built using a builder.
+  *
+  * All factories provided by such a collection must be of the `*WithBuilder` variant
+  * even though this is not formally enforced through types.
+  *
```
julienrf: This is unfortunate that we can not enforce this constraint :(

src/main/scala/strawman/collection/SortedOps.scala
```diff
@@ -3,10 +3,12 @@ package strawman.collection
 import scala.{Ordering, Option, Some}
 
 /** Base trait for sorted collections */
-trait SortedOps[A, +C] {
+trait SortedOps[A, +C, +CC[_]] {
```
julienrf: So far we usually order the type parameters such that `CC` is before `C`.

src/main/scala/strawman/collection/SortedMap.scala
```diff
@@ -12,7 +12,9 @@ trait SortedMap[K, +V]
 
 trait SortedMapOps[K, +V, +CC[X, Y] <: SortedMap[X, Y] with SortedMapOps[X, Y, CC, _], +C <: SortedMap[K, V]]
   extends MapOps[K, V, Map, C]
-     with SortedOps[K, C] {
+     with SortedOps[K, C, SortedSet] {
+
+  def sortedIterableFactory = SortedSet
```
julienrf: Does that mean that when I navigate through the `sortedIterableFactory` of a `SortedMap` then I get back a `SortedSet`?

src/main/scala/strawman/collection/immutable/TreeSet.scala
```diff
@@ -113,4 +115,5 @@ object TreeSet extends SortedIterableFactory[TreeSet] {
       case _ => empty[E] ++ it
     }
 
+  def newBuilder[A : Ordering](): Builder[A, TreeSet[A]] = new ArrayBuffer[A].mapResult(sortedFromIterable[A] _)
```
julienrf: What’s the advantage of using an intermediate `ArrayBuffer` over directly building a `TreeSet`?

src/main/scala/strawman/collection/immutable/BitSet.scala
```diff
@@ -27,6 +27,7 @@ sealed abstract class BitSet
   def empty: BitSet = BitSet.empty
 
   def iterableFactory = Set
+  def sortedIterableFactory = SortedSet
```
julienrf: Why can’t we tell that the factory of a `BitSet` collection is the `BitSet` object?

src/main/scala/strawman/collection/mutable/TreeSet.scala
```diff
@@ -175,10 +177,11 @@ sealed class TreeSet[A] private (tree: RB.Tree[A, Null])(implicit val ordering:
...
   *
   */
-object TreeSet extends SortedIterableFactory[TreeSet] {
+object TreeSet extends SortedIterableFactoryWithBuilder[TreeSet] {
 
   def empty[A : Ordering]: TreeSet[A] = new TreeSet[A]()
 
   def sortedFromIterable[E : Ordering](it: collection.Iterable[E]): TreeSet[E] = Growable.fromIterable(empty[E], it)
 
+  def newBuilder[A : Ordering](): Builder[A, TreeSet[A]] = new ArrayBuffer[A].mapResult(sortedFromIterable[A] _)
```
julienrf: Why not directly grow a `TreeSet`?

src/main/scala/strawman/collection/Iterable.scala
szeiger: We can relax it from *must* to *should* though. Some strict collections may not be able to provide a more efficient builder anyway so it would be OK to fall back to a an ArrayBuffer. Implementing a `WithBuilder` factory would be an optimization, just like overriding a default implementation of some collection operation.

src/main/scala/strawman/collection/immutable/TreeSet.scala
szeiger: Hm, right, the current collections append directly to the immutable `TreeSet`. We should do the same.

src/main/scala/strawman/collection/immutable/BitSet.scala
szeiger: Because the `SortedIterableFactory` can build a collection for every element type with an `Ordering`.

src/main/scala/strawman/collection/SortedMap.scala
szeiger: Yes. This is a side-effect of `SortedOps` being shared by sorted sets and maps. I was also baffled by this at first but it does make sense. A sorted map is sorted by key. If you assume a standard `Ordering` on tuples, it degrades nicely to a sorted set. The ordering for `(K,V)` with unique `K` values is the same and with `K` already being unique, so is `(K,V)`.

odersky:
Hi Stefan, can you push this the strawman repo? Makes it easier for me to check it out locally and play with it.

szeiger:
@odersky Done. I can confirm that this PR is the cause of the problem. Master compiles fine on the current Dotty nightly, this PR fails.

src/main/scala/strawman/collection/Factories.scala
szeiger: We have many methods without a parameter list that can return objects with different identities, for example `Iterable.tail` or `IterableFactory.empty`. The name `newBuilder` in particular makes it clear that you get a new object. I would prefer to omit empty parameter lists wherever possible.

src/main/scala/strawman/collection/immutable/TreeSet.scala
szeiger: I'll address those issues in the Set PR. Doing it the old way requires a SetBuilder, which in turn requires a `+` operation on Sets that we do not have yet...

szeiger:
I've added the missing map factory support and ported some of the additional tests from #91 (which use Map and SortedMap types). Next up: Collapsing strict building and lazy building again (like in #45), which should remove a lot of the boilerplate.

src/main/scala/strawman/collection/Factories.scala
```diff
@@ -1,20 +1,106 @@
...
   * @tparam C Type of collection (e.g. `List[Int]`, `TreeMap[Int, String]`, etc.)
   */
-trait FromSpecificIterable[-A, +C] extends Any {
+trait FromSpecificIterable[-A, +C] extends Any with BuildFrom[Any, A, C] {
+  def fromSpecificIterable(from: Any)(it: Iterable[A]): C = fromSpecificIterable(it)
   def fromSpecificIterable(it: Iterable[A]): C
 }
 
+/** A more specific `FromSpecificIterable` for strict collection types which can provide a `Builder`. */
+trait FromSpecificIterableWithBuilder[-A, +C] extends Any with FromSpecificIterable[A, C] with StrictBuildFrom[Any, A, C] {
```
julienrf: Do we need this type? It does not seem to be consumed anywhere.

src/test/scala/strawman/collection/test/GenericTest.scala
```diff
@@ -0,0 +1,59 @@
...
+    (parseA.parse(parts(0)), parseB.parse(parts(1))) match {
+      case (Some(a), Some(b)) => Some((a, b))
+      case _ => None
+    }
+  }
+
+  def parseCollection[A, C](bf: BuildFrom[Any, A, C])
+                                    (implicit parseA: Parse[A]): Parse[C] = { (s: String) =>
+    val parts = s.split("\\|")
+    parts.foldLeft[Option[Builder[A, C]]](Some(Builder.from(bf, null))) { (maybeBuilder, s) =>
```
julienrf: The fact that the first parameter of `BuildFrom` is `Any` and the second parameter of `Build.from` is `null` gives me the feeling that `BuildFrom` is not really the appropriate abstraction.

src/main/scala/strawman/collection/Factories.scala
szeiger: Not per so but we do need `FromSpecificIterable with StrictBuildFrom`

src/test/scala/strawman/collection/test/GenericTest.scala
szeiger: `FromSpecificIterable` should be sufficient. AFAIR I ran into some problem with the implicits with that.

PR #102, opened 2017-06-07
Add Range

julienrf:
I also added NumericRange but we should explore (later) a way to unify `Range` and `NumericRange`.


src/main/scala/strawman/collection/immutable/Range.scala
```diff
@@ -0,0 +1,342 @@
...
+  *  @param step        the step for the range.
+  *  @param isInclusive whether the end of the range is included or not
+  */
+@SerialVersionUID(7618862778670199309L)
+final class Range(
+  val start: Int,
+  val end: Int,
+  val step: Int,
+  val isInclusive: Boolean
+)
```
julienrf: The implementation is essentially copy-pasted from the current collections, with a few differences. For instance, there is no more `Inclusive` subtype of `Range`.

pathikrit:
Is there a way to make Range generic e.g. `Range[A](implicit succ: Succesor[A])`? This way we can define `Range[sql.Date]` etc.

Ichoran:
@pathikrit - No, that's definitely `NumericRange` territory except without even a numeric representation.  `Range` is for integers (since you need something efficient to cover integers, and the generic implementations have not been efficient enough).

Ichoran:
Belated, LGTM

PR #108, opened 2017-06-09
Add groupBy

julienrf:
# Summary

- Add a `groupBy` method to `Iterable`
    - Three implementations: a generic implementation in `Buildable` (copied from the current collections), and custom implementations in `View` and `LazyList`
- Add `getOrElse` method to `Map`
- Add `getOrElseUpdate` method to `mutable.Map`
- Add a `newBuilder()` method to appropriate collection factories

# Discussion

The current (I mean, in the current collections) implementation of `groupBy` relies on an available `newBuilder` method.

In our strawman we didn’t have such builders so I wanted to experiment with different implementations.

I compared three implementations:
 - one based on builders (current implementation)
 - one for immutable collections, which depends on an `empty` and a `cons` method to build the groups
 - one for mutable collections, based on `Growable` to build the groups

I compared the implementations with collections (`List`, `ImmutableArray`, `HashSet` and `ArrayBuffer`) of various sizes (from 1 to 7,312,102), containing elements of type `Long`. The benchmark calls the `groupBy` method with the following function `x => x % 5`. (So, the result is a `Map` of 5 groups of equivalent size)

The benchmarks show that for immutable collections with more than 7 elements it’s faster to go with the current implementation.

For mutable collections, the approach based on `Growable`s has the same performance as the current implementation, and it has even slightly better performance on small collections (less than 4 elements).

Therefore, I decided to keep the same implementation as in the current collections (based on builders). We could override the implementation of mutable collections if you think it is worth it…

For reference, here are the different `groupBy` implementations, followed by the code of the benchmarks and the charts.

~~~ scala
// Implementation similar to what we have in the current collections
def groupByBuilder[A, K, C <: Iterable[A]](
  as: C
)(
  newBuilder: () => mutable.Builder[A, C]
)(
  f: A => K
): immutable.Map[K, C] = {
  val m = mutable.Map.empty[K, mutable.Builder[A, C]]
  for (elem <- as) {
    val key = f(elem)
    val bldr = m.getOrElseUpdate(key, newBuilder())
    bldr += elem
  }
  var result = immutable.Map.empty[K, C]
  m.foreach { case (k, v) =>
    result = result + ((k, v.result))
  }
  result
}

// Alternative implementation for immutable collections
// Note that the `cons` function can either prepend or
// append the new value (we don’t guarantee to preserve
// the order of elements within groups)
def groupByImmutable[A, K, C <: Iterable[A]](
  as: C
)(
  empty: C, cons: (A, C) => C
)(
  f: A => K
): immutable.Map[K, C] = {
  var result = immutable.Map.empty[K, C]
  for (elem <- as) {
    val key = f(elem)
    val values = cons(elem, result.getOrElse(key, empty))
    result = result + ((key, values))
  }
  result
}

// Alternative implementation for growable collections
def groupByGrowable[A, K, C <: Iterable[A] with mutable.Growable[A]](
  as: C
)(
  empty: () => C
)(
  f: A => K
): immutable.Map[K, C] = {
  var result = immutable.Map.empty[K, C]
  for (elem <- as) {
    val key = f(elem)
    result.get(key) match {
      case None =>
        val values = empty()
        values += elem
        result = result + ((key, values))
      case Some(values) =>
        values += elem
    }
  }
  result
}
~~~

~~~ scala
object ListGroupByImmutable {
  @Benchmark
  def listGroupBy(bh: Blackhole): Unit = {
    val result = GroupBys.groupByImmutable[Long, Long, List[Long]](xs)(Nil, _ :: _)(_ % 5)
    bh.consume(result)
  }
}

object ListGroupByBuilder {
  @Benchmark
  def listGroupBy(bh: Blackhole): Unit = {
    val result = GroupBys.groupByBuilder[Long, Long, List[Long]](xs)(() => List.newBuilder())(_ % 5)
    bh.consume(result)
  }
}
~~~

listgroupby

immutablearraygroupby

hashsetgroupby

arraybuffergroupby

mutablehashsetgroupby

### DIFF
--- a/src/main/scala/strawman/collection/View.scala
+++ b/src/main/scala/strawman/collection/View.scala
@@ -1,6 +1,7 @@
 package strawman.collection
 
-import strawman.collection.mutable.ArrayBuffer
+import strawman.collection.immutable.ImmutableArray
+import strawman.collection.mutable.{ArrayBuffer, Builder}
 
 import scala.{Any, Boolean, Equals, Int, Nothing, annotation}
 import scala.Predef.intWrapper
@@ -17,15 +18,15 @@ trait View[+A] extends Iterable[A] with IterableOps[A, View, View[A]] {
   override def className = "View"
 
   def groupBy[K](f: (A) => K): immutable.Map[K, View[A]] = {
-    val m = mutable.Map.empty[K, ArrayBuffer[A]]
+    val m = mutable.Map.empty[K, Builder[A, ImmutableArray[A]]]
     for (elem <- coll) {
       val key = f(elem)
-      val bldr = m.getOrElseUpdate(key, ArrayBuffer.empty)
+      val bldr = m.getOrElseUpdate(key, ImmutableArray.newBuilder())
       bldr += elem
     }
     var result = immutable.Map.empty[K, View[A]]
     m.foreach { case (k, v) =>
-      result = result + ((k, v.view))
+      result = result + ((k, v.result().view))
     }
     result
   }
--- a/src/main/scala/strawman/collection/View.scala
+++ b/src/main/scala/strawman/collection/View.scala
@@ -1,5 +1,7 @@
 package strawman.collection
 
+import strawman.collection.mutable.ArrayBuffer
+
 import scala.{Any, Boolean, Equals, Int, Nothing, annotation}
 import scala.Predef.intWrapper
 
@@ -13,6 +15,20 @@ trait View[+A] extends Iterable[A] with IterableOps[A, View, View[A]] {
     fromIterable(coll)
 
   override def className = "View"
+
+  def groupBy[K](f: (A) => K): immutable.Map[K, View[A]] = {
+    val m = mutable.Map.empty[K, ArrayBuffer[A]]
+    for (elem <- coll) {
+      val key = f(elem)
+      val bldr = m.getOrElseUpdate(key, ArrayBuffer.empty)
+      bldr += elem
+    }
+    var result = immutable.Map.empty[K, View[A]]
+    m.foreach { case (k, v) =>
+      result = result + ((k, v.view))
+    }
+    result
+  }
 }
 
 /** This object reifies operations on views as case classes */
--- a/src/main/scala/strawman/collection/immutable/BitSet.scala
+++ b/src/main/scala/strawman/collection/immutable/BitSet.scala
@@ -70,9 +70,7 @@ object BitSet extends SpecificIterableFactory[Int, BitSet] {
   def empty: BitSet = new BitSet1(0L)
 
   def newBuilder(): Builder[Int, BitSet] =
-    new ImmutableBuilder[Int, BitSet](empty) {
-      def add(elem: Int): this.type = { elems = elems + elem; this }
-    }
+    mutable.BitSet.newBuilder().mapResult(bs => fromBitMaskNoCopy(bs.elems))
 
   private def createSmall(a: Long, b: Long): BitSet = if (b == 0L) new BitSet1(a) else new BitSet2(a, b)
 
--- a/src/main/scala/strawman/collection/immutable/BitSet.scala
+++ b/src/main/scala/strawman/collection/immutable/BitSet.scala
@@ -3,7 +3,7 @@ package collection
 package immutable
 
 import BitSetOps.{LogWL, updateArray}
-import mutable.Builder
+import mutable.{Builder, ImmutableBuilder}
 
 import scala.{Array, Boolean, Int, Long, Ordering, SerialVersionUID, Serializable, Unit}
 import scala.Predef.require
@@ -22,6 +22,7 @@ sealed abstract class BitSet
     with collection.BitSet
     with SortedSetOps[Int, SortedSet, BitSet]
     with collection.BitSetOps[BitSet]
+    with Buildable[Int, BitSet]
     with Serializable {
 
   def empty: BitSet = BitSet.empty
@@ -30,7 +31,7 @@ sealed abstract class BitSet
 
   protected[this] def fromSpecificIterable(coll: collection.Iterable[Int]): BitSet = BitSet.fromSpecificIterable(coll)
   protected[this] def sortedFromIterable[B : Ordering](it: collection.Iterable[B]): SortedSet[B] = SortedSet.sortedFromIterable(it)
-
+  protected[this] def newBuilder: Builder[Int, BitSet] = BitSet.newBuilder()
 
   protected[collection] def fromBitMaskNoCopy(elems: Array[Long]): BitSet = BitSet.fromBitMaskNoCopy(elems)
 
@@ -66,6 +67,13 @@ object BitSet extends SpecificIterableFactory[Int, BitSet] {
       case _          => empty.++(it)
     }
 
+  def empty: BitSet = new BitSet1(0L)
+
+  def newBuilder(): Builder[Int, BitSet] =
+    new ImmutableBuilder[Int, BitSet](empty) {
+      def add(elem: Int): this.type = { elems = elems + elem; this }
+    }
+
   private def createSmall(a: Long, b: Long): BitSet = if (b == 0L) new BitSet1(a) else new BitSet2(a, b)
 
   /** A bitset containing all the bits in an array */
@@ -90,8 +98,6 @@ object BitSet extends SpecificIterableFactory[Int, BitSet] {
     else if (len == 2) createSmall(elems(0), elems(1))
     else new BitSetN(elems)
   }
-
-  def empty: BitSet = new BitSet1(0L)
 
   @SerialVersionUID(2260107458435649300L)
   class BitSet1(val elems: Long) extends BitSet {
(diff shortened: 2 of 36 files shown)


src/main/scala/strawman/collection/View.scala
```diff
@@ -13,6 +15,20 @@ trait View[+A] extends Iterable[A] with IterableOps[A, View, View[A]] {
...
+    for (elem <- coll) {
+      val key = f(elem)
+      val bldr = m.getOrElseUpdate(key, ArrayBuffer.empty)
+      bldr += elem
+    }
+    var result = immutable.Map.empty[K, View[A]]
+    m.foreach { case (k, v) =>
+      result = result + ((k, v.view))
+    }
+    result
```
julienrf: This is the implementation of `groupBy` in `View`: each group is represented as an `ArrayBuffer`, and eventually we get `view`s of them.

src/main/scala/strawman/collection/View.scala
```diff
@@ -13,6 +15,20 @@ trait View[+A] extends Iterable[A] with IterableOps[A, View, View[A]] {
     fromIterable(coll)
 
   override def className = "View"
+
+  def groupBy[K](f: (A) => K): immutable.Map[K, View[A]] = {
+    val m = mutable.Map.empty[K, ArrayBuffer[A]]
+    for (elem <- coll) {
+      val key = f(elem)
+      val bldr = m.getOrElseUpdate(key, ArrayBuffer.empty)
```
Ichoran: I think a normal `ArrayBuffer` allocates too much memory.  If there isn't much redundancy you could end up increasing the memory usage by about 5x (!).  I am not sure we have a great alternative data structure right now, but we should think about one.  At the very least I think we should have an array builder and a view of the built array so at *least* we can reclaim all the extra memory after construction is complete.

src/main/scala/strawman/collection/immutable/BitSet.scala
```diff
@@ -66,6 +67,13 @@ object BitSet extends SpecificIterableFactory[Int, BitSet] {
       case _          => empty.++(it)
     }
 
+  def empty: BitSet = new BitSet1(0L)
+
+  def newBuilder(): Builder[Int, BitSet] =
+    new ImmutableBuilder[Int, BitSet](empty) {
```
Ichoran: This is a performance bug waiting to happen.  Larger immutabe bitsets should absolutely not be built this way; they should be built with a mutable bitset, and upon asking for the result the array should be handed over and/or copied.

src/main/scala/strawman/collection/immutable/HashMap.scala
```diff
@@ -28,8 +30,9 @@ import java.lang.{Integer, String, System}
 @SerialVersionUID(2L)
 sealed trait HashMap[K, +V]
   extends Map[K, V]
-     with MapOps[K, V, HashMap, HashMap[K, V]]
-     with Serializable {
+    with MapOps[K, V, HashMap, HashMap[K, V]]
+    with Buildable[(K, V), HashMap[K, V]]
```
Ichoran: Why doesn't `MapOps[K, V, HashMap, HashMap[K, V]]` imply `Buildable[(K, V), HashMap[K, V]]`?

src/main/scala/strawman/collection/immutable/HashSet.scala
```diff
@@ -25,15 +25,18 @@ import java.lang.Integer
 @SerialVersionUID(2L)
 sealed trait HashSet[A]
   extends Set[A]
-     with SetOps[A, HashSet, HashSet[A]]
-     with Serializable {
+    with SetOps[A, HashSet, HashSet[A]]
+    with Buildable[A, HashSet[A]]
```
Ichoran: Same question about `SetOps` implying `Buildable`

src/main/scala/strawman/collection/immutable/ListSet.scala
```diff
@@ -124,5 +127,10 @@ object ListSet extends IterableFactory[ListSet] {
 
   def empty[A]: ListSet[A] = EmptyListSet.asInstanceOf[ListSet[A]]
 
+  def newBuilder[A](): Builder[A, ListSet[A]] =
+    new ImmutableBuilder[A, ListSet[A]](empty) {
+      def add(elem: A): this.type = { elems = elems + elem; this }
```
Ichoran: Any way to push this implementation up into the trait without losing performance?  There's a lot of repetition of this pattern.

src/main/scala/strawman/collection/mutable/ListBuffer.scala
```diff
@@ -12,7 +12,7 @@ import scala.Predef.{assert, intWrapper}
 
 /** Concrete collection type: ListBuffer */
 class ListBuffer[A]
-  extends Seq[A]
+  extends GrowableSeq[A]
```
Ichoran: Why do we need to separately say this is growable, buildable, and a builder?  Also, can't every mutable collection be treated as its own builder?

src/main/scala/strawman/collection/immutable/BitSet.scala
julienrf: Oh, indeed, definitely!

src/main/scala/strawman/collection/immutable/HashMap.scala
julienrf: I think that the decision of extending `Buildable` can only be made at the leaf level (for each concrete collection type). For instance, `Seq` does not imply `Buildable`, which is good because we don’t want `LazyList` to implement `Buildable`.

src/main/scala/strawman/collection/immutable/ListSet.scala
julienrf: I could make `ImmutableBuilder[A, C]` take as parameter a `(C, A) => C` function defining how to add an element to a given collection. Then at use site that would look like the following:

~~~ scala
new ImmutableBuilder[A, ListSet[A]](empty, _ + _)
~~~

WDYT?

src/main/scala/strawman/collection/mutable/ListBuffer.scala
julienrf: > Why do we need to separately say this is growable, buildable, and a builder?

Indeed we could probably assume that `Growable` imply `Buildable`. I see no example where we would have the former without also having the latter.

I decided to put the method that returns the builder in the companion object because I think that could be useful for other purposes (see #97).

> can't every mutable collection be treated as its own builder?

That’s a good question. Actually, @odersky initially designed mutable collections to be their own builders. I thought that it would make the hierarchy slightly more complicated (that’s yet another type that shows up in the linearization) and also I think the public `result` method of builders makes little sense on a concrete collection.

julienrf: > Why do we need to separately say this is growable, buildable

Actually, in the current design `Growable[A]` takes just one type parameter (the type of the elements). If we want to make `Growable[A]` imply `Buildable[A, C]` then we need to add the `C` type parameter to `Growable` and on all its subclasses. A consequence is that this `C` should be supplied by a `XxxOps` trait, which means that we would have a `GrowableIterableOps` branch in the hierarchy that would be refined for `Set`, `Map` and `Seq`. That would be a lot of additional traits :(

Or, could we assume that all mutable collections are `Growable`? (in such a case we could make `mutable.IterableOps` extend `Buildable` and get rid of the `mutable.Iterable` / `mutable.GrowableIterable` distinction)

Ichoran: I forgot that Arrays are not growable.  I guess there is a distinction there to maintain.

src/main/scala/strawman/collection/immutable/ListSet.scala
Ichoran: That's not much better, and we want to be really careful about per-element overhead.  I was thinking about pushing it up into a `SetFactory` which extends `IterableFactory`.  Not sure if there will be performance consequences, though.

src/main/scala/strawman/collection/immutable/HashMap.scala
Ichoran: So it's just because someone might write `MapOps[K, V, Map, Map[K, V]]` that we don't want to be `Buildable`?

julienrf:
@Ichoran can you summarize what needs to be changed?

julienrf:
I rebased the PR and implemented `groupBy` in `Range` and `NumericRange`. I moved the actual implementation from `Buildable` to a `collection.generic.GroupBy` object so that I could reuse it even in collections types that do not extend `Buildable` (such as `Range` and `NumericRange`)

src/main/scala/strawman/collection/immutable/HashMap.scala
szeiger: For `Map` it probably makes sense to extend `Buildable`, just like for `Set` (which I am working on at the moment)

Ichoran:
@julienrf - I'll try to check it this evening.

Ichoran:
Hrm, not sure if there's a way to "unreview".  Dismissing the review doesn't seem to be what I wanted.

julienrf:
Superseded by #111.

PR #109, opened 2017-06-12
Add Vector

### LINKED ISSUE #101: Add Vector

julienrf:
Fixes #101.

### DIFF
--- a/src/main/scala/strawman/collection/Seq.scala
+++ b/src/main/scala/strawman/collection/Seq.scala
@@ -18,6 +18,8 @@ trait LinearSeq[+A] extends Seq[A] with LinearSeqOps[A, LinearSeq, LinearSeq[A]]
 /** Base trait for indexed sequences that have efficient `apply` and `length` */
 trait IndexedSeq[+A] extends Seq[A] with IndexedSeqOps[A, IndexedSeq, IndexedSeq[A]]
 
+object IndexedSeq extends IterableFactory.Delegate[IndexedSeq](immutable.Vector)
+
 /** Base trait for Seq operations */
 trait SeqOps[+A, +CC[X], +C] extends Any
   with IterableOps[A, CC, C]
@@ -50,7 +52,7 @@ trait SeqOps[+A, +CC[X], +C] extends Any
    */
   def reverseIterator: Iterator[A] = reverse.iterator()
 
-  /** Do the elements of this collection are the same (and in the same order)
+  /** Are the elements of this collection the same (and in the same order)
     * as those of `that`?
     */
   def sameElements[B >: A](that: IterableOnce[B]): Boolean =
--- a/src/main/scala/strawman/collection/Seq.scala
+++ b/src/main/scala/strawman/collection/Seq.scala
@@ -24,12 +24,31 @@ trait SeqOps[+A, +CC[X], +C] extends Any
   with ArrayLike[A]
   with Equals {
 
+  /** Evidence that type `C` is (or can be converted to) an `IterableOnce[A]` */
+  implicit protected[this] def toCollection: C => IterableOnce[A]
+
+  /** Returns new $coll with elements in reversed order.
+   *
+   *  $willNotTerminateInf
+   *
+   *  @return A new $coll with all elements of this $coll in reversed order.
+   */
   def reverse: C = {
     var xs: List[A] = Nil
     val it = coll.iterator()
     while (it.hasNext) xs = it.next() :: xs
     fromSpecificIterable(xs)
   }
+
+  /** An iterator yielding elements in reversed order.
+   *
+   *   $willNotTerminateInf
+   *
+   * Note: `xs.reverseIterator` is the same as `xs.reverse.iterator` but might be more efficient.
+   *
+   *  @return  an iterator yielding the elements of this $coll in reversed order
+   */
+  def reverseIterator: Iterator[A] = reverse.iterator()
 
   /** Do the elements of this collection are the same (and in the same order)
     * as those of `that`?
--- a/src/main/scala/strawman/collection/StringOps.scala
+++ b/src/main/scala/strawman/collection/StringOps.scala
@@ -10,11 +10,14 @@ import scala.reflect.ClassTag
 
 class StringOps(val s: String)
   extends AnyVal
-     with SeqOps[Char, Seq, String]
-     with Buildable[Char, String]
-     with ArrayLike[Char] {
+    with IterableOnce[Char]
+    with SeqOps[Char, Seq, String]
+    with Buildable[Char, String]
+    with ArrayLike[Char] {
 
-  protected def coll = new StringView(s)
+  protected[this] def coll = new StringView(s)
+
+  protected[this] def toCollection: String => IterableOnce[Char] = strawman.collection.stringToStringOps
 
   protected[this] def fromSpecificIterable(coll: Iterable[Char]): String = {
     val sb = new StringBuilder
--- a/src/main/scala/strawman/collection/generic/Updated.scala
+++ b/src/main/scala/strawman/collection/generic/Updated.scala
@@ -1,0 +1,39 @@
+package strawman
+package collection.generic
+
+import strawman.collection.mutable.Builder
+
+import scala.{Any, IndexOutOfBoundsException, Int}
+
+object Updated {
+
+  /** Implementation of the `updated` method for strict collections */
+  def strict[A, C](index: Int, elem: A, coll: collection.immutable.Seq[A], newBuilder: () => Builder[A, C]): C = {
+    if (index < 0) throw new IndexOutOfBoundsException(index.toString)
+    val b = newBuilder()
+    var i = 0
+    val it = coll.iterator()
+    while (i < index && it.hasNext) {
+      b += it.next()
+      i += 1
+    }
+    if (!it.hasNext) throw new IndexOutOfBoundsException(index.toString)
+    b += elem
+    it.next()
+    while (it.hasNext) b += it.next()
+    b.result()
+  }
+
+  def toImmutableArray[A](index: Int, elem: A, coll: collection.immutable.IndexedSeq[A]): collection.immutable.ImmutableArray[A] = {
+    val arr = scala.Array.ofDim[Any](coll.size)
+    val it = coll.iterator()
+    var i = 0
+    while (it.hasNext) {
+      arr(i) = it.next()
+      i += 1
+    }
+    arr(index) = elem
+    new collection.immutable.ImmutableArray[A](arr)
+  }
+
+}
--- a/src/main/scala/strawman/collection/immutable/LazyList.scala
+++ b/src/main/scala/strawman/collection/immutable/LazyList.scala
@@ -5,7 +5,6 @@ package immutable
 import scala.{None, Nothing, Option, Some, StringContext, Any, Int}
 import scala.Predef.{???, identity}
 import scala.annotation.tailrec
-import mutable.Builder
 
 class LazyList[+A](expr: => LazyList.Evaluated[A])
   extends Seq[A]
@@ -38,6 +37,14 @@ class LazyList[+A](expr: => LazyList.Evaluated[A])
 
   protected[this] def toCollection: LazyList[A] => IterableOnce[A] = identity
 
+  def updated[B >: A](index: Int, elem: B): LazyList[B] =
+    zipWithIndex.map { case (a, i) => if (i == index) elem else a }
+
+  def zipWithIndex: LazyList[(A, Int)] =
+    LazyList.unfold((0, this)) { case (i, as) =>
+      as.force.map { case (a, _as) => ((a, i), (i + 1, _as)) }
+    }
+
   override def className = "LazyList"
 
   override def toString =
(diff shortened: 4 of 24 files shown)


src/main/scala/strawman/collection/immutable/Vector.scala
```diff
@@ -0,0 +1,1140 @@
...
+ *  @define orderDependentFold
+ *  @define mayNotTerminateInf
+ *  @define willNotTerminateInf
+ */
+@SerialVersionUID(-1334388273712300479L)
+final class Vector[+A] private[immutable] (private[collection] val startIndex: Int, private[collection] val endIndex: Int, focus: Int)
+  extends IndexedSeq[A]
+    with IndexedSeqOps[A, Vector, Vector[A]]
+    with VectorPointer[A @uncheckedVariance]
+    with Serializable { self =>
```
julienrf: I essentially copy-pasted the current implementation and adapted it to the new design when needed.

src/main/scala/strawman/collection/immutable/Vector.scala
```diff
@@ -0,0 +1,1140 @@
...
+//          v
+//        case n if this.size < (n >>> Log2ConcatFaster) && that.isInstanceOf[Vector[_]] =>
+//          var v = that.asInstanceOf[Vector[B]]
+//          val ri = this.reverseIterator
+//          while (ri.hasNext) v = ri.next +: v
+//          v
+//        case _ => super.concat(that)
+//      }
+//    }
+//  }
```
julienrf: These optimizations can not be applied in the new design because `concat` is defined in `IterableOps` as a method taking as parameter a `IterableOnce`. This type is extremely limited because we can consume its elements only once… In these optimizations we do it more than once (e.g. `isEmpty` followed by `size`).

One solution would be to change the type signature of `concat` to take an `Iterable` instead of `IterableOnce`. Do you see a reason **not** to do that?

src/main/scala/strawman/collection/immutable/Vector.scala
```diff
@@ -0,0 +1,1140 @@
...
+   *      y: List[Int] = List(2, 1)
+   *
+   *      scala> println(x)
+   *      List(1)
+   *    }}}
+   *
+   *    @return a new $coll consisting of `value` followed
+   *            by all elements of this $coll.
+   */
+  def appendFront[B >: A](value: B): Vector[B] = {
```
julienrf: I kept the name `appendFront` (and `appendBack`, below) from the current implementation of `Vector`. These methods were private and I made them public because we need an alphabetic alias for `+:` and `:+`. Are you ok with these names or do you prefer `prepend` and `append`?

src/main/scala/strawman/collection/immutable/Vector.scala
```diff
@@ -0,0 +1,1140 @@
...
+      }
+    } else {
+      val elems = new Array[AnyRef](32)
+      elems(0) = value.asInstanceOf[AnyRef]
+      val s = new Vector(0, 1, 0)
+      s.depth = 1
+      s.display0 = elems
+      s
+    }
+  }
```
julienrf: Everything below this line has been literally copy-pasted without any change.

src/main/scala/strawman/collection/immutable/Vector.scala
```diff
@@ -0,0 +1,1143 @@
...
+
+  override def slice(from: Int, until: Int): Vector[A] =
+    take(until).drop(from)
+
+  override def splitAt(n: Int): (Vector[A], Vector[A]) = (take(n), drop(n))
+
+  // concat (suboptimal but avoids worst performance gotchas)
+  override def concat[B >: A](that: IterableOnce[B]): Vector[B] =
+    that match {
+      case it: Iterable[B] =>
```
julienrf: I have to perform this runtime test in order to apply the `concat` optimizations. An alternative solution would be to change the type signature of `concat` (in `Iterable`) to take an `Iterable` as parameter instead of an `IterableOnce`.

src/main/scala/strawman/collection/immutable/Vector.scala
ShaneDelmore: I'd vote for prepend and append.

src/main/scala/strawman/collection/Seq.scala
```diff
@@ -18,20 +18,41 @@ trait LinearSeq[+A] extends Seq[A] with LinearSeqOps[A, LinearSeq, LinearSeq[A]]
...
+object IndexedSeq extends IterableFactory.Delegate[IndexedSeq](immutable.Vector)
+
 /** Base trait for Seq operations */
 trait SeqOps[+A, +CC[X], +C] extends Any
   with IterableOps[A, CC, C]
   with ArrayLike[A]
   with Equals {
 
+  /** Evidence that type `C` is (or can be converted to) an `IterableOnce[A]` */
+  implicit protected[this] def toCollection: C => IterableOnce[A]
```
julienrf: In order to provide a *default implementation* to some methods of `Seq` (e.g. `reverseIterator`), I had to add this implicit conversion from `C` to `IterableOnce[A]`. Note that putting an upper bound `C <: IterableOnce[A]` would not work here because `StringOps` and `ArrayOps` instantiate their `C` type to `java.lang.String` and `scala.Array[A]`, respectively, which would not conform to the upper bound.

src/main/scala/strawman/collection/immutable/Seq.scala
```diff
@@ -2,16 +2,28 @@ package strawman
...
+    *  @param  elem   the replacing element
+    *  @tparam B        the element type of the returned $coll.
+    *  @return a new $coll which is a copy of this $coll with the element at position `index` replaced by `elem`.
+    *  @throws IndexOutOfBoundsException if `index` does not satisfy `0 <= index < length`.
+    *
+    *    @inheritdoc
+    *
+    *    @return a copy of this $coll with the element at position `index` replaced by `elem`.
+    */
+  def updated[B >: A](index: Int, elem: B): CC[B]
```
julienrf: I introduced the `updated` operation in *immutable* `Seq`. In the current collections, this operation takes an implicit `CanBuildFrom` that builds the resulting collection.

Here, I chose to leave it abstract (no default implementation) and implemented it in each concrete collection. For `List` I reused the same implementation as the current one (even though we could improve it by reusing the tail (the elements after `index`) instead of re-creating a new `List`), but for `Range`, `Vector` and `LazyList` I wrote custom implementations.

src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -36,6 +35,16 @@ class LazyList[+A](expr: => LazyList.Evaluated[A])
 
   protected[this] def fromSpecificIterable(coll: collection.Iterable[A]): LazyList[A] = fromIterable(coll)
 
+  protected[this] def toCollection: LazyList[A] => IterableOnce[A] = identity
+
+  def updated[B >: A](index: Int, elem: B): LazyList[B] =
+    zipWithIndex.map { case (a, i) => if (i == index) elem else a }
```
julienrf: The `updated` implementation for `LazyList` is not particularly efficient but preserves laziness. It relies on a (lazy) `zipWithIndex` method itself implemented on top of a (lazy) `unfold` builder.

src/main/scala/strawman/collection/generic/Updated.scala
```diff
@@ -0,0 +1,39 @@
+package strawman
+package collection.generic
+
+import strawman.collection.mutable.Builder
+
+import scala.{Any, IndexOutOfBoundsException, Int}
+
+object Updated {
```
julienrf: I put generic implementations of the `updated` method of `Seq` in this object.

src/main/scala/strawman/collection/immutable/Vector.scala
szeiger: Not only should they have simple names like `prepend` and `append`, they should also be defined higher up in the hierarchy.

src/main/scala/strawman/collection/Seq.scala
szeiger: How is this different from `IterableOps.coll` and why does it have to be implicit?

src/main/scala/strawman/collection/immutable/Vector.scala
szeiger: Looks good to me. Appending an `IterableOnce` is a valid use case.

src/main/scala/strawman/collection/Seq.scala
julienrf: This is because `reverseIterator`’s default implementation relies on `reverse`, which return `C`, not `coll.type`.

There is no particular reason to make it implicit. I could make it explicit as well…

julienrf:
Superseded by #115.

src/main/scala/strawman/collection/StringOps.scala
```diff
@@ -10,11 +10,14 @@ import scala.reflect.ClassTag
 
 class StringOps(val s: String)
   extends AnyVal
-     with SeqOps[Char, Seq, String]
-     with Buildable[Char, String]
-     with ArrayLike[Char] {
+    with IterableOnce[Char]
```
szeiger: Do we need this in StringOps? There's already a conversion to `IterableOnce[Char]` via StringView.

src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -36,6 +35,16 @@ class LazyList[+A](expr: => LazyList.Evaluated[A])
...
+
+  def zipWithIndex: LazyList[(A, Int)] =
```
szeiger: There ought to be a more efficient way to implement this at a lower level but we can leave that for a future performance improvement sprint.

src/main/scala/strawman/collection/immutable/NumericRange.scala
```diff
@@ -80,6 +82,8 @@ final class NumericRange[T](
     else locationAfterN(idx)
   }
 
+  def updated[B >: T](index: Int, elem: B): IndexedSeq[B] = collection.generic.Updated.toImmutableArray(index, elem, coll)
```
szeiger: This could be implemented in a lazy way without materializing the Range. I find it counter-intuitive that it does that. We should either make it lazy or add a warning in the scaladoc.

src/main/scala/strawman/collection/immutable/LazyList.scala
julienrf: We can discuss that later but what’s your idea?

src/main/scala/strawman/collection/immutable/NumericRange.scala
julienrf: Why should it be lazy?

src/main/scala/strawman/collection/StringOps.scala
julienrf: To implement `IterableOnce` we only need to implement the `iterator()` method, which is already present in `StringOps`. Said otherwise, extending `IterableOnce` is “free”.

src/main/scala/strawman/collection/Seq.scala
szeiger: Does it have to be a function? I'd prefer a method, or a different `reverseIterator`, for example via a View.

src/main/scala/strawman/collection/immutable/LazyList.scala
szeiger: `unfold` does some wrapping that could be avoided. We can probably also get rid of the `Option` in the `LazyList` encoding and use null instead.

src/main/scala/strawman/collection/immutable/NumericRange.scala
szeiger: Because it *can* be. Materializing a Range can be an expensive operation. If you want that you can call `to(IndexedSeq)` explicitly.

src/main/scala/strawman/collection/StringOps.scala
szeiger: I don't think `StringOps` should be a "proper" collection type (i.e. one that implements a trait in the basic IterableOnce -> Iterable -> ... hierarchy). Every method you add here is available on all Strings via an implicit conversion in Predef. We already have a second one for WrappedString (2.12) or StringView (strawman) which is an Iterable. By having only one such conversion in Predef you can easily opt out of the implicit conversion from String to a collection but still keep the StringOps methods.

src/main/scala/strawman/collection/immutable/NumericRange.scala
julienrf: The `Range` is not materialized here, only the resulting `IndexedSeq`. I would expect this operation to behave with a strict semantics since `Range` is not a lazy collection type.

src/main/scala/strawman/collection/Seq.scala
julienrf: Yes, it could be advantageously be a view.

src/main/scala/strawman/collection/StringOps.scala
julienrf: You think that we might get ambiguities if we keep the `extends IterableOnce[Char]`?

src/main/scala/strawman/collection/Seq.scala
julienrf: I fixed the implementation in the new PR (#115). Now the default implementation is based on the `reversed` protected method (which is also used by some `Traversable` methods, see #119). The `reversed` method builds a new collection by default. It is overriden in `IndexedSeq` to use an indexed view instead. There is no need for a `toCollection` method anymore.

PR #113, opened 2017-06-15
Drop Scala 2.12.x support

julienrf:
This lets us remove some type annotations in the tests.

In the `memoryBenchmarks` project I now use json4s instead of jawn to produce JSON, because jawn is not yet compatible with 2.13.

### DIFF
--- a/benchmarks/memory/src/main/scala/strawman/collection/MemoryFootprint.scala
+++ b/benchmarks/memory/src/main/scala/strawman/collection/MemoryFootprint.scala
@@ -2,7 +2,7 @@ package bench
 
 import strawman.collection.immutable.{LazyList, List, Range, NumericRange}
 
-import scala.{Any, AnyRef, App, Int, Long, Seq, StringContext}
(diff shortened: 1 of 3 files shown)


szeiger:
I'll update https://github.com/scala/scala/pull/5742 for 2.13 (to be backported to 2.12 later if it doesn't cause any problems in 2.13). Let's wait before we have have 2.13.0-M2 with this fix, otherwise most type errors blow up the compiler instead of giving a proper error message.

Ichoran:
I'm wary of dropping 2.12 support so early given that there may be a number of external libraries that could be useful to us for testing, benchmarking, and so on.  On the other hand, we do need to test the syntax as it will be in 2.13, so maybe we can't hold off any longer.

As far as the changes themselves go, LGTM.

SethTisue:
I have now dropped this repo from the 2.12 community build (as per #320)

julienrf:
The last release that supported 2.12 was 0.9.0. Then the code moved to the scala/scala repository. From now on, the changes will be available in 2.13 only.

PR #114, opened 2017-06-16
Unify strict and lazy building

szeiger:
This is an extension of #97 which provides Builders in BuildFrom and
FromSpecificIterable for all collection types because we need them
for always-strict operations like groupBy and we can always get them
anyway.

This greatly simplifies the design of BuildFrom by getting rid of half
of the factory types.

### DIFF
--- a/src/main/scala/strawman/collection/Factories.scala
+++ b/src/main/scala/strawman/collection/Factories.scala
@@ -20,16 +20,30 @@ trait BuildFrom[-From, -A, +C] extends Any {
   def fromSpecificIterable(from: From)(it: Iterable[A]): C
 }
 
-object BuildFrom {
-  /** Build the source collection type from an IterableOps */
-  implicit def buildFromIterableOps[C[X] <: IterableOps[X, C, _], A, E]: BuildFrom[C[A], E, C[E]] = new BuildFrom[C[A], E, C[E]] {
+object BuildFrom extends BuildFromLowPriority {
+  /** Build the source collection type from a MapOps */
+  implicit def buildFromMapOps[CC[K, V] <: Map[K, V] with MapOps[K, V, CC, _], A, B, E, F]: BuildFrom[CC[A, B], (E, F), CC[E, F]] = new BuildFrom[CC[A, B], (E, F), CC[E, F]] {
     //TODO: Reuse a prototype instance
-    def fromSpecificIterable(from: C[A])(it: Iterable[E]): C[E] = from.iterableFactory.fromIterable(it)
+    def fromSpecificIterable(from: CC[A, B])(it: Iterable[(E, F)]): CC[E, F] = from.mapFactory.fromIterable(it)
+  }
+
+  /** Build the source collection type from a SortedMapOps */
+  implicit def buildFromSortedMapOps[CC[K, V] <: SortedMap[K, V] with SortedMapOps[K, V, CC, _], A, B, E : Ordering, F]: BuildFrom[CC[A, B], (E, F), CC[E, F]] = new BuildFrom[CC[A, B], (E, F), CC[E, F]] {
+    //TODO: Reuse a prototype instance
+    def fromSpecificIterable(from: CC[A, B])(it: Iterable[(E, F)]): CC[E, F] = from.sortedMapFactory.fromSpecificIterable(it)
+  }
+}
+
+trait BuildFromLowPriority {
+  /** Build the source collection type from an IterableOps */
+  implicit def buildFromIterableOps[CC[X] <: IterableOps[X, CC, _], A, E]: BuildFrom[CC[A], E, CC[E]] = new BuildFrom[CC[A], E, CC[E]] {
+    //TODO: Reuse a prototype instance
+    def fromSpecificIterable(from: CC[A])(it: Iterable[E]): CC[E] = from.iterableFactory.fromIterable(it)
   }
 
   /** Build the source collection type from an Iterable with SortedOps */
-  implicit def buildFromSortedOps[C[X] <: Iterable[X] with SortedOps[X, C[X], C], A, E : Ordering]: BuildFrom[C[A], E, C[E]] = new BuildFrom[C[A], E, C[E]] {
-    def fromSpecificIterable(from: C[A])(it: Iterable[E]): C[E] = from.sortedIterableFactory.fromSpecificIterable(it)
+  implicit def buildFromSortedOps[CC[X] <: Iterable[X] with SortedOps[X, CC[X], CC], A, E : Ordering]: BuildFrom[CC[A], E, CC[E]] = new BuildFrom[CC[A], E, CC[E]] {
+    def fromSpecificIterable(from: CC[A])(it: Iterable[E]): CC[E] = from.sortedIterableFactory.fromSpecificIterable(it)
   }
 }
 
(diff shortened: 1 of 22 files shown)


julienrf:
Also, we have to figure out the issue with the dotty compilation in this PR.

szeiger:
@julienrf As discussed previously we want strict methods like `groupBy` on `Iterable`, so providing builders for everything is unavoidable. Building without a source collection is also supported here via `FromSpecificIterable`. This is unchanged from #85. The only difference is that `FromSpecificIterable` also provides the right `Builder`. This is now used in `GenericTest` instead of `BuildFrom`.

src/main/scala/strawman/collection/Factories.scala
```diff
@@ -1,133 +1,165 @@
...
-
   def apply[A](xs: A*): CC[A] = fromIterable(View.Elems(xs: _*))
-
   def fill[A](n: Int)(elem: => A): CC[A] = fromIterable(View.Fill(n)(elem))
-
-}
-
-trait IterableFactoryWithBuilder[+CC[_]] extends IterableFactory[CC] {
-  def newBuilder[A](): Builder[A, CC[A]]
+  def newBuilder[A](): Builder[A, CC[A]] = new ArrayBuffer[A]().mapResult(fromIterable _)
```
szeiger: This default is used for all collection types that cannot provide a better builder. This includes all lazy collections.

julienrf:
> As discussed previously we want strict methods like groupBy on Iterable, so providing builders for everything is unavoidable.

Well, this builder can be `protected` and used by operation implementations only (as in #110 and #111)

src/main/scala/strawman/collection/Factories.scala
```diff
@@ -1,133 +1,165 @@
...
+  }
+}
+
 /**
   * Builds a collection of type `C` from elements of type `A`
   * @tparam A Type of elements (e.g. `Int`, `Boolean`, etc.)
   * @tparam C Type of collection (e.g. `List[Int]`, `TreeMap[Int, String]`, etc.)
   */
-trait FromSpecificIterable[-A, +C] extends Any {
+trait FromSpecificIterable[-A, +C] extends Any with BuildFrom[Any, A, C] {
```
szeiger: Whenever you pass a factory object instead of an implicit `BuildFrom`. All factories provide an implicit conversion to `FromSpecificIterable`.

src/main/scala/strawman/collection/Factories.scala
julienrf: In the case of lazy collections, what about using an `ImmutableArray` instead of an `ArrayBuffer` to avoid the extra space?

src/main/scala/strawman/collection/SortedMap.scala
```diff
@@ -12,7 +12,11 @@ trait SortedMap[K, +V]
...
 trait SortedMapOps[K, +V, +CC[X, Y] <: SortedMap[X, Y] with SortedMapOps[X, Y, CC, _], +C <: SortedMap[K, V]]
   extends MapOps[K, V, Map, C]
-     with SortedOps[K, C] {
+     with SortedOps[K, C, SortedSet] {
+
+  def sortedIterableFactory = SortedSet
+
+  def sortedMapFactory: SortedMapFactory[CC]
 
   protected[this] def sortedMapFromIterable[K2, V2](it: collection.Iterable[(K2, V2)])(implicit ordering: Ordering[K2]): CC[K2, V2]
```
szeiger: Yes, I think we can delegate this methods now the same way it is already done for `fromSpecificIterable`.

src/main/scala/strawman/collection/Factories.scala
szeiger: You don't know the size ahead of time so you can either copy everything or again or live with some wasted space.

odersky:
I'd like to review this but can only do it after PLDI. So it will be probably a week until you have the review. Thanks for holding that long...

src/main/scala/strawman/collection/Factories.scala
```diff
@@ -1,133 +1,165 @@
...
 
+
+/** Builds a collection of type `C` from elements of type `A` when a source collection of type `From` is available.
+  * Implicit instances of `BuildFrom` are available for all collection types.
+  *
+  * @tparam From Type of source collection
+  * @tparam A Type of elements (e.g. `Int`, `Boolean`, etc.)
+  * @tparam C Type of collection (e.g. `List[Int]`, `TreeMap[Int, String]`, etc.)
+  */
+trait BuildFrom[-From, -A, +C] extends Any {
```
Ichoran: What does making the `C` a type parameter instead of a type member do for us?

src/main/scala/strawman/collection/Factories.scala
```diff
@@ -1,133 +1,165 @@
...
+object BuildFrom extends BuildFromLowPriority {
+  /** Build the source collection type from a MapOps */
+  implicit def buildFromMapOps[CC[K, V] <: Map[K, V] with MapOps[K, V, CC, _], A, B, E, F]: BuildFrom[CC[A, B], (E, F), CC[E, F]] = new BuildFrom[CC[A, B], (E, F), CC[E, F]] {
+    //TODO: Reuse a prototype instance
+    def newBuilder(from: CC[A, B]): Builder[(E, F), CC[E, F]] = from.mapFactory.newBuilder[E, F]()
+    def fromSpecificIterable(from: CC[A, B])(it: Iterable[(E, F)]): CC[E, F] = from.mapFactory.fromIterable(it)
+  }
+
+  /** Build the source collection type from a SortedMapOps */
+  implicit def buildFromSortedMapOps[CC[K, V] <: SortedMap[K, V] with SortedMapOps[K, V, CC, _], A, B, E : Ordering, F]: BuildFrom[CC[A, B], (E, F), CC[E, F]] = new BuildFrom[CC[A, B], (E, F), CC[E, F]] {
```
Ichoran: Most people won't have to look at this, but it's still pretty imposing with seven (?!) separate type parameters (five "real", two parameterizing `CC`).

Ichoran:
I'm also quite comfortable with this approach.  It feels a little more flexible but a little more imposing than Julien's #112.  I can't easily tell what one could accomplish here that one couldn't there, but overall the type member strategy, despite requiring a little more magic, seems overall a bit friendlier.

Both are pretty good, though, for what the tests cover.  Do either of you have a clear idea of what the most important differences are for user-level code?  I wasn't able to easily distinguish them.

Also, I think there's a principle underlying builders that should simplify matters: the only reason you care about the `From` collection type is when you want it to be the same as the `To` type (or as close as possible).  Otherwise, you just want to get the elements and you don't really care what `From` is (beyond any constraints you put on the type of the collection itself, but the method etc parameters will see to that).  So you might care about `From, A` or `A, To`, but not about `From, A, To`.

This suggests to me that we probably don't need an explicit `From, A, To` set of types like in CBF, but can do with type members as in #112.  But that's a hunch based on the logical structure; are there counterexamples?

src/main/scala/strawman/collection/Factories.scala
julienrf: I mean that even *after* an `ArrayBuffer` has been completely built there might still be some unused extra space. If you build an `ImmutableArray` (or just call `toArray` on the `ArrayBuffer`, which is almost exactly the same thing), then you trim the end of the buffer.

src/main/scala/strawman/collection/Factories.scala
szeiger: I would prefer a type member (and I did use one in my original implementation in #45) but `FromSpecificIterable` in Martin's simplified version uses a type parameter so I used the same here. I think we should use type members for both. Unlike `CanBuildFrom` lookup for `BuildFrom` and `FromSpecificIterable` is never determined by this type. It is strictly an "out" parameter.

src/main/scala/strawman/collection/Factories.scala
szeiger: Is there any precedent for this in other collection methods? AFAIK trimming to the current size is always an explicit operation. The default is to avoid copying at the expense of some unused space.

julienrf:
> Do either of you have a clear idea of what the most important differences are for user-level code?

Both solutions seem to be equivalent. The only difference I noted is that in #112, we can get an implicit `Parse[List[…]]` instance (see [here](https://github.com/scala/collection-strawman/pull/112/files#diff-c5f2797a856a5eac5ee347e003321a00R55)), whereas in this PR we have to explicitly provide the factory (see [here](https://github.com/scala/collection-strawman/pull/114/files#diff-c5f2797a856a5eac5ee347e003321a00R53)). But I think it might just be a choice from @szeiger and that it should be possible to get implicit instances (wdyt, @szeiger?).

Also, in #112 I keep differentiating factories that have builders from those that don’t. I’d rather not expose a `LazyList.newBuilder()` method nor a `View.newBuilder()` method.

szeiger:
It should be possible to add an implicit method to each factory type that provides a `BuildFrom[Any, ...]`. This would give you type-driven lookups. This may require also moving the implicits from the `BuildFrom` companion object to get the priorities right when the source type is known.

szeiger:
One important difference is that in #112 AFAICT all Builders are constructed by expected target type, not driven by the source collection, whereas here the default implicits fall back to the source collection. This allows you to do transformations independent of the static type: The static type of the source determines the level of detail of see of the static type of the target but the implementation class is always the same. This fixes one of the consistency problem of the current collections' `CanBuildFrom` (where the type-driven lookup is similar to #112).

szeiger:
Also note that the source-driven transformation is consistent with the new way of performing operations without `CanBuildFrom`. In the current collections it works in some cases but not in others because there are no advanced builder types for maps and sorted collections.

julienrf:
> in #112 AFAICT all Builders are constructed by expected target type, not driven by the source collection

They are if you ask for an implicit `BuildFrom`.

But if you ask for an implicit `BuildTo` then that’s the opposite: you directly ask for a builder for your target collection.

> The static type of the source determines the level of detail of see of the static type of the target but the implementation class is always the same.

Indeed this feature is not supported by my PR: if the static type of the source is `Seq` then you’ll get a default `Seq` implementation as a result of the transformation, whatever the runtime type of the source collection was.

szeiger:
I've been playing around with type-driven building. I run into the problem that I already encountered previously that `BuildFrom` for sorted maps and sorted sets are ambiguous. This is the reason why this PR has a `BuildFromLowPriority` but this solution doesn't work for implicits in the factories. I believe the solution is a `SortedIterableOps` in parallel to `SortedMapOps` so we get a clear split between unary and binary sorted collections.

This still leaves the problem of distinguishing between the static and dynamic BuildFrom instances. With `From` being contravariant, a `BuildFrom[Any, ...]` is preferred but we want it the other way around.

Ichoran:
With regard to the actual type, not the statically visible type, being the collection: I think that's important behavior, probably important enough to favor this approach.  But if it can be made to work with the simplified type parameters (pushing `To` to a type member, and using `Aux` under a friendly name if needed), that would be even better.

Why do the implicits in the factories not obey priority?  Or is it just that you need to create a low priority thing for them each time which is awkward?

szeiger:
For sorted maps you get a `BuildFrom[..., ..., TreeSet[...]]` and a `BuildFrom[..., ..., TreeMap[...]]`, neither of which is more specific than the other. I'll squash and rebase this PR and I will also try to fix this issue by restricting `buildFromSortedOps` to a new `SortedIterableOps` type so it doesn't apply to `SortedMapOps` anymore.

szeiger:
OK, restricting to `SortedSetOps` instead of `SortedOps` doesn't solve the problem. There are still ambiguities between either one of `buildFromMapOps` / `buildFromSortedMapOps` and `buildFromIterableOps`. This used to work in #45. It is not clear to me why it doesn't work anymore with the simplified design. Both Scala and Dotty give the same results and switching back to a `type To` type member doesn't make a difference, either. I'll keep the `BuildFromLowPriority` for now.

PR #117, opened 2017-06-20
Add collect

julienrf:


src/main/scala/strawman/collection/SortedMap.scala
```diff
@@ -26,6 +26,12 @@ trait SortedMapOps[K, +V, +CC[X, Y] <: SortedMap[X, Y] with SortedMapOps[X, Y, C
   def flatMap[K2, V2](f: ((K, V)) => IterableOnce[(K2, V2)])(implicit ordering: Ordering[K2]): CC[K2, V2] =
     sortedMapFromIterable(View.FlatMap(coll, f))
 
+  def collect[K2, V2](pf: PartialFunction[(K, V), (K2, V2)])(implicit ordering: Ordering[K2]): CC[K2, V2] =
+    flatMap { (kv: (K, V)) =>
+      if (pf.isDefinedAt(kv)) View.Single(pf(kv))
+      else View.Empty
+    }
```
julienrf: Note that the type annotation `kv: (K, V)` is required with both dotty and scalac 2.13.

src/test/scala/strawman/collection/test/Test.scala
```diff
@@ -469,6 +469,8 @@ class StrawmanTest {
     val xs9t: mutable.HashMap[Int, String] = xs9
     val xs10 = xs1 ++ xs2
     val xs11: immutable.SortedMap[String, Int] = xs10
+    val xs12 = xs11.collect({ case (k, v) if k.length == v => (v, k) }: scala.PartialFunction[(String, Int), (Int, String)])
```
julienrf: The type annotation (`scala.PartialFunction[…]`) is required with dotty, Scala 2.12 and Scala 2.13.

julienrf:
Note that the fact that type annotations are required (see in tests) is an important problem. We will have to fix type inference in both Scala and Dotty.

julienrf:
During our last meeting we emit the idea that we don’t need `collect` on `Sorted` collection types (such as `SortedSet` and `SortedMap`). Actually, I disagree, I think that it is important to be able to write things like:

~~~ scala
val xs: TreeSet[Int] = TreeSet(1, 2, 3).collect { case x if x % 2 == 0 => x }
~~~

(And same with `SortedMap`).

Thus, we have to fix type inference in Scala and Dotty.

odersky:
I just checked that we have that functionality in existing Scala. So, yes, we should support it if possible. But it's not a show-stopper if we can't.

src/main/scala/strawman/collection/SortedMap.scala
szeiger: This works for me on 2.12 and 2.13. A failure in overload resolution could be due to https://github.com/scala/bug/issues/10194 but the fix for it is present in our currently used PR snapshot and in 2.13.0-M1. The original implementation in https://github.com/scala/scala/pull/5307 was based on Dotty. Maybe it has a similar bug?

src/test/scala/strawman/collection/test/Test.scala
szeiger: 2.12 and Dotty complain about missing types. 2.13 looks more sinister:

```
[error] /Users/szeiger/code/collection-strawman/src/test/scala/strawman/collection/test/Test.scala:504: constructor cannot be instantiated to expected type;
[error]  found   : (T1, T2)
[error]  required: Unit
[error]     val xs12 = xs11.collect({ case (k, v) if k.length == v => (v, k) } /*: scala.PartialFunction[(String, Int), (Int, String)] */)
[error]                                    ^
[error] one error found
```

szeiger: Simplified:

```scala
trait SortedMap[K, V] {
  def collect[B](pf: PartialFunction[(K, V), B]): Int = 0
  def collect[K2 : Ordering, V2](pf: PartialFunction[(K, V), (K2, V2)]): Int = 0

  def foo(xs: SortedMap[String, scala.Int]) = xs.collect { case (k, v) => 1 }
}
```
It works with `Function1` instead of `PartialFunction`. Could be related to 9cd3becde5bb7c37af64ad5ac1bd15aa4fafd1b4.

szeiger: Yes, https://github.com/scala/scala/pull/5698/commits/9cd3becde5bb7c37af64ad5ac1bd15aa4fafd1b4 is indeed the culprit. It changes the expected type from `?` to `Unit`.

szeiger: Just to be clear, this cannot work in 2.12 because support for overloaded methods with `PartialFunction` formal arguments was only introduced in 2.13 via https://github.com/scala/scala/commit/9cd3becde5bb7c37af64ad5ac1bd15aa4fafd1b4. I'm investigating the error there (it should behave the same way as it does for `Function1`). It should work in Dotty where the same changes were made in https://github.com/lampepfl/dotty/pull/2015.

szeiger: Fix for 2.13: https://github.com/scala/scala/pull/5975

szeiger: Dotty ticket: https://github.com/lampepfl/dotty/issues/2866

src/main/scala/strawman/collection/SortedMap.scala
szeiger: Dotty ticket: https://github.com/lampepfl/dotty/issues/2867

PR #126, opened 2017-06-22
Add most of the current methods of Traversable to IterableOps

### LINKED ISSUE #119: Add missing operations of Traversable

julienrf:
Essentially, the approach is to implement them directly in `IterableOps` when they are “terminal” methods (e.g. `max`, `reduce`) and to have a view based implementation when a collection is returned (e.g. `dropWhile`, `scanLeft`). When we return a collection but have no other choice than consuming all the elements (e.g. `scanRight`), a custom implementation is done in `IterableOps`.

Fixes #119.

### DIFF
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -3,7 +3,8 @@ package collection
 
 import scala.annotation.unchecked.uncheckedVariance
 import scala.reflect.ClassTag
-import scala.{Any, Array, Boolean, `inline`, Int, Numeric, Ordering, StringContext, Unit}
+import scala.{Any, Array, Boolean, `inline`, Int, None, Numeric, Option, Ordering, StringContext, Some, Unit}
+import scala.Predef.<:<
 import java.lang.{String, UnsupportedOperationException}
 
 import strawman.collection.mutable.{ArrayBuffer, Builder, StringBuilder}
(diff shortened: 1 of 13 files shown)


src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -47,19 +47,160 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
...
+   *           where `x,,1,,, ..., x,,n,,` are the elements of this $coll.
+   *  @throws UnsupportedOperationException if this $coll is empty.   */
+  def reduceLeft[B >: A](op: (B, A) => B): B = {
+    if (isEmpty)
+      throw new UnsupportedOperationException("empty.reduceLeft")
+
+    var first = true
+    var acc: B = 0.asInstanceOf[B]
+
+    for (x <- coll) {
```
Ichoran: Do we know that this ever does substantially better than the iterator-based version?

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -47,19 +47,160 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
...
+   *             op(x_1, op(x_2, ..., op(x_{n-1}, x_n)...))
+   *           }}}
+   *           where `x,,1,,, ..., x,,n,,` are the elements of this $coll.
+   *  @throws UnsupportedOperationException if this $coll is empty.
+   */
+  def reduceRight[B >: A](op: (A, B) => B): B = {
+    if (isEmpty)
+      throw new UnsupportedOperationException("empty.reduceRight")
+
+    reversed.reduceLeft[B]((x, y) => op(y, x))
```
Ichoran: I'm not at all sure this is a good idea for performance, as it requires you to rebuild a new collection of the same type which might be a lot slower than explicitly building the fastest collection one could.  (As a placeholder to get it working, it's fine.)

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -279,7 +553,63 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
...
+    *    List(1, 2, 3, 4).scanRight(0)(_ + _) == List(10, 9, 7, 4, 0)
+    *  }}}
+    *
+    *  @tparam B      the type of the elements in the resulting collection
+    *  @param z       the initial value
+    *  @param op      the binary operator applied to the intermediate result and the element
+    *  @return        collection with intermediate results
+    */
+  def scanRight[B](z: B)(op: (A, B) => B): CC[B] = {
+    var scanned = z :: immutable.Nil
```
Ichoran: I would not have thought List would be the highest performance way to do this, except maybe with collections of size <= 2.  Although many collections are small, I'd expect `scan` variants to be used on longer collections most of the time.

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -302,4 +632,29 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
...
+    *  @tparam A1    the type of the first half of the element pairs
+    *  @tparam A2    the type of the second half of the element pairs
+    *  @param asPair an implicit conversion which asserts that the element type
+    *                of this $coll is a pair.
+    *  @return       a pair of ${coll}s, containing the first, respectively second
+    *                half of each element pair of this $coll.
+    */
+  def unzip[A1, A2](implicit asPair: A <:< (A1, A2)): (CC[A1], CC[A2]) = {
+    val unzipped = View.Unzip(coll)
+    (fromIterable(unzipped.left), fromIterable(unzipped.right))
```
Ichoran: Not quite relevant here but I didn't catch it before: `left` and `right` are unfortunate names given their meaning in `Either`.

julienrf: I can call them `first` and `second` if you prefer?

src/main/scala/strawman/collection/Iterable.scala
julienrf: I took the implementation of the current collections. We can improve it later.

src/main/scala/strawman/collection/Iterable.scala
julienrf: Which collection type should we use instead? I think `List` has the fastest `cons` implementation (modulo the effects due to garbage collection…).

src/main/scala/strawman/collection/Iterable.scala
julienrf: `reversed` builds a `List`.

src/main/scala/strawman/collection/Iterable.scala
Ichoran: I guess we should leave it for now.  I'd try a manual unrolling strategy, but as long as we don't worry about GC list does have a fast cons.

Issue #127, opened 2017-06-24
LazyList.#:: not lazy due to parser rewriting of right-associative infix operators


pnf:
Evaluating
```.scala
  def bleh(i: Int) = {println(s"bleh $i"); i}
  val xs20 = bleh(1) #:: bleh(2) #:: bleh(3) #:: LazyList.empty
```
will immediately execute all the `println`s, even though `.force` is never called and `.evaluated` remains `false`.   If one defines a left associative `prepend` with the same definition as `#::`, then
```.scala
    val x21 = LazyList.empty.prepend(bleh(3)).prepend(bleh(2)).prepend(bleh(1))
```
will behave in a properly lazy manner.

This is no great mystery, as the parser converts the definition of `xs20` to:
```.scala
 val xs20 = {   <synthetic> <artifact> val x$68 = bleh(1);
    {  <synthetic> <artifact> val x$67 = bleh(2);
  {  <synthetic> <artifact> val x$66 = bleh(3);
  LazyList.empty.$hash$colon$colon(x$66)}.$hash$colon$colon(x$67)}.$hash$colon$colon(x$68)  };
```

This seems to be scala/bug#7333, which was closed as a duplicate of the somewhat less general scala/bug#1980

SethTisue:
if you're aware of #7333 and #1980, then why did you open this ticket?

SethTisue:
what do you have in mind in opening this ticket, given the existence of scala/bug#1980?

pnf:
Thank you for the correction.
Given that scala/bug#1980 has been around since 2009 and has now been relegated to the backlog., perhaps removing the `#::` operator from `LazyList` would be a good idea?  I was initially trying to add an assertion to verify laziness in `lazyListTestOps`, but that didn't work out so well.

smarter:
FWIW, this should be much easier to fix in dotty where the problematic desugaring is not done during parser but during typechecking, where we should have enough information to make the intermediate vals `lazy`. I opened https://github.com/lampepfl/dotty/issues/2808 to track this.

pnf:
Cool. I was concocting a scheme to  restore the laziness post-typer in a plugin, but it would clearly be cleaner to have the desugaring entirely post typer.

In the mean time, having a LazyList with an eager cons seems at the very least misleading, so maybe this issue has a collection-specific value independent of the existing scalac issues.

szeiger:
[...]

PR #128, opened 2017-06-25
Laziness tweaks

pnf:
Make `LazyList.#::`'s argument non-lazy, as it will actually be evaluated eagerly; add a truly lazy `prepend` method.
Somewhat related: stub out `View.toString` so views don't get forced prematurely.

### DIFF
--- a/src/main/scala/strawman/collection/View.scala
+++ b/src/main/scala/strawman/collection/View.scala
@@ -16,6 +16,8 @@ trait View[+A] extends Iterable[A] with IterableOps[A, View, View[A]] {
   protected[this] def newSpecificBuilder(): Builder[A, View[A]] =
     immutable.IndexedSeq.newBuilder().mapResult(_.view)
 
+  override def toString = "View(?)"
+
   override def className = "View"
 }
 
--- a/src/main/scala/strawman/collection/immutable/LazyList.scala
+++ b/src/main/scala/strawman/collection/immutable/LazyList.scala
@@ -30,7 +30,6 @@ class LazyList[+A](expr: => LazyList.Evaluated[A])
 
   @tailrec final def length: Int = if (isEmpty) 0 else tail.length
 
-  def #:: [B >: A](elem: B): LazyList[B] = new LazyList(Some((elem, this)))
   def prepend[B >: A](elem: => B): LazyList[B] = new LazyList(Some((elem, this)))
 
   def iterableFactory = LazyList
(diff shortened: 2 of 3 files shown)


julienrf:
@SethTisue Do you think we could fix the issue in scalac 2.13?

julienrf:
If the issue can not be resolved in the compiler, then we should use the same workaround as we use in the current `Stream` collection, as noted by @szeiger in #127.

pnf:
I changed it to use a `Stream`-like implicit trick and added the classic Fibonacci example (which now works) to the tests as well.  It's a bit irritating that the first element is still strict, but this applies to `Stream` as well.

pnf:
So, the bad news is that the new tests don't pass with dotty.  The even worse news is that they don't pass under dotty even with `scala.collection.immutable.Stream`...

julienrf:
What if you explicitly instantiate the correct type of elements? `List[BigInt](0, 1, 1, 2)`

julienrf:
Hey @pnf, we discussed this issue and came to the conclusion that we should not have lazy right-associative operators unless they are properly supported by the language.

Consequently, we should remove the `#::` and just have a `prepend` method on `LazyList`.

julienrf:
@pnf I think we should completely remove the `#::` operator (and rename the homonym extractor). The workaround that relies on an wrapping class is only partially lazy.

pnf:
@julienrf I'm not so sure... The `#::` operator is "sufficiently lazy" as to be useful for `Stream`, where it gets extensive use in real code, and it could be made genuinely lazy with a little compiler work. 
That said, I'm unclear on whether `LazyList` is intended to replace `Stream` or to supplement it, and, if the latter, why one would choose one over the other.

julienrf:
`LazyList` replaces `Stream`.

I really dislike having an `#::` operator that’s only partially lazy. I think that could lead to very confusing situations. I prefer the semantics of operations on `LazyList` to be clearly always lazy excepted for “terminal” operations (like `fold`).

pnf:
It seems like there are a few probable routes by which it would become fully lazy by the time the new collections are stable: eg.the underlying SI might be fixed, or we may be able to use scalameta. If LL is to replace Stream and we remove the right-associative syntax, the message to users is that we're removing functionality they use, for their own good, because of a bug in another part of the ecosystem that we don't feel like addressing (for some meaning of "we").

julienrf:
Alright, that’s good for now, could you please squash your commits into a single one?

julienrf:
Thanks @pnf, I squashed your commits and merged them to master: https://github.com/scala/collection-strawman/commit/950e7ddb432e0d1e1ebff34d292ff457908e07ba

PR #130, opened 2017-06-27
Add most common operations of Iterable

### LINKED ISSUE #120: Add missing operations of Iterable

julienrf:
Fixes #120.

### DIFF
--- a/src/main/scala/strawman/collection/Iterator.scala
+++ b/src/main/scala/strawman/collection/Iterator.scala
@@ -1,7 +1,8 @@
 package strawman.collection
 
 import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
-import scala.{Any, Array, Boolean, Int, None, NoSuchElementException, Nothing, Option, Some, Unit}
+import scala.{Any, Array, Boolean, Int, None, NoSuchElementException, Nothing, Option, StringContext, Some, Unit}
+import scala.Predef.{intWrapper, require}
 import strawman.collection.mutable.ArrayBuffer
 
 /** A core Iterator class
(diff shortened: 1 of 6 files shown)


src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -71,6 +72,186 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+      }
+      lazy val len = xs.length
+      lazy val incomplete = len < count
+
+      // if 0 elements are requested, or if the number of newly obtained
+      // elements is less than the gap between sequences, we are done.
+      def deliver(howMany: Int) = {
+        (howMany > 0 && (isFirst || len > gap)) && {
+          if (!isFirst)
+            buffer dropInPlace (step min prevSize)
```
julienrf: This code has been copied from the current collections, excepted that I replaced `trimStart` with `dropInPlace`.

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -71,6 +72,186 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+      // The order of terms in the following condition is important
+      // here as self.hasNext could be blocking
+      while (i < size && self.hasNext) {
+        buf += self.next
+        i += 1
+      }
+      buf
+    }
+
+    private def padding(x: Int) = immutable.ImmutableArray.fill(x)(pad.get())
```
julienrf: Here a `List` was used but I thought that an `ImmutableArray` would be more efficient since the size is known.

PR #135, opened 2017-06-28
Add most of missing Seq operations

julienrf:


src/main/scala/strawman/collection/Seq.scala
```diff
@@ -26,6 +31,9 @@ trait SeqOps[+A, +CC[X], +C] extends Any
   with ArrayLike[A]
   with Equals {
 
+  protected[this] def c: C
+  protected[this] def seq: Seq[A]
```
julienrf: This is unfortunate. For reasons explained in #134, I had to introduce these two methods. I need more time to think of a better solution.

Ichoran: I didn't check all the `c` uses, but the `seq` uses don't obviously need to be a `seq` as far as I can tell.  None of them seemed to rely upon fast random access, and anyway, `List` doesn't have that.

Anyway, if we really do need this functionality, we should make it public as `toSeq`.  There is no case I can think of where we need it to implement methods, but we can be sure that nobody else needs it.

julienrf:
> I didn't check all the `c` uses, but the `seq` uses don't obviously need to be a seq as far as I can tell

They are both needed to implement `permutations`. We do use `sortBy` which is only available on `Seq` collections.

Note that we can replace `c` with `fromSpecificIterable(coll)` and `seq` with `fromIterable(coll)` (if we put an upper bound on `CC`) but that would be ridiculously inefficient for `ArrayOps` and `StringOps`.

> Anyway, if we really do need this functionality, we should make it public as toSeq. There is no case I can think of where we need it to implement methods, but we can be sure that nobody else needs it.

I don’t think it is useful to have a `toSeq` method on `Seq`. The problem here is a self-type problem for our internal implementation. I’m not sure users are affected by that.

src/main/scala/strawman/collection/Seq.scala
julienrf: > I didn't check all the `c` uses, but the `seq` uses don't obviously need to be a seq as far as I can tell

They are both needed to implement `permutations`. We do use `sortBy` which is only available on `Seq` collections.

Note that we can replace `c` with `fromSpecificIterable(coll)` and `seq` with `fromIterable(coll)` (if we put an upper bound on `CC`) but that would be ridiculously inefficient for `ArrayOps` and `StringOps`.

> Anyway, if we really do need this functionality, we should make it public as toSeq. There is no case I can think of where we need it to implement methods, but we can be sure that nobody else needs it.

I don’t think it is useful to have a `toSeq` method on `Seq`. The problem here is a self-type problem for our internal implementation. I’m not sure users are affected by that.

Ichoran: How do you write generic code that does seq-like things but works on `Array` and `String` too?

If users can do it that way (whatever that way is), we should just do it that way too and not try for this inheritance scheme.  Just inherit the parts that always work and reimplement the rest (possibly by deferring to a wrapping collection instead of an ops collection?).

julienrf: I’m going to postpone these issues so that the 0.2.0 gets the contents of this PR. #134 keeps track of them.

PR #136, opened 2017-06-29
Fixes for #114

julienrf:
@szeiger is that OK for you to merge this?


julienrf:
I rebased #114 on master and squashed all the commits into a single one. Then I fixed the compilation errors.

szeiger:
So did I because that's what we discussed in this week's meeting. Except that my rebase effort wasn't quite finished yet because #110 split up the factories into strict and lazy ones despite our preference for having strict building everywhere in the previous meeting. We need to be more efficient with how we spent our time.

PR #138, opened 2017-06-29
Add Set operations.

julienrf:


src/main/scala/strawman/collection/Set.scala
```diff
@@ -26,8 +27,81 @@ trait SetOps[A, +CC[X], +C <: Set[A]]
...
+    */
   def subsetOf(that: Set[A]): Boolean = this.forall(that)
 
+  /** An iterator over all subsets of this set of the given size.
+    *  If the requested size is impossible, an empty iterator is returned.
+    *
+    *  @param len  the size of the subsets.
+    *  @return     the iterator.
+    */
+  def subsets(len: Int): Iterator[C] = {
```
julienrf: These `subsets` operations return an `Iterator` in the current collections, so I did that also here. But what do you think of returning a `View` instead?

src/main/scala/strawman/collection/mutable/Set.scala
```diff
@@ -48,6 +48,8 @@ trait SetOps[A, +CC[X], +C <: Set[A]]
     res
   }
 
+  def diff(that: collection.Set[A]): C = that.foldLeft(clone())(_ -= _)
```
julienrf: Maybe this would be better?

~~~ scala
coll.foldLeft(empty)((result, elem) => if (that.contains(elem)) result else result += elem)
~~~

src/main/scala/strawman/collection/immutable/Set.scala
```diff
@@ -43,7 +43,8 @@ trait SetOps[A, +CC[X], +C <: Set[A] with SetOps[A, Set, C]]
     result
   }
 
-  def diff(that: collection.Set[A]): C = fromSpecificIterable(that.foldLeft(coll)(_ - _))
+  def diff(that: collection.Set[A]): C =
+    coll.foldLeft(empty)((result, elem) => if (that contains elem) result else result + elem)
```
Ichoran: Yes, this is better.

PR #144, opened 2017-07-03
View transformers

pnf:
POC for composing view transformations, ala transducers.

### DIFF
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -244,7 +244,7 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
   /** A view representing the elements of this collection. */
   def view: View[A] = View.fromIterator(coll.iterator())
 
-  def transform[B](vt: Iterable[A] => View[B]): CC[B] =  fromIterable(vt(coll))
+  def transform[B](vt: ViewTransformer[A,B]): CC[B] =  fromIterable(vt(coll))
 
   /** Given a collection factory `fi`, convert this collection to the appropriate
     * representation for the current element type `A`. Example uses:
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -244,6 +244,8 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
   /** A view representing the elements of this collection. */
   def view: View[A] = View.fromIterator(coll.iterator())
 
+  def transform[B](vt: Iterable[A] => View[B]): CC[B] =  fromIterable(vt(coll))
+
   /** Given a collection factory `fi`, convert this collection to the appropriate
     * representation for the current element type `A`. Example uses:
     *
--- a/src/main/scala/strawman/collection/View.scala
+++ b/src/main/scala/strawman/collection/View.scala
@@ -157,6 +157,11 @@ object View extends IterableFactory[View] {
     override def knownSize = underlying.knownSize
   }
 
+  case class Collect[A, B](underlying: Iterable[A], pf: scala.PartialFunction[A,B]) extends View[B] {
+    def iterator() = underlying.iterator().filter(pf.isDefinedAt(_)).map(pf.apply(_))
+    override def knownSize = underlying.knownSize
+  }
+
   /** A view that flatmaps elements of the underlying collection. */
   case class FlatMap[A, B](underlying: Iterable[A], f: A => IterableOnce[B]) extends View[B] {
     def iterator() = underlying.iterator().flatMap(f)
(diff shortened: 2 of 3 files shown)


src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -244,6 +244,8 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
   /** A view representing the elements of this collection. */
   def view: View[A] = View.fromIterator(coll.iterator())
 
+  def transform[B](vt: Iterable[A] => View[B]): CC[B] =  fromIterable(vt(coll))
```
julienrf: I’m not a fan of the `transform` name. Maybe `transformView`?

src/main/scala/strawman/collection/View.scala
```diff
@@ -250,6 +255,19 @@ object View extends IterableFactory[View] {
...
+object ViewTransformer {
+  type VT[-A,+B] = Iterable[A] => View[B]
+  implicit class ViewTransFormerOps[A,B](val vt: VT[A,B]) {
+    def map[C](f: B => C): VT[A,C] = v => new View.Map(vt(v),f)
+    def flatMap[C](f: B => IterableOnce[C]): VT[A,C] = v => new View.FlatMap(vt(v),f)
+    def filter(f: B => Boolean): VT[A,B] = v => new View.Filter(vt(v),f)
+    def drop(n: Int): VT[A,B] = v => new View.Drop(vt(v),n)
+    def collect[C](pf: scala.PartialFunction[B,C]): VT[A,C] = v => new View.Collect(vt(v),pf)
+    def zipwithIndex: VT[A,(B,Int)] = v => new View.ZipWithIndex[B](vt(v))
+  }
```
julienrf: Why choose to use a type alias + an implicit class rather than a proper class?

src/main/scala/strawman/collection/View.scala
```diff
@@ -250,6 +255,19 @@ object View extends IterableFactory[View] {
...
+  def source[A]: VT[A,A] = (v: Iterable[A]) => View.fromIterable(v)
```
julienrf: I’m not a big fan of the `source` name. What about `id` (since this view transformer preserves the same collection)?

julienrf (commented):
Thanks for putting this work together. I like the overall idea. I left some minor comments.

src/main/scala/strawman/collection/Iterable.scala
pnf: Except in this case it starts out as an `Iterable` and ends up as one.  The intermediate representation as `View`s is really an implementation detail.  The closest equivalent in Clojure's transducer kit would be `sequence`, but that's even worse.

src/main/scala/strawman/collection/View.scala
pnf: I did have it as a real class originally, and after playing around a bit, the typeclass approach seemed neater.  I'm not committed to this though.  I'll fiddle around some more.

src/main/scala/strawman/collection/View.scala
pnf: Hah!  It was originally called `id`.  You seem to have inferred all my first draft ideas.  Anyway, I'm fine changing it back.

src/main/scala/strawman/collection/View.scala
pnf: It ends up looking like
```.scala
abstract class ViewTransformer[-A,+B] {
  def apply(v: Iterable[A]): View[B]
  def map[C](f: B => C) = new ViewTransformer[A,C]  {
    def apply(v: Iterable[A]): View[C] = new View.Map(ViewTransformer.this(v),f)
  }
  ...
}
```
which is fine I guess, but feels clunkier (to me).

src/main/scala/strawman/collection/Iterable.scala
pnf: Actually, whether it ends up as a `View`, depends only on whether it started out as one, which is nice.

src/main/scala/strawman/collection/View.scala
Ichoran: I think this form is more easily grasped by a larger fraction of Scala users (as it depends on a smaller more widely used set of language features).  Elegance is fine, but approachability is usually more important.

julienrf: What’s clunkier exactly? You can also define a factory method in the `ViewTransformer` companion:

~~~ scala
object ViewTransformer {
  def apply[A, B](f: Iterable[A] => View[B]): ViewTransformer[A, B] = …
}

class ViewTransformer[-A, +B] {
  def apply(as: Iterable[A]): View[B]
  def map[C](f: B => C) = ViewTransformer(it => View.Map(apply(v)), f)
}
~~~

Or, you can also use the “lambda” syntax (which has been generalized to SAM types):

~~~ scala
trait ViewTransformer[-A, +B] {
  def apply(as: Iterable[A]): View[B]
  def map[C](f: B => C): ViewTransformer[A, C] = as => View.Map(apply(as), f)
}
~~~

pnf: I didn't realize we could use lambda sugar for any SAM.  I'll do that...

julienrf:
I’m wondering if we really need `ViewTransformer` _operations_.

If we go back to a form where `ViewTransformer[-A, +B]` is just an alias for `View[A] => View[B]` we can still combine `ViewTransformer` instances as follows by defining functions. Your example would look as follows:

~~~ scala
val x: ViewTransformer[Int, Double] = xs => xs.map(i => i * 2).flatMap(i => Seq(i + 0.0, i + 0.5))
val x2: ViewTransformer[Double, (String, Int)] = xs => xs.collect { case d if d < 3.0 => s"Hey $d" }.zipWithIndex
xs.transform(x andThen x2)
~~~

hepin1989:
https://github.com/knutwalker/transducers-scala

SethTisue:

PR #148, opened 2017-07-04
Fix LinearSeq length implementation.

### LINKED ISSUE #146: List.length and LazyList.length don't work
[`List.length`](https://github.com/scala/collection-strawman/blob/f55ad8ee920120da9d463b8cd2bfaebd8cb7253b/src/main/scala/strawman/collection/immutable/List.scala#L24) and [`LazyList.length`](https://github.com/scala/collection-strawman/blob/f55ad8ee920120da9d463b8cd2bfaebd8cb7253b/src/main/scala/strawman/collection/immutable/LazyList.scala#L31) always return `0` with their current tail-recursive implementation.

I think they need to be implemented similarly to [in `LinearSeqOptimized` in 2.12](https://github.com/scala/scala/blob/65980ea57893996d071b576a826004d5e3bbcc62/src/library/scala/collection/LinearSeqOptimized.scala#L47). I don't know if there should be a similar trait in the strawman from which both `List` and `LazyList` should inherit.

julienrf:
Fixes #146.

### DIFF
--- a/src/main/scala/strawman/collection/Seq.scala
+++ b/src/main/scala/strawman/collection/Seq.scala
@@ -1,10 +1,12 @@
 package strawman.collection
 
-import scala.{Any, AnyRef, Array, Boolean, Equals, IndexOutOfBoundsException, Int, math, Ordering, Unit}
-import scala.Predef.{intWrapper}
+import scala.{Any, AnyRef, Array, Boolean, Equals, IndexOutOfBoundsException, Int, Ordering, Unit, math}
+import scala.Predef.intWrapper
 import java.lang.Object
+
 import strawman.collection.immutable.Range
 
+import scala.annotation.tailrec
 import scala.annotation.unchecked.uncheckedVariance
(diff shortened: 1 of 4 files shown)


src/main/scala/strawman/collection/Seq.scala
```diff
@@ -662,6 +664,18 @@ trait LinearSeqOps[+A, +CC[X] <: LinearSeq[X], +C <: LinearSeq[A]] extends Any w
...
+    var these = coll
+    var len = 0
+    while (!these.isEmpty) {
+      len += 1
+      these = these.tail
+    }
+    len
+  }
+
+  override def length: Int = size
```
Ichoran: We should have a policy about which method overrides which so we don't get in tangles.  In any case, when you have two synonymous methods, one should be final.  So in this case, `final def length: Int = size`, though it's a bit odd that the normal `size` defers to `iterator.length`.  It would seem more natural to me to have `length` contain the implementation here.

PR #150, opened 2017-07-04
Remove dropRight and takeRight from Iterator

### LINKED ISSUE #131: Remove `dropRight` from `Iterator`
(and same for methods that imply buffering on `Iterator`)

julienrf:
Fixes #131.

Note that I kept the `grouped` and `sliding` methods of `Iterator` even though they also use buffers. Since these methods do exist in 2.12 we have to discuss before we remove them…

### DIFF
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -519,7 +519,7 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
     */
   def dropRight(n: Int): C = {
     val b = newSpecificBuilder()
-    if (n >= 0) b.sizeHint(coll, -n)
+    if (n >= 0) b.sizeHint(coll, delta = -n)
     val lead = coll.iterator() drop n
     val it = coll.iterator()
     while (lead.hasNext) {
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -475,7 +475,18 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
   def take(n: Int): C = fromSpecificIterable(View.Take(coll, n))
 
   /** A collection containing the last `n` elements of this collection. */
-  def takeRight(n: Int): C = fromSpecificIterable(View.TakeRight(coll, n))
+  def takeRight(n: Int): C = {
+    val b = newSpecificBuilder()
+    b.sizeHintBounded(n, coll)
+    val lead = coll.iterator() drop n
+    val it = coll.iterator()
+    while (lead.hasNext) {
+      lead.next()
+      it.next()
+    }
+    while (it.hasNext) b += it.next()
+    b.result()
+  }
 
   /** Takes longest prefix of elements that satisfy a predicate.
     *  $orderDependent
@@ -506,7 +517,17 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
   /** The rest of the collection without its `n` last elements. For
     *  linear, immutable collections this should avoid making a copy.
     */
-  def dropRight(n: Int): C = fromSpecificIterable(View.DropRight(coll, n))
+  def dropRight(n: Int): C = {
+    val b = newSpecificBuilder()
+    if (n >= 0) b.sizeHint(coll, -n)
+    val lead = coll.iterator() drop n
+    val it = coll.iterator()
+    while (lead.hasNext) {
+      b += it.next
+      lead.next()
+    }
+    b.result()
+  }
 
   /** Skips longest sequence of elements of this iterator which satisfy given
     *  predicate `p`, and returns an iterator of the remaining elements.
--- a/src/main/scala/strawman/collection/View.scala
+++ b/src/main/scala/strawman/collection/View.scala
@@ -273,9 +273,8 @@ object IndexedView {
   }
 
   class TakeRight[A](underlying: IndexedView[A], n: Int) extends IndexedView[A] {
-    private val normN = n max 0
-    private val delta = underlying.length - normN
-    def length = underlying.length min normN
+    private[this] val delta = (underlying.length - (n max 0)) max 0
+    def length = underlying.length - delta
     def apply(i: Int) = underlying.apply(i + delta)
   }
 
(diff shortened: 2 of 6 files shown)


src/main/scala/strawman/collection/View.scala
```diff
@@ -113,14 +113,6 @@ object View extends IterableFactory[View] {
       if (underlying.knownSize >= 0) (underlying.knownSize - normN) max 0 else -1
   }
 
-  /** A view that drops trailing elements of the underlying collection. */
-  case class DropRight[A](underlying: Iterable[A], n: Int) extends View[A] {
-    def iterator() = underlying.iterator().dropRight(n)
-    protected val normN = n max 0
-    override def knownSize =
-      if (underlying.knownSize >= 0) (underlying.knownSize - normN) max 0 else -1
-  }
```
julienrf: I removed the `DropRight` view which was based on the `dropRight` method of `Iterator`. (same for `TakeRight` below)

src/main/scala/strawman/collection/View.scala
```diff
@@ -269,6 +253,8 @@ trait IndexedView[+A] extends View[A] with ArrayLike[A] { self =>
     }
   }
 
+  override def knownSize: Int = length
```
julienrf: That’s unrelated to the PR, actually, but I thought we could override `knownSize` in the case of `IndexedView`. We should do the same for `IndexedSeq`.

src/main/scala/strawman/collection/View.scala
```diff
@@ -286,11 +272,11 @@ object IndexedView {
...
 
-  class TakeRight[A](underlying: IndexedView[A], n: Int)
-  extends View.TakeRight(underlying, n) with IndexedView[A] {
-    override def iterator() = super.iterator() // needed to avoid "conflicting overrides" error
+  class TakeRight[A](underlying: IndexedView[A], n: Int) extends IndexedView[A] {
+    private val normN = n max 0
+    private val delta = underlying.length - normN
     def length = underlying.length min normN
-    def apply(i: Int) = underlying.apply(i)
+    def apply(i: Int) = underlying.apply(i + delta)
```
julienrf: I kept the `IndexedView` based implementation of `takeRight` (and `dropRight`, below)

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -475,7 +475,18 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
...
 
   /** A collection containing the last `n` elements of this collection. */
-  def takeRight(n: Int): C = fromSpecificIterable(View.TakeRight(coll, n))
+  def takeRight(n: Int): C = {
+    val b = newSpecificBuilder()
+    b.sizeHintBounded(n, coll)
+    val lead = coll.iterator() drop n
+    val it = coll.iterator()
+    while (lead.hasNext) {
+      lead.next()
```
Ichoran: This is a cute algorithm, but I can't imagine it's faster than just going
```
var m = this.size
val it = coll.iterator()
while (m > n) it.next
while (it.hasNext) b += it.next()
```

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -506,7 +517,17 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
   /** The rest of the collection without its `n` last elements. For
     *  linear, immutable collections this should avoid making a copy.
     */
-  def dropRight(n: Int): C = fromSpecificIterable(View.DropRight(coll, n))
+  def dropRight(n: Int): C = {
+    val b = newSpecificBuilder()
+    if (n >= 0) b.sizeHint(coll, -n)
```
Ichoran: Why `-n`?

src/main/scala/strawman/collection/View.scala
```diff
@@ -286,11 +272,11 @@ object IndexedView {
     def apply(i: Int) = underlying.apply(i)
   }
 
-  class TakeRight[A](underlying: IndexedView[A], n: Int)
-  extends View.TakeRight(underlying, n) with IndexedView[A] {
-    override def iterator() = super.iterator() // needed to avoid "conflicting overrides" error
+  class TakeRight[A](underlying: IndexedView[A], n: Int) extends IndexedView[A] {
+    private val normN = n max 0
```
Ichoran: You don't need to store this value.  `def length = underlying.length - delta` is enough.

src/main/scala/strawman/collection/View.scala
```diff
@@ -300,16 +286,13 @@ object IndexedView {
     def apply(i: Int) = underlying.apply(i + normN)
   }
 
-  class DropRight[A](underlying: IndexedView[A], n: Int)
-  extends View.DropRight(underlying, n) with IndexedView[A] {
-    override def iterator() = super.iterator()
+  class DropRight[A](underlying: IndexedView[A], n: Int) extends IndexedView[A] {
+    private val normN = n max 0
```
Ichoran: You can just store the actual length here; you never need `n max 0` separately.

Ichoran:
Looksreasonable.  I made some comments but didn't do a formal review.

src/main/scala/strawman/collection/Iterable.scala
julienrf: That would have the benefit of requiring only one traversal, indeed (excepted in the case of linear collections). I should give it a try.

src/main/scala/strawman/collection/Iterable.scala
julienrf: This version of `sizeHint` takes a collection and a delta to apply to the collection’s size. Here we want to say that the size of the result will be the size of `coll` minus the number of dropped elemnts. I could write explicitly `sizeHint(coll, delta = -n)`, for clarity.

src/main/scala/strawman/collection/Iterable.scala
julienrf: I tried it but didn’t notice a difference (in `List`):

takeright

Ichoran: Wait, what?  List shouldn't rebuild on `takeRight` anyway--if it does, that needs to be fixed.  Also, I didn't suggest the other method is faster, just more obviously correct.

julienrf: Indeed good point, that’s anyway something that we should fix…

PR #151, opened 2017-07-04
Split Seq.scala file into three files

julienrf:


Ichoran:
LGTM but if you're just going to merge it, kind of odd to ask for a review?

julienrf:
Sorry Rex, in general your reviews are very helpful, in that case that’s a very minor change but it can introduce merge conflicts if another PR is merged in the meanwhile (because I just moved things). So, I wanted to merge it as fast as possible. I still added you as a reviewer so that you get a notification.

Ichoran:
It's not a problem, I just didn't understand what you intended to have happen.

PR #153, opened 2017-07-05
Add `@throws` annotations; fix iterators

### LINKED ISSUE #152: Add throw annotations
Add `@throws` annotations to relevant methods, such as `Iterator.next()` and `ArrayLike.apply(Int)`.

I don't know if this is a priority right now, but for both of the methods above (and there may be others), implementations should in most cases have a check which throws an appropriate exception, and `@throws` annotations might help to make that more obvious.

NthPortal:
Add `@throws` annotations for `Map.apply`, `Seq.apply` and `Iterator.next()`.

Fix two cases where iterators did not bounds check.

Attempts to resolve #152 (though I may not have thought of some methods which should have `@throws` annotations).

### DIFF
--- a/src/main/scala/strawman/collection/ArrayOps.scala
+++ b/src/main/scala/strawman/collection/ArrayOps.scala
@@ -1,7 +1,7 @@
 package strawman
 package collection
 
-import scala.{AnyVal, Array, Char, Int}
+import scala.{AnyVal, Array, ArrayIndexOutOfBoundsException, Char, Int, throws}
 import scala.Predef.???
 import mutable.{ArrayBuffer, GrowableBuilder}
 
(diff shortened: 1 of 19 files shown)


src/main/scala/strawman/collection/ArrayOps.scala
```diff
@@ -19,6 +19,7 @@ class ArrayOps[A](val xs: Array[A])
   protected[this] def seq: Seq[A] = iterableFactory.fromIterable(coll)
 
   def length = xs.length
+  @throws[ArrayIndexOutOfBoundsException]
   def apply(i: Int) = xs.apply(i)
```
julienrf: Isn’t it enough to put the annotation on the parent (`ArrayLike`) class only?

NthPortal: There are a few reasons I added the annotation there.

- It is a more specific exception than that thrown by its `super`
- Though somewhat unlikely, the class could conceivably be inherited from (I've inherited from `Ordering.Ops` before)
- Even if not necessary for the implementer, it's helpful for a caller to be able to see that it throws an exception

Additionally, I don't feel that there's a significant cost incurred by adding the annotation, though you and others may disagree.

julienrf:
Thanks :)

PR #156, opened 2017-07-11
Add missing parentheses

nicolasstucki:


nicolasstucki:
This change is mostly for dotty as these became errors.

julienrf:
Thanks, looks good to me :)

PR #158, opened 2017-07-11
Upgrade Dotty to 0.2.0-RC1, including IDE support

smarter:
Dotty now has IDE support! To try it out, see http://dotty.epfl.ch/docs/usage/ide-support.html


szeiger:
Do you have any ideas for the type inference problems on Dotty mentioned in https://github.com/scala/collection-strawman/pull/117?

smarter:
Nope, please open issues against dotty for that and we'll have a look!

build.sbt
```diff
@@ -55,6 +60,8 @@ val timeBenchmark =
     .enablePlugins(JmhPlugin)
     .settings(commonSettings: _*)
     .settings(
+      // FIXME: does not compile with dotty
+      crossScalaVersions := List(scalaVersion.value),
```
julienrf: That’s weird, what’s the compilation error you get?

build.sbt
```diff
@@ -1,6 +1,6 @@
 // Convenient setting that allows writing `set scalaVersion := dotty.value` in sbt shell to switch from Scala to Dotty
 val dotty = settingKey[String]("dotty version")
-dotty in ThisBuild := dottyLatestNightlyBuild.get
+dotty in ThisBuild := "0.2.0-RC1"
```
julienrf: What’s the benefit of 0.2.0-RC1 over the latest nightly build?

smarter: You weren't actually using the latest nightly build previously, because the version of sbt-dotty used looked for the latest nightly in dotty_0.1 but we're now at dotty_0.2 (and soon at _0.3). I think it makes more sense to use a stable version now that we have one, and use the https://github.com/lampepfl/dotty-community-build to make sure the latest nightly still works.

build.sbt
smarter: Various java parsing errors:
```scala
[error] -- [E006] Unbound Identifier Error: /home/smarter/opt/collection-strawman/benchmarks/time/src/main/java/streams/Java8Streams.java:19:14 
[error] 19 |    public int size;
[error]    |              ^
[error]    |              not found: ???
```
And this:
```scala
[error] -- Error: /home/smarter/opt/collection-strawman/benchmarks/time/src/main/scala/strawman/collection/immutable/ScalaTreeSetBenchmark.scala:60:62 
[error] 60 |  def map(bh: Blackhole): Unit = bh.consume(xs.map(x => x + 1))
[error]    |                                                              ^
[error]    |ambiguous implicits: both method newCanBuildFrom in object SortedSet and method newCanBuildFrom in class SortedSetFactory match type scala.collection.generic.CanBuildFrom[
[error]    |  scala.collection.immutable.TreeSet[scala.Long]
[error]    |, scala.Long, That] of parameter bf of method map in trait SetLike
[error] -- [E050] Reference Error: /home/smarter/opt/collection-strawman/benchmarks/time/src/main/scala/strawman/collection/immutable/ScalaTreeSetBenchmark.scala:60:36 
[error] 60 |  def map(bh: Blackhole): Unit = bh.consume(xs.map(x => x + 1))
[error]    |                                 ^^^^^^^^^^
[error]    |Ambiguous overload. The overloaded alternatives of method consume in class Blackhole with types
[error]    | (x$0: scala.Double): scala.Unit
[error]    | (x$0: scala.Float): scala.Unit
[error]    | (x$0: scala.Long): scala.Unit
[error]    | (x$0: scala.Int): scala.Unit
[error]    | (x$0: scala.Short): scala.Unit
[error]    | (x$0: scala.Char): scala.Unit
[error]    | (x$0: scala.Boolean): scala.Unit
[error]    | (x$0: scala.Byte): scala.Unit
[error]    | (x$0: scala.Any): scala.Unit
[error]    |all match arguments (That)
```

I haven't had time to minimize them and open bug reports against dotty unfortunately

build.sbt
julienrf: OK, thanks for the clarification!

build.sbt
julienrf: I think the ambiguous implicits are expected. We should explicitly insert these parameters.

For the Java parsing errors, I guess this is a bug in Dotty?

julienrf: I added the `-noPredef` option to avoid ambiguities / name clashes with standard collections but I don’t think we need it. Alternatively, we could fix Dotty’s Java parser by generating `scala.Predef.???` instead of just `???`.

odersky: I would classify it as a bug. The Java parser of Dotty should be fixed. I'll do a PR shortly. But how can we used a fixed Dotty in strawman nightly builds? /cc @smarter

smarter: We can just update the dotty version in the build.sbt to be a nightly build with the fix in.

julienrf:
I added a commit to explicitly supply the `CanBuildFrom` instance.

The benchmarks still don’t compile because Dotty 0.2.0-RC1 does not include https://github.com/lampepfl/dotty/pull/2865 but this is not a blocker for me.

julienrf:
Thanks @smarter

PR #161, opened 2017-07-18
Add StrictOptimizedSeqOps

### LINKED ISSUE #125: Add a StrictOptimizedSeqOps trait with more efficient updated, append and prepend methods
As explained in #115, the current implementation is less efficient that what we can do with strict builders.

~~~ scala
package strawman
package collection
package immutable

trait StrictOptimizedSeqOps[+A, +CC[_], +C] extends StrictOptimizedIterableOps[A, C] {

  protected[this] def newBuilder[B](): Builder[B, CC[B]]

  // override `updated`, `prepend` and `append` methods

}
~~~

julienrf:
Fixes #125.

We get a performance boost on some collection types by overriding some operations to use strict builders. For instance, here is the benchmark for `span` on `Vector`:

span

Here is the benchmark for `append` on `List`:

append

Some operations are worse when implemented on top of builders (so, for those, I kept the view-based implementation). E.g. `padTo` on `List`:

padto

I’m still not 100% sure it is worth adding this `StrictOptimizedSeqOps` trait because it is currently inherited only by `Vector`, `List` and `ImmutableArray`, but only `List` actually benefits from the optimizations brought to `append` and `updated` because `Vector` and `ImmutableArray` override these operations anyway. Maybe we should just override `List` methods without introducing `StrictOptimizedSeqOps`…

### DIFF
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/ListBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/ListBenchmark.scala
@@ -21,6 +21,7 @@ class ListBenchmark {
 
   var xs: List[Long] = _
   var xss: scala.Array[List[Long]] = _
+  var zipped: List[(Long, Long)] = _
   var randomIndices: scala.Array[Int] = _
 
   @Setup(Level.Trial)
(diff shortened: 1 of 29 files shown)


julienrf:
Actually my last comment was wrong, the overrides performed in `StrictOptimizedSeqOps` also benefit to `Range` and `NumericRange` in addition to `List`. So let’s keep it!

Issue #78, opened 2017-05-11
Separating interfaces and implementations


julienrf:
# Context

`XxxOps` types provide convenient default implementations for most methods. Some of them rely on protected `coll`, `fromIterable` and `fromSpecificIterable` members ([example](https://github.com/scala/collection-strawman/blob/c7b52041058c8aa4fb36376b5f6e9188acd1f4e9/src/main/scala/strawman/collection/Iterable.scala#L156)). Often the `coll` member’s type is refined so that more complex operations can be implemented (see [here](https://github.com/scala/collection-strawman/blob/c7b52041058c8aa4fb36376b5f6e9188acd1f4e9/src/main/scala/strawman/collection/immutable/Set.scala#L17), and [there](https://github.com/scala/collection-strawman/blob/c7b52041058c8aa4fb36376b5f6e9188acd1f4e9/src/main/scala/strawman/collection/immutable/Set.scala#L42-L47)). For convenience, this member is implemented in `Iterable`, as follows:

~~~ scala
def coll: this.type = this
~~~

# Problems

- When we refine the type of `coll` to be `C` or `CC[A]`, we then have to put upper bounds on these `C` and `CC` type parameters. Then these upper bounds show up everywhere and make the code less readable ([example](https://github.com/scala/collection-strawman/blob/c7b52041058c8aa4fb36376b5f6e9188acd1f4e9/src/main/scala/strawman/collection/immutable/SortedSet.scala#L14-L16)) ;
- The signature of `fromIterable` and `fromSpecificIterable` is impossible to implement for specialized `View`s (like `IndexedView`). See [this comment](https://github.com/scala/collection-strawman/pull/76#issuecomment-299475201) for more details. A derived consequence of this is that transformation operations on `IndexedView` return a `View` unless they are explicitly overriden ([example](https://github.com/scala/collection-strawman/blob/c7b52041058c8aa4fb36376b5f6e9188acd1f4e9/src/main/scala/strawman/collection/View.scala#L170-L175)).

# Proposal

- Having traits with no implemented methods would make it possible to remove the bounds on the type parameters. (But that might just delay the problem because if we want to have reusable implementations inherited by concrete collections then those reusable implementations will have the bounds)
- I think that another consequence will be that Java collection’s decorators and Views will be better integrated (e.g. it will be possible to make `IndexedView[A]` extend `Seq[A]`, and it will be simpler to specialize its operations’ return type to `IndexedView[A]`).

[...]

PR #165, opened 2017-07-20
Add withFilter

### LINKED ISSUE #145: Implement withFilter

julienrf:
This started nicely. But then… things got worse when it was time to support sorted and Map collections.

I used Views to implement `WithFilter`. I noticed that in the current collections `withFilter` calls are fused, so I implemented `filter` fusion on Views too.

That’s it for the nice part.

Now, what’s the problem with Map and sorted collections? The implementation of `WithFilter` delegates to `iterableFactory` to build the collection resulting of a `map` or `flatMap` call subsequent to `withFilter`. However, in the case of `Map`, `iterableFactory` builds an `Iterable` instead of the expected concrete `Map`, and in the case of `SortedSet`, `iterableFactory` builds a `Set` instead of the expected concrete `SortedSet`. Consequently, I created subclasses of `WithFilter` for each case (`SortedWithFilter`, `MapWithFilter` and `SortedMapWithFilter`). Then, I refined the `withFilter` member in `Map`, `SortedSet` and `SortedMap` to return the corresponding specialized `XxxWithFilter`. Since this is a refinement, it means that all of these `XxxWithFilter` classes have to extend the initial `WithFilter` class (in the same way that all collection types extend `Iterable`). I’m not happy with the result because there are lots of parameters and type parameters. However, I don’t have a better solution :(

Fixes #145.


src/test/scala/strawman/collection/test/WithFilterTest.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  @Test
+  def sorted(): Unit = {
+    val xs1 =
+      for {
+        x <- TreeSet(1, 2, 3)
+        if x % 2 == 0
+      } yield x + 1
+    val xs1t: TreeSet[Int] = xs1
+    assert(xs1 == TreeSet(3))
```
julienrf: This test fails on Dotty. This is probably something related to hashcodes. We observed a weird behavior [here](https://github.com/scala/collection-strawman/pull/48#issuecomment-311615632) too. I’ll create an issue in Dotty.

julienrf: On Dotty `xs1 == TreeSet()`.

julienrf: Actually that’s not related to hashcodes, but maybe to inlining in Dotty. I added a commit that un-inlines some methods and now the tests pass on Dotty. I’m trying to minimize the problem.

Ichoran:
I agree this looks pretty bad but I haven't had time to come up with anything better, if it's even possible to come up with something better.  Is it possible to simply not provide a default implementation for the difficult cases?  If we were worrying about CBF, something with five type parameters is not going to go over well.

julienrf:
What do you mean by “not provide a _default_ implementation”? If we remove the complex subclasses of `WithFilter` (e.g. the one in `SortedMap`) then people using an `if` in a `for` expression whose generator is a `SortedMap` will get back a `Map`. I think that would be a bit ridiculous: `for` without `if` would return back the sorted `Map`, but as soon as you introduce an `if` you loose sortedness.

In my opinion, the PR is fine to merge, these complex classes are not aimed to be manipulated by end-users.

Ichoran:
I mean that we manually write the `WithFilter` for each leaf class with no abstraction at all (beyond the element type).  It gives us a bit more work to do but it makes the idea of implementing your own class more approachable.  (More work to actually do, but less work to understand, and usually the understanding is more of a barrier than the doing.)

julienrf:
> I mean that we manually write the WithFilter for each leaf class with no abstraction at all (beyond the element type).

For `Map` collection types that would mean implement 6 methods (two overloads of `map`, two overloads of `flatMap`, `foreach` and `withFilter`.

Ichoran:
Yes, but we don't have _that_ many map collection types.  Anyway, it's not a very satisfying answer, but it's the best idea I have right now (aside from the scheme you already came up with).

odersky:
I wonder, why is the implementation in the existing collections so much simpler? What do they do that we cannot? CBF?

Ichoran:
I haven't had time to work this all the way through, but it seems to me if the `WithFilter` hierarchy exactly parallels the `Ops` hierarchy, you "only" need exactly the same type arguments (no extra ones).  This is a little more bulky than what you've already got, but it's pretty close.

I haven't yet had time to play with making `WithFilter` an inner class of the `Ops` trait.  In that case it would simply adopt all the type parameters, and I think that would be enough, but I haven't worked out whether it's possible or desirable to provide alternate implementations.  If nobody else gets to it, I'll play with this next time I have a chance.  Please let me know if you have, though, so I don't duplicate work.

julienrf:
@Ichoran I’m trying something with `BuildFrom`. So far it works great (only one implementation of `WithFilter`) but I have issues related to #134 / #78. Maybe I’ll fix all of them at the same time.

Without `BuildFrom`, even with an inner class I’m afraid that you still need 4 kinds of `WithFilter` because we currently have no way to abstract over the arity of the collection type constructor and over the implicit constraints.

Ichoran:
I don't think there's anything wrong with four kinds of `WithFilter`, just with five generic type parameters.  If instead it has _zero_ type parameters because it adopts everything it needs from its parent class, it presents a much friendlier interface.  Yes, it does make abstracting over `WithFilter`s awkward, but `WithFilter` is strange anyway so I don't think we lose much.

julienrf:
OK, I see. I’ll experiment with that tomorrow.

julienrf:
I moved WithFilter to be inner classes. The resulting classes have only one parameter (the filtered collection).

However now the API doc shows this:

withfilters

Last, I had variance errors in Dotty but not in Scala 2.12 nor 2.13. I added `@uncheckedVariance` annotations but I’m actually sure who is right, Dotty or Scala?

PR #168, opened 2017-07-24
Implicit factories

### LINKED ISSUE #139: Type-driven builders
Explore how we can turn [this line](https://github.com/scala/collection-strawman/blob/24af74ec093f0f011cc729e7bd19b223e976853e/src/test/scala/strawman/collection/test/GenericTest.scala#L51) into one that `implicitly` get the parser (instead of having it explicitly supplied).

julienrf:
Fixes #139.

The approach is to provide not just implicit `BuildFrom` instances but implicit `FromSpecificIterable` instances (which have a `newBuilder()` method that takes no initial collection).

However this solution raises a ambiguous implicits error when used with sorted Map collections (e.g. `TreeMap`). But I believe this is a limitation of the current way we prioritize implicits. See my comments in the diff for more details.

### DIFF
--- a/src/main/scala/strawman/collection/Factories.scala
+++ b/src/main/scala/strawman/collection/Factories.scala
@@ -24,36 +24,6 @@ trait BuildFrom[-From, -A, +C] extends Any {
   def newBuilder(from: From): Builder[A, C]
 }
 
-object BuildFrom extends BuildFromLowPriority {
-  /** Build the source collection type from a MapOps */
-  implicit def buildFromMapOps[CC[K, V] <: Map[K, V] with MapOps[K, V, CC, _], A, B, E, F]: BuildFrom[CC[A, B], (E, F), CC[E, F]] = new BuildFrom[CC[A, B], (E, F), CC[E, F]] {
-    //TODO: Reuse a prototype instance
-    def newBuilder(from: CC[A, B]): Builder[(E, F), CC[E, F]] = from.mapFactory.newBuilder[E, F]()
-    def fromSpecificIterable(from: CC[A, B])(it: Iterable[(E, F)]): CC[E, F] = from.mapFactory.fromIterable(it)
-  }
-
-  /** Build the source collection type from a SortedMapOps */
-  implicit def buildFromSortedMapOps[CC[K, V] <: SortedMap[K, V] with SortedMapOps[K, V, CC, _], A, B, E : Ordering, F]: BuildFrom[CC[A, B], (E, F), CC[E, F]] = new BuildFrom[CC[A, B], (E, F), CC[E, F]] {
-    def newBuilder(from: CC[A, B]): Builder[(E, F), CC[E, F]] = from.sortedMapFactory.newBuilder[E, F]()
-    def fromSpecificIterable(from: CC[A, B])(it: Iterable[(E, F)]): CC[E, F] = from.sortedMapFactory.fromSpecificIterable(it)
-  }
-}
-
-trait BuildFromLowPriority {
-  /** Build the source collection type from an IterableOps */
-  implicit def buildFromIterableOps[CC[X] <: Iterable[X] with IterableOps[X, CC, _], A, E]: BuildFrom[CC[A], E, CC[E]] = new BuildFrom[CC[A], E, CC[E]] {
-    //TODO: Reuse a prototype instance
-    def newBuilder(from: CC[A]): Builder[E, CC[E]] = from.iterableFactory.newBuilder[E]()
-    def fromSpecificIterable(from: CC[A])(it: Iterable[E]): CC[E] = from.iterableFactory.fromIterable(it)
-  }
-
-  /** Build the source collection type from an Iterable with SortedOps */
-  implicit def buildFromSortedOps[CC[X] <: Iterable[X] with SortedOps[X, CC, _], A, E : Ordering]: BuildFrom[CC[A], E, CC[E]] = new BuildFrom[CC[A], E, CC[E]] {
-    def newBuilder(from: CC[A]): Builder[E, CC[E]] = from.sortedIterableFactory.newBuilder[E]()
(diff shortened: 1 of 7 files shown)


src/main/scala/strawman/collection/Factories.scala
```diff
@@ -24,36 +24,6 @@ trait BuildFrom[-From, -A, +C] extends Any {
...
-    def newBuilder(from: CC[A]): Builder[E, CC[E]] = from.iterableFactory.newBuilder[E]()
-    def fromSpecificIterable(from: CC[A])(it: Iterable[E]): CC[E] = from.iterableFactory.fromIterable(it)
-  }
-
-  /** Build the source collection type from an Iterable with SortedOps */
-  implicit def buildFromSortedOps[CC[X] <: Iterable[X] with SortedOps[X, CC, _], A, E : Ordering]: BuildFrom[CC[A], E, CC[E]] = new BuildFrom[CC[A], E, CC[E]] {
-    def newBuilder(from: CC[A]): Builder[E, CC[E]] = from.sortedIterableFactory.newBuilder[E]()
-    def fromSpecificIterable(from: CC[A])(it: Iterable[E]): CC[E] = from.sortedIterableFactory.fromSpecificIterable(it)
-  }
-}
```
julienrf: These definitions are not anymore necessary: they have been superseded by the implicit definitions in factories.

src/test/scala/strawman/collection/test/GenericTest.scala
```diff
@@ -48,10 +47,8 @@ class GenericTest {
 
   @Test
   def genericTest: Unit = {
-    assert(Parse.parseCollection[Int, immutable.List[Int]](immutable.List).parse("1|2|3").contains(1 :: 2 :: 3 :: immutable.Nil))
-
-    // TODO wrap with assert when HashMap’s equality is correctly implemented
-    Parse.parseCollection[(Int, Int), immutable.HashMap[Int, Int]](immutable.HashMap).parse("1-2|3-4").contains(immutable.HashMap((1, 2), (3, 4)))
+    assert(Parse.parseCollection[Int, immutable.List[Int]].parse("1|2|3").contains(1 :: 2 :: 3 :: immutable.Nil))
+    assert(Parse.parseCollection[(Int, Int), immutable.HashMap[Int, Int]].parse("1-2|3-4").contains(immutable.HashMap((1, 2), (3, 4))))
```
julienrf: The factory is now provided implicitly, which was the initial goal of this PR.

src/main/scala/strawman/collection/Factories.scala
```diff
@@ -73,6 +43,7 @@ trait IterableFactory[+CC[_]] {
   def apply[A](xs: A*): CC[A] = fromIterable(View.Elems(xs: _*))
   def fill[A](n: Int)(elem: => A): CC[A] = fromIterable(View.Fill(n)(elem))
   def newBuilder[A](): Builder[A, CC[A]]
+  implicit def canBuildIterable[A]: FromSpecificIterable[A, CC[A]] = IterableFactory.toSpecific(this)
```
julienrf: BTW, I’d like to rename `FromSpecificIterable[A, C]` to `CanBuild[A, C]`. If users write code that takes parameters of that type, I think `CanBuild` conveys a clearer meaning (“ability to build a collection `C` from elements of type `A`”) than `FromSpecificIterable`. WDYT?

src/test/scala/strawman/collection/test/TraverseTest.scala
```diff
@@ -128,12 +128,16 @@ class TraverseTest {
     val xs3: immutable.List[String] = xs2
 
     val xs4 = immutable.TreeMap((1, "1"), (2, "2"))
-    val xs5 = flatCollect(xs4) { case (2, v) => immutable.List((v, v)) }
+    val xs5 = flatCollect(xs4) { case (2, v) => immutable.List((v, v)) }(immutable.TreeMap)
```
julienrf: So, this is the issue. Here the `TreeMap` factory hasn’t been inferred because of an ambiguity with `SortedSet`.

Here is what happens, if I understand correctly. We are looking for an implicit `BuildFrom[TreeMap[Int, String], (String, String), C]`. The two candidates we find are:
- `SortedSet.canBuildSortedIterable`, of type `[A] => BuildFrom[Any, A, SortedSet[A]]`
- `TreeMap.canBuildSortedMap`, of type `[K, V] => BuildFrom[Any, (K, V), TreeMap[K, V]]`

Why are these candidates ambiguous? Because none is more specific than the other: `TreeMap[?, ?]` and `SortedSet[?]` are not subtypes of each other.

Why are they both available in the implicit scope? The one from the companion object `TreeMap` is available because we are looking for an implicit `BuildFrom[TreeMap[Int, String], ?, ?]`. For the other one, I guess this is because somewhere in the hierarchy of `SortedMap` we have a `SortedSet` that shows up (`SortedMap[K, V] extends SortedOps[K, SortedSet, SortedMap[K, V]]`). Consequently, the companion object of `SortedSet` contributes to the implicit scope.

How to fix the problem? Unfortunately, I don’t think we can make the types of one of the two implicits more specific than the other (e.g. we can not make `SortedMap[K, V]` _extend_ `SortedSet[K]` because they extend `Iterable[(K, V)]` and `Iterable[K]`, respectively). One solution could be to **change the priority of implicits**: when exploring companion objects, the further we go in the hierarchy (ie when we explore companions of _ancestors_ of `TreeMap`), the lower the priority of their implicit definitions should be. Does that make sense? With this change, I don’t think we could break existing code. However, code that doesn’t compile now (because of an ambiguity) may compile after this change.

julienrf: I created https://github.com/scala/bug/issues/10427 and https://github.com/lampepfl/dotty/issues/2911

julienrf:
I just pushed a workaround for the ambiguity: I no longer mention `SortedSet` in the definition of `SortedMap`, so that `SortedSet`’s companion does not contribute to the implicit scope when we manipulate sorted Map collections.

julienrf:
The PR is ready to merge.

PR #172, opened 2017-07-25
Drop PartialFunction type annotations.

odersky:
Type inference in dotty can now handle this without needing
the annotation.


odersky:
It works on my dotty master. If it fails, maybe retry tomorrow after getting the changes in the nightly.

julienrf:
This is great! I can confirm that it compiles with the latest Dotty nightly and I was even able to remove one more type annotation.

However, this code still doesn’t compile on 2.12 nor 2.13 since https://github.com/scala/scala/pull/5975 hasn’t been merged yet.

szeiger:
Should work in 2.13.x now but not in 2.12.x. We can do this after moving to scala/scala.

julienrf:
Superseded by https://github.com/scala/scala/pull/6511

PR #173, opened 2017-07-26
Add WithFilter (using BuildFrom)

julienrf:
This is an alternative to #165 using `BuildFrom` to abstract over the different kinds of collections (`Map` and `Sorted`).

The resulting implementation is probably simpler (only one class `WithFilter`, no subclasses of it). However the user facing `map` and `flatMap` methods of `WithFilter` now take an implicit `BuildFrom` parameter (which is not consistent with the way `map` and `flatMap` are defined in `Iterable` and … well it re-introduces the `BuildFrom` that we tried to get rid of…).

Also, to implement this version I had to generalize the solution I implemented to make `StringOps` and `ArrayOps` able to extend `SeqOps`. More details in the diff.

### DIFF
--- a/src/main/scala/strawman/collection/BitSet.scala
+++ b/src/main/scala/strawman/collection/BitSet.scala
@@ -12,8 +12,6 @@ trait BitSetOps[+C <: BitSet with BitSetOps[C]]
   extends SortedSetOps[Int, SortedSet, C] {
   import BitSetOps._
 
-  protected[this] def coll: C
-
   final def ordering: Ordering[Int] = Ordering.Int
 
   /** The number of words (each with 64 bits) making up the set */
@@ -88,7 +86,7 @@ trait BitSetOps[+C <: BitSet with BitSetOps[C]]
   }
 
   def rangeImpl(from: Option[Int], until: Option[Int]): C = {
-    val a = coll.toBitMask
+    val a = c.toBitMask
     val len = a.length
     if (from.isDefined) {
       var f = from.get
@@ -111,7 +109,7 @@ trait BitSetOps[+C <: BitSet with BitSetOps[C]]
       }
       if (w < len) a(w) &= (1L << b)-1
     }
-    coll.fromBitMaskNoCopy(a)
+    c.fromBitMaskNoCopy(a)
   }
 
   /** Computes the symmetric difference of this bitset and another bitset by performing
@@ -122,11 +120,11 @@ trait BitSetOps[+C <: BitSet with BitSetOps[C]]
     *              bitset or the other bitset that are not contained in both bitsets.
     */
   def xor(other: BitSet): C = {
-    val len = coll.nwords max other.nwords
+    val len = c.nwords max other.nwords
     val words = new Array[Long](len)
     for (idx <- 0 until len)
-      words(idx) = coll.word(idx) ^ other.word(idx)
-    coll.fromBitMaskNoCopy(words)
+      words(idx) = c.word(idx) ^ other.word(idx)
+    c.fromBitMaskNoCopy(words)
   }
 
   @`inline` final def ^ (other: BitSet): C = xor(other)
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -32,6 +32,8 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
 
   protected[this] def coll: Iterable[A]
 
+  protected[this] def c: C
+
   protected[this] def fromSpecificIterable(coll: Iterable[A]): C
 
   protected[this] def fromIterable[E](it: Iterable[E]): CC[E] = iterableFactory.fromIterable(it)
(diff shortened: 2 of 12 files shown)


src/main/scala/strawman/collection/BitSet.scala
```diff
@@ -12,8 +12,6 @@ trait BitSetOps[+C <: BitSet with BitSetOps[C]]
   extends SortedSetOps[Int, SortedSet, C] {
   import BitSetOps._
 
-  protected[this] def coll: C
```
julienrf: This refinement of the type of `coll` is now unnecessary because all `Iterable`s have a `c: C` member (see below).

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -32,6 +32,8 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
 
   protected[this] def coll: Iterable[A]
 
+  protected[this] def c: C
```
julienrf: I add this member to `IterableOps[A, CC[_], C]`. Why is it useful? In this PR it’s needed by `WithFilter`, which uses a `BuildFrom` instance to build the result of `map` and `flatMap`, and `BuildFrom` needs an initial collection `C` to create a builder.

Note, that this member was previously already existing in `SeqOps`. It was needed because `SeqOps` is extended by (at least) `StringOps` and `List`, and to implement methods of `SeqOps` we sometimes needed a `C` (ie. `String` or `List[A]`) value and sometimes an `Iterable[A]` (ie. `StringView` or `List[A]`) value. So, with just one member (`coll`) we couldn’t get both a `String` and an `Iterable[Char]`, that’s why we needed two members: `coll: Iterable[A]` and `c: C`.

src/main/scala/strawman/collection/Map.scala
```diff
@@ -8,16 +8,16 @@ import scala.annotation.unchecked.uncheckedVariance
 import scala.util.hashing.MurmurHash3
 
 /** Base Map type */
-trait Map[K, +V] extends Iterable[(K, V)] with MapOps[K, V, Map, Map[K, V]]
+trait Map[K, +V] extends Iterable[(K, V)] with MapOps[K, V, Map, Map[K, V]] {
+  protected[this] def c: this.type = this
```
julienrf: I provide a default implementation of `c` in the `Map` branch (and also `Set` and `Seq`, see below). The implementation of `c` differs only in the case of decorators (such as `StringOps` and `ArrayOps` -- and, more generally, if we want to be able to decorate any Java collection we will be able to use the same pattern).

src/main/scala/strawman/collection/WithFilter.scala
```diff
@@ -0,0 +1,76 @@
...
+
+import scala.{Boolean, Unit}
+
+/** A template trait that contains just the `map`, `flatMap`, `foreach` and `withFilter` methods
+  *  of trait `Iterable`.
+  *
+  * @tparam A Element type (e.g. `Int`)
+  * @tparam C Collection type (e.g. `List[Int]`)
+  */
+class WithFilter[+A, +C](coll: C, iterable: Iterable[A], p: A => Boolean) {
```
julienrf: Here we need both a `coll: C` and an `iterable: Iterable[A]` parameters because `BuildFrom` needs the former to create a builder and we need the latter to iterate through the elements of the collection (we can not get an evidence that `C` is an `Iterable[A]` because that’s not always the case, as explained previously, e.g. think of `StringOps`)

src/main/scala/strawman/collection/WithFilter.scala
```diff
@@ -0,0 +1,76 @@
...
+  /** Further refines the filter for this $coll.
+    *
+    *  @param q   the predicate used to test elements.
+    *  @return    an object of class `WithFilter`, which supports
+    *             `map`, `flatMap`, `foreach`, and `withFilter` operations.
+    *             All these operations apply to those elements of this $coll which
+    *             satisfy the predicate `q` in addition to the predicate `p`.
+    */
+  def withFilter(q: A => Boolean): WithFilter[A, C] =
+    new WithFilter(coll, iterable, (a: A) => p(a) && q(a))
```
julienrf: I implement the fusion of filter here, but we should also implement it in `View.Filter`, as I did in #165, but that’s a different issue (related to `View` and not to `WithFilter`).

src/test/scala/strawman/collection/test/WithFilterTest.scala
```diff
@@ -0,0 +1,65 @@
...
+    val xs1t: Map[Char, Int] = xs1
+    assert(xs1 == Map('b' -> 2))
+
+    val xs2 =
+      for {
+        (k, v) <- Map(1 -> 'a', 2 -> 'b', 3 -> 'c')
+        if k % 2 == 0
+      } yield v
+    val xs2t: Iterable[Char] = xs2
+    assert(List('b') == xs2)
```
julienrf: We see that if we start with a `Map` we will get either another `Map` (`xs1`), as expected if we yield tuples, or an `Iterable` (`xs2`) if we yield single values.

julienrf:
My position about this work: I prefer the version of `WithFilter` given in #165 because it is more consistent with what we already have (no `BuildFrom` parameter in `map` and `flatMap`).

However I think this PR provides a good solution to #134 (unrelated to `WithFilter`, but more about the design of the `View` hierarchy and the decorators).

julienrf:
I’d rather go with #165.

PR #174, opened 2017-07-26
Distinction between collection type C and Iterable[A]

julienrf:
This PR extracts the changes performed in #173 that are unrelated with `withFilter`, and polishes them.

For more details, see [this comment](https://github.com/scala/collection-strawman/pull/173#discussion_r129514133):

> I add [a `coll: C`] member to `IterableOps[A, CC[_], C]`. Why is it useful? In this PR it’s needed by `WithFilter`, which uses a `BuildFrom` instance to build the result of `map` and `flatMap`, and `BuildFrom` needs an initial collection `C` to create a builder.
> 
> Note, that this member was previously already existing in `SeqOps`. It was needed because `SeqOps` is extended by (at least) `StringOps` and `List`, and to implement methods of `SeqOps` we sometimes needed a `C` (ie. `String` or `List[A]`) value and sometimes an `Iterable[A]` (ie. `StringView` or `List[A]`) value. So, with just one member (`coll`) we couldn’t get both a `String` and an `Iterable[Char]`, that’s why we needed two members: `coll: Iterable[A]` and `c: C`.

For me, these changes solve #134 and #78 with the benefit of not introducing a new intermediate layer (interfaces with no implementations) in the hierarchy.

### DIFF
--- a/src/main/scala/strawman/collection/ArrayOps.scala
+++ b/src/main/scala/strawman/collection/ArrayOps.scala
@@ -16,7 +16,7 @@ class ArrayOps[A](val xs: Array[A])
 
   def toIterable = ArrayView(xs)
   protected[this] def coll: Array[A] = xs
-  def toSeq: Seq[A] = iterableFactory.fromIterable(toIterable)
+  def toSeq: Seq[A] = fromIterable(toIterable)
 
   def length = xs.length
   @throws[ArrayIndexOutOfBoundsException]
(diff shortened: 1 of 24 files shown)


julienrf:
I’m not sure this is enough, actually… I tried to make `IndexedView` implement `SeqOps` (so that we benefit from all the operations defined in `SeqOps`, such that `startsWith`), but ran into lots of complications…

julienrf:
Actually I’m happy with these changes. I also renamed the `seq` method to `toSeq`, as suggested by @Ichoran.

Any thoughts, @szeiger?

PR #176, opened 2017-07-27
Port List optimizations to the strawman.

### LINKED ISSUE #162: Override List methods for performance
Several methods of `List` are overriden for performance. We should apply the same optimizations.

julienrf:
Also add `prependAll` (`++:`) operation.

Fixes #162.


julienrf:
@Ichoran I merged because of the release today but feel free to leave some comments, I’ll address them when I’m back from vacation.

Issue #181, opened 2017-08-02
Add appendAll (:++), add (+) and addAll (++) operations


odd:
With `prependAll` added we currently have the following membership affecting operations:
* `prepend` (aliased as `+:`)
* `append` (aliased as `:+`)
* `prependAll` (aliased as `++:`)
* `concat`(aliased as `++`)

For symmetry the following operations should be added:
* `appendAll` (aliased as `:++`)
* `add` (aliased as `+`)
* `addAll` (aliased as `++`, alias for `concat`)

Further, the operations `add`/`addAll`/`concat` could then be defined in `Iterable` (thus available for `Set`) as they do not entail any sequential positioning, while `prepend`/`prependAll`, `append`/`appendAll` could be defined in `Seq` since they do in fact entail sequential positioning (thus not avaliable for `Set`). For a specific `Seq` implementation `concat` could either be implemented as an alias for `prependAll` or `appendAll` depending on which is more efficient (e.g. the former for `List` and the latter for `ImmutableArray`).

julienrf:
In `Set` collections we have `+` (and its alphabetic name is `incl` instead of `add` to not be confused with the `add` of the Java collections -- which mutates the collection in place). I don’t think we should have `+` on `Seq` collections.

I don’t think we should have different names for `++` (`concat`, `addAll`, `appendAll`). It could make sense to remove `++` on `Seq` collections and have only `:++` (`appendAll`) and `++:` (`prependAll`), by symmetry with `:+` and `+:` but that will require one more step for existing code bases to migrate to the strawman (specifically, replacing `++` with `:++`).

odd:
I see your point regarding `add`/`addAll`. I still see a real value in having `appendAll` for `Seq` collections though. If removing `concat` from `Seq` as you described is considered a too big a change for ease-of-migration reasons, I believe adding `appendAll` (aliased to `:++`) while still keeping `concat` in `Seq`(thus having both methods available for `Seq` collections) is the best choice (i.e. symmetry and ease of discovery beats DRY in this case IMHO).

Ichoran:
I agree.  Symmetry beats DRY.  But that is more true of `:++` than `appendAll`, I think.  `concat` is straightforward enough.

julienrf:
Closed in #184.

PR #184, opened 2017-08-05
Add `appendAll` operator and generalize it (and `concat`) to accept `IterableOnce` (instead of `Iterable`).

odd:


odd:
See #181. This PR only provides `appendAll` (aliased to `:++`) and makes `concat` an alias to `appendAll` in `Seq`. This allows implementations of `Seq` to override (for performance) the methods: `append`, `prepend`, `appendAll` and `prependAll` (instead of `append`, `prepend`, `concat` and `prependAll`).

src/main/scala/strawman/collection/View.scala
```diff
@@ -154,14 +154,14 @@ object View extends IterableFactory[View] {
...
+  /** A view that concatenates elements of the prefix collection or iterator with the elements
+   *  of the suffix collection or iterator.
    */
-  case class Concat[A](underlying: Iterable[A], other: IterableOnce[A]) extends View[A] {
-    def iterator() = underlying.iterator() ++ other
-    override def knownSize = other match {
-      case other: Iterable[_] if underlying.knownSize >= 0 && other.knownSize >= 0 =>
-        underlying.knownSize + other.knownSize
+  case class Concat[A](prefix: IterableOnce[A], suffix: IterableOnce[A]) extends View[A] {
+    def iterator() = prefix.iterator() ++ suffix.iterator()
```
julienrf: The fact that `View` extends `Iterable` (and not just `IterableOnce`) means that it should be possible to repeatedly iterate on a same view. However, since we accept `IterableOnce`s as parameters, this promise doesn’t hold anymore. I’m wondering if that could be the source of subtle bugs…

Note that it was already the case before your PR.

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -770,6 +770,12 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
       else View.Empty
     }
 
+  /** Alias for `prependAll` */
+  def concat[B >: A](suffix: IterableOnce[B]): CC[B] = appendAll(suffix)
```
odd: Yes, I initially did this but then removed the inlining to allow inheritors to provide a more optimal implementation of concat directly if they wanted to, but I'll make it `final` (and add `@inline` as well of Dotty allows it).

src/main/scala/strawman/collection/View.scala
odd: That's interesting. In the case when a view is created from a method other than `view` of an eager collection, the view will never be iterated again but instead eagerly drained to create the result of the method (i.e. the view is method local and does not escape). Accepting `IterableOnce` for views created in such a way should be safe, while views created by calling `view` or by direct instantiation are not. Perhaps the view constructors accepting `IterableOnce` should be made package private while constructors accepting `Iterable` could be made public.

src/main/scala/strawman/collection/Iterable.scala
julienrf: Do you have an example of situation where one would want to have different implementations for `appendAll` and `concat`?

btw I just noticed that the scaladoc is wrong for this method (should be `appendAll` instead of `prependAll`).

src/main/scala/strawman/collection/View.scala
julienrf: > Perhaps the view constructors accepting IterableOnce should be made package private while constructors accepting Iterable could be made public

But then we would duplicate most of the constructors :(

julienrf: Well, only a few constructors are in that situation. We could change the `Concat` constructor to be `collection` private and maybe create another one, public, that would take two `Iterable[A]` as parameters.

julienrf:
I just pushed a commit that makes `zip` and `concat` methods take `Iterable`s instead of `IterableOnce`. This is needed because these methods are defined in `Iterable` and are supposed to return `Iterable`s but in the case of non strict collections (`LazyList` and `View`) it was possible to build an `Iterable` that was actually iterable only once:

~~~ scala
val xs = LazyList.empty[Int]
val ys = xs ++ Iterator.single(1) // `ys` has type `LazyList[Int]`
ys.foreach(println) // Prints “1”
ys.foreach(println) // Prints nothing
~~~

julienrf:
Still, if one wants to write a custom lazy collection type that doesn’t memoize it’s better to have `Iterable`s instead of `IterableOnce`s as parameters for `concat` and `zip`.

src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -770,6 +770,12 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
       else View.Empty
     }
 
+  /** Alias for `appendAll` */
+  @`inline` final def concat[B >: A](suffix: Iterable[B]): CC[B] = appendAll(suffix)
+
+  /** Alias for `concat` */
+  @`inline` final def ++ [B >: A](suffix: Iterable[B]): CC[B] = concat(suffix)
```
Ichoran: Don't have double-indirection if you don't need it.  If the "real" method is called `appendAll`, then `++` should forward directly to it.

src/main/scala/strawman/collection/immutable/Vector.scala
```diff
@@ -184,29 +184,27 @@ final class Vector[+A] private[immutable] (private[collection] val startIndex: I
...
+        case n if this.size < (n >>> Log2ConcatFaster) && that.isInstanceOf[Vector[_]] =>
+          var v = that.asInstanceOf[Vector[B]]
+          val ri = this.reverseIterator()
+          while (ri.hasNext) v = ri.next() +: v
+          v
+        case _ => super.appendAll(that)
+      }
     }
+  }
+  // TODO: ODD 2017-08-05: Add optimized implementation of `prependAll` analogous to `appendAll` above.
```
Ichoran: Maybe we can do this now instead of leaving it for later when it might be forgotten?

julienrf:
@Ichoran Done!

PR #185, opened 2017-08-05
Add patch operation to immutable.Seq.

odd:


src/main/scala/strawman/collection/View.scala
```diff
@@ -205,6 +205,65 @@ object View extends IterableFactory[View] {
...
+              current = oit.next()
+              isCurrent = true
+              true
+            } else if (uit.hasNext) {
+              current = uit.next()
+              isCurrent = true
+              true
+            } else false
+          }
+        }
```
julienrf: Your implementation doesn’t seem to behave exactly like in the current collections: you may copy more than `replaced` elements if the `other` collection contains more elements.

odd: Hmm, isn't that how it is supposed to work? 
According to the [ScalaDoc](http://www.scala-lang.org/api/current/scala/collection/SeqLike.html#patch(from:Int,that:scala.collection.GenSeq[A],replaced:Int):Seq[A]):
`def patch(from: Int, that: GenSeq[A], replaced: Int): Array[A]`:
`from`: the index of the first replaced element
`replaced`: the number of elements to drop in the original array
`returns`: a new array consisting of all elements of this array except that `replaced` elements starting from `from` are replaced by `patch`.

julienrf: Yes, good point. I think your implementation is correct with respect to this spec…

I’m a bit worried about breaking existing code, though :(

Ichoran:
The collections-laws tests (which pass) for `patch` are as follows:

```
(x.patch(n, y, m).take(n) theSameAs x.take(n))
(x.patch(n, y, m).drop(n).take(y.size) theSameAs y)
(x.patch(n, y, m).drop((0 max n)+y.size) theSameAs x.drop((0 max n)+(0 max m)))
```

If the new implementation follows these (looks okay?) we should be good.

src/main/scala/strawman/collection/View.scala
julienrf: @odd Is there any reason to not have reused the implementation of the `patch` method on `Iterator` (in the old collections)?

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -531,6 +531,50 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
     // If *both* of them have no elements then the collections are the same
     hasNext == those.hasNext
   }
+
+  /** Returns this iterator with patched values.
+    * Patching at negative indices is the same as patching starting at 0.
+    * Patching at indices at or larger than the length of the original iterator appends the patch to the end.
+    * If more values are replaced than actually exist, the excess is ignored.
```
julienrf: @odd I copy-pasted the Scaladoc of the old collections but noticed that the first two points about patching at negative indices or larger than length are not respected by your implementation. I would prefer to stick to the same specification as the old collections. We can just copy-paste the old collections’ implementation, unless you see a reason to not do that?

julienrf:
> We should also check if we can take advantage of builders in strict collections and override this implementation.

I implemented that in https://github.com/scala/collection-strawman/pull/185/commits/e6e405a15d074b6b78aaa68a4e0923b495ededb3

src/main/scala/strawman/collection/View.scala
```diff
@@ -205,6 +205,11 @@ object View extends IterableFactory[View] {
       }
   }
 
+  private[collection] class Patched[A](underlying: Iterable[A], from: Int, other: IterableOnce[A], replaced: Int) extends View[A] {
```
julienrf: I also made this class private because it takes an `IterableOnce` as parameter.

src/main/scala/strawman/collection/immutable/Seq.scala
```diff
@@ -93,6 +93,18 @@ trait SeqOps[+A, +CC[_], +C] extends collection.SeqOps[A, CC, C] {
...
+    *
+    *  @param  from     the index of the first replaced element
+    *  @param  other    the replacement sequence
+    *  @param  replaced the number of elements to drop in the original $coll
+    *  @tparam B        the element type of the returned $coll.
+    *  @return          a new $coll consisting of all elements of this $coll
+    *                   except that `replaced` elements starting from `from` are replaced
+    *                   by `other`.
+    */
+  def patch[B >: A](from: Int, other: IterableOnce[B], replaced: Int): CC[B] = fromIterable(new View.Patched(coll, from, other, replaced))
```
julienrf: I noted that you gave a default value of 0 for the `replaced` parameter. What was the reason for that? (I removed it but I can add it back)

odd: No good reason other than that replaced=0 is the safest choice since no elements get omitted, but on second thought keeping it the same way as in the old collections is more important.

src/main/scala/strawman/collection/Iterator.scala
odd: Yes please do that (I can't recall why I didn't just copy that in the first place).

src/main/scala/strawman/collection/immutable/Seq.scala
julienrf: I don’t follow you: `replaced = 0` means that no elements get replaced, so the `patch` operation is a no op, right?

src/main/scala/strawman/collection/Iterator.scala
julienrf: Done.

src/main/scala/strawman/collection/immutable/Seq.scala
odd: No, as I understand it `replaced = 0` means that the other collection is injected into the original collection at index `from` (but it does not replace any elements from the rest of the original collection, whereas `replaced = 3` would mean that the other collection is injected into the original collection at index `from` and the first `replaced` elements from the original are replaced by the other collection (which is always injected in its entirety regardless of `replaced`).

This is also how it works in Scala 2.12.3:
```
scala> List(1, 2, 3).patch(0, List(7, 8, 9), 0)
res0: List[Int] = List(7, 8, 9, 1, 2, 3)

scala> List(1, 2, 3).patch(0, List(7, 8, 9), 2)
res1: List[Int] = List(7, 8, 9, 3)

scala> List(1, 2, 3).patch(2, List(7, 8, 9), 0)
res2: List[Int] = List(1, 2, 7, 8, 9, 3)

scala> List(1, 2, 3).patch(2, List(7, 8, 9), 2)
res3: List[Int] = List(1, 2, 7, 8, 9)
```

julienrf: Oh ok, I understand better now. And I think it makes sense to have a default to 0…

odd: Perhaps `other` should have a default of `Seq.empty`. That would mean that to inject `other` without replacing anything you call it like `patch(index, other)` and to eject a slice you call it like `patch(index, replaced)`. Or we could add `def inject[B >: A](from: Int, other: IterableOnce[B]): CC[B] = patch(from, other, 0)` and `def eject(from: Int, until: Int): C = patch(from, Seq.empty, until)` instead.

julienrf: Note that we already have a `remove` operation that is similar to `eject`.

PR #188, opened 2017-08-09
Add more factory methods

### LINKED ISSUE #169: Add more factory methods
We should add `iterate`, `range`, `tabulate` and `concat` to all factories.

julienrf:
Fixes #169.

### DIFF
--- a/src/main/scala/strawman/collection/Factories.scala
+++ b/src/main/scala/strawman/collection/Factories.scala
@@ -62,7 +62,7 @@ trait IterableFactory[+CC[_]] {
     *  @param   elem the element computation
     *  @return  A $coll that contains the results of `n1 x n2` evaluations of `elem`.
     */
-  def fill[A](n1: Int, n2: Int)(elem: => A): CC[CC[A] @uncheckedVariance] = tabulate(n1)(_ => fill(n2)(elem))
+  def fill[A](n1: Int, n2: Int)(elem: => A): CC[CC[A] @uncheckedVariance] = fill(n1)(fill(n2)(elem))
 
   /** Produces a three-dimensional $coll containing the results of some element computation a number of times.
     *  @param   n1  the number of elements in the 1st dimension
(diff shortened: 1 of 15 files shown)


src/main/scala/strawman/collection/Factories.scala
```diff
@@ -36,14 +38,146 @@ trait CanBuild[-A, +C] extends Any with BuildFrom[Any, A, C] {
...
+    */
   def fill[A](n: Int)(elem: => A): CC[A] = fromIterable(View.Fill(n)(elem))
+
+  /** Produces a two-dimensional $coll containing the results of some element computation a number of times.
+    *  @param   n1  the number of elements in the 1st dimension
+    *  @param   n2  the number of elements in the 2nd dimension
+    *  @param   elem the element computation
+    *  @return  A $coll that contains the results of `n1 x n2` evaluations of `elem`.
+    */
+  def fill[A](n1: Int, n2: Int)(elem: => A): CC[CC[A] @uncheckedVariance] = tabulate(n1)(_ => fill(n2)(elem))
```
Ichoran: Why does this defer to `tabulate` instead of `fill`?

src/main/scala/strawman/collection/View.scala
```diff
@@ -69,18 +69,22 @@ object View extends IterableFactory[View] {
...
-      new Iterator[A] {
-        private var i = 0
-        def hasNext: Boolean = i < n
-        def next(): A = {
-          i = i + 1
-          if (i <= n) elem else Iterator.empty.next()
-        }
-      }
+    def iterator() = Iterator.fill(n)(elem)
     override def knownSize: Int = n
```
Ichoran: This should be fixed to be `0 max n` like the others.

Ichoran:
What's there looks good (with a couple tweaks I suggested).  But tests would help us be more confident!

julienrf:
@Ichoran I applied your suggestions and added some tests.

I also moved the `tabulate` and `fill` factory methods to a new `SeqFactory` trait because these methods are only relevant for `Seq` collections (they make no sense on `Set`, for instance).

Ichoran:
LGTM

PR #189, opened 2017-08-10
Add extractors for head/tail (+:) and init/last (:+).

odd:


julienrf:
Would you mind adding some tests?

PR #190, opened 2017-08-10
Fix mutable sequence hierarchy

julienrf:
I don’t remember why but we had a separate branch of the hierarchy for `Growable` mutable collection, even though all our collections are `Growable`. I made `mutable.Iterable` extend `Growable` and also fixed `ArrayBuffer` and `ListBuffer` hierarchy to include `Shrinkable` (so that now they have a `-=` operation, which was actually missing).

### DIFF
--- a/src/main/scala/strawman/collection/ArrayOps.scala
+++ b/src/main/scala/strawman/collection/ArrayOps.scala
@@ -38,10 +38,24 @@ class ArrayOps[A](val xs: Array[A])
   override def className = "Array"
 
   def iterator(): Iterator[A] = coll.iterator()
+
   def map[B: ClassTag](f: A => B): Array[B] = fromTaggedIterable(View.Map(coll, f))
+
+  def mapInPlace(f: A => A): Array[A] = {
+    var i = 0
+    while (i < xs.length) {
+      xs.update(i, f(xs(i)))
+      i = i + 1
+    }
+    xs
+  }
+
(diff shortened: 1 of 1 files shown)


sjrd:
> even though all our collections are Growable

`WrappedArray` isn't.

julienrf:
Ah yes, maybe that was the reason. Currently we have `ArrayOps`, which extends `IndexedSeq` but not `mutable.IndexedSeq`, that’s why I didn’t spot the problem.

julienrf:
Do we need `WrappedArray`?

sjrd:
`WrappedArray` is what allows an `Array[A]` to be assigned to a `Seq[A]`, so yes, it's pretty fundamental. Possibly more important even, it is what's used by the compiler when calling a varargs method with a fixed number of arguments: it puts the elements in an `Array[A]` and wraps that into a `WrappedArray[A]` to conform to `Seq[A]`.

odd:
FWIW, one potential use for `Spandex` (PR #52) is to replace `WrappedArray` in the varargs-supporting role (perhaps it could also replace it in the array-wrapping role, thus removing the need for `WrappedArray` completely).

julienrf:
We plan to use an `immutable.Seq[A]` collection to carry varargs parameters (`Spandex` or `ImmutableArray`).

It would be nice to be able to assign an `Array[A]` to a `Seq[A]`, though…

PR #193, opened 2017-08-14
Rename benchmarks to better reflect the operations measured.

odd:


benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
```diff
@@ -21,48 +21,121 @@ class HashSetBenchmark {
...
   @Benchmark
-//  @OperationsPerInvocation(size)
-  def cons(bh: Blackhole): Unit = {
+  def incl(bh: Blackhole): Unit = {
     var ys = HashSet.empty[Long]
     var i = 0L
     while (i < size) {
       ys = ys + i
-      i = i + 1
+      i += 1
```
Ichoran: We should be consistent with style.  Now `i` and `ys` are updated with different styles.  Either is fine, as long as they're both that way.

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
```diff
@@ -21,48 +21,121 @@ class HashSetBenchmark {
...
+    }
+  }
+
+  @Benchmark
+  def loop_iterator(bh: Blackhole): Any = {
+    var n = 0
+    val it = xs.iterator()
+    while (it.hasNext) {
+      bh.consume(it.next())
+      n += 1
```
Ichoran: What is the point of `n`?

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
```diff
@@ -21,48 +21,121 @@ class HashSetBenchmark {
...
+    }
+    bh.consume(n)
+  }
+
+  @Benchmark
+  @OperationsPerInvocation(1000)
+  def contains(bh: Blackhole): Unit = {
+    var i = 0
+    while (i < 1000) {
+      bh.consume(xss(i)(i))
```
Ichoran: It would be clearer to actually write this as `contains`.  Why require maintainers to remember that sets are also functions `A => Boolean`?

benchmarks/time/src/main/scala/strawman/collection/immutable/ImmutableArrayBenchmark.scala
```diff
@@ -95,20 +157,65 @@ class ImmutableArrayBenchmark {
...
+  @Benchmark
   def groupBy(bh: Blackhole): Unit = {
     val result = xs.groupBy(_ % 5)
     bh.consume(result)
   }
 
   @Benchmark
   @OperationsPerInvocation(1000)
-  def randomPatch(bh: Blackhole): Unit = {
+  def updated(bh: Blackhole): Unit = {
```
Ichoran: `updated` should be the name for the in-order one, I think.  Random should be something else.  (Updating and indexing should be symmetric.)

Ichoran:
LGTM

PR #194, opened 2017-08-14
Remove typecasting from list.collect

Jacoby6000:


Ichoran:
`List` is supposed to be able to contain `null`.  That's why the code was written the way it was.

Jacoby6000:
Ah right. I always forget about pesky java interop.

Issue #195, opened 2017-08-15
Add TrieMap collection type


julienrf:
We should adapt the code of the current collection’s `TrieMap` to the strawman design.

Ichoran:
Note that the existing (2.12) TrieMap fails to respect custom hashing on hash collisions.  We should make sure the adaptation here fares better.  See https://github.com/scala/bug/issues/10481

julienrf:
I rebased to master and enabled the existing junit tests for `TrieMap`

julienrf:
Closed by #204

Issue #77, opened 2017-05-09
Explore how this library can support Java collections.


Ichoran:
People who are using Scala it interoperate heavily with Java typically need to convert back and forth between Java and Scala collections, e.g. with JavaConverters.  We should write converters for a couple of collections (maybe ArrayList and HashMap?) to shake out any pain points that might exist.

julienrf:
I think if such a thing exists it should be an independent project. But we might want to experiment the feasibility here…

Ichoran:
Right.  I wouldn't want to include it here (most likely--there is an argument to keep it here if we need package private access); I just think we should know how it goes and see if there are any obvious hurdles etc..  I don't anticipate any, but without trying a couple of conversions each way it's a little hard to know.

szeiger:
We currently have these converters in the standard library so they will need to be updated to work with the new collections.

julienrf:
What about providing decorators only rather than converters? (idea suggested by @jvican)

jvican:
Hello. I think providing decorators for Java Map and Set is the way to go and would suppose a significant improvement.
  
As I work on the Zinc API, which has a user-facing Java API implemented by Scala, I've realised that we're doing too many inefficient conversions between Java and Scala that could be solved if we were wrapping Java collections instead of reimplementing them, which causes the VM to allocate memory and takes time. This is usually worst than just one conversion, since we have two conversions: one from Java to Scala and then another one from Scala to Java.
  
I am happy to help on this task.

jvican:
By the way, just a comment on @szeiger's appreciation:
> We currently have these converters in the standard library so they will need to be updated to work with the new collections.
  
Aside from the fact these are expensive, I just want to remark that they are not complete: for example, there's not a conversation between `java.util.Set` to `scala.collection.immutable.Set`.

szeiger:
Where do you see the inefficient conversions? I had a quick look at some of the implementations for both directions and didn't see any copying. All collections are wrapped, and when converting back and forth the wrappers are detected and unwrapped.

julienrf:
Note that @szeiger ported the `JavaConverters` to the strawman in the scala-integration branch: https://github.com/scala/collection-strawman/tree/scala-integration

julienrf:
Closed by #279

PR #200, opened 2017-08-18
Port junit tests to the strawman

julienrf:
This PR is a first step towards porting the current collections tests to the strawman. I ported the JUnit tests found [here](https://github.com/scala/scala/tree/2.13.x/test/junit/scala) (then I’ll work on the scalacheck tests).

To make the review easier I initially copy-pasted the tests without any changes in a [tests-base](https://github.com/scala/collection-strawman/compare/tests-base) branch, then I created a `tests` branch from there, where I incrementally enabled the tests and fixed things in the strawman if needed. The base branch of this PR is the `tests-base` branch, so the diff you will see and comment on are the changes that I had to apply to the tests to make them work with the strawman.

The tests that haven’t been ported are:
 - `TrieMapTest` : we first need to implement `TrieMap` (#195)
 - tests in `scala.collection.convert` and `CollectionConversionsTest` : we have no converters from or to Java collections yet (#77)
 - `PagedSeqTest`, `QueueTest`, `LinkedHashMapTest`, `LinkedHashSetTest`, `MutableListTest`, `OpenHashMapTest`, `PriorityQueueTest`, `UnrolledBufferTest` : we don’t have an equivalent of these collections (should we?)
 - `StringLikeTest` : tests both sequence operations and non-sequence operations (e.g. `toInt`), which are not implemented in the strawman
 - `SearchingTest` : we have no equivalent in the strawman yet.

I will add comments in the diff to explain the changes I’ve made.


test/junit/src/test/scala/strawman/collection/IndexedSeqTest.scala
```diff
@@ -475,25 +476,25 @@ package IndexedTestImpl {
...
+//    import mutable.WrappedArray
+//    override def underTest(size: Int): WrappedArray[BoxedUnit] = {
+//      val res = new Array[Unit](size)
+//      for (i <- 0 until size) {
+//        jlArray.set(res, i, expectedValueAtIndex(i))
+//      }
+//      WrappedArray.make(res.asInstanceOf[Array[Unit]])
+//    }
+//  }
+//  class RefWrappedArrayTest extends WrappedArrayTest[String](classOf[String]) with StringTestData
```
julienrf: I commented these tests because we don’t have `WrappedArray`

test/junit/src/test/scala/strawman/collection/mutable/VectorTest.scala
```diff
@@ -29,20 +30,13 @@ class VectorTest {
...
     val cats = vecs.map(a => vecs.map(a ++ _))
     assert( cats == ans )
   }
 
   @Test
-  def iteratorCat() {
-    def its = vecs.map(_.toList.toIterator)
-    val cats = vecs.map(a => its.map(a ++ _))
-    assert( cats == ans )
-  }
```
julienrf: I removed this one because `concat` now takes an `Iterable`, not an `IterableOnce`.

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -399,6 +399,60 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+  def withFilter(p: A => Boolean): Iterator[A] = filter(p)
+
+  /** Creates an iterator over all the elements of this iterator which
+    *  do not satisfy a predicate p.
+    *
+    *  @param p the predicate used to test values.
+    *  @return  an iterator which produces those values of this iterator which do not satisfy the predicate `p`.
+    *  @note    Reuse: $consumesAndProducesIterator
+    */
+  def filterNot(p: A => Boolean): Iterator[A] = filter(!p(_))
```
Ichoran: This is an inefficient way to do filtering.  The fast way is to have a `filterImpl` that takes a boolean argument and defer to that with opposite sense for `filter` and `filterNot` (and make `filter` and `filterNot` final).  Maybe it should be called `filterWhen` if it's part of the public API.  Not sure this is the right time to make this change, but I can't think of a better time.

src/main/scala/strawman/collection/Map.scala
```diff
@@ -67,20 +68,23 @@ trait MapOps[K, +V, +CC[X, Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]]
...
-    override def size = toIterable.size
-    override def foreach[U](f: K => U) = keysIterator() foreach f
+  protected class KeySet extends Set[K] with GenKeySet {
     def iterableFactory: IterableFactory[Set] = Set
-    def empty: Set[K] = iterableFactory.empty
-    def diff(that: Set[K]): Set[K] = fromSpecificIterable(iterableFactory.fromIterable(toIterable).diff(that))
-    protected[this] def fromSpecificIterable(coll: Iterable[K]): Set[K] = iterableFactory.fromIterable(coll)
+    protected[this] def fromSpecificIterable(coll: Iterable[K]): Set[K] = fromIterable(coll)
     protected[this] def newSpecificBuilder(): Builder[K, Set[K]] = iterableFactory.newBuilder()
+    def diff(that: Set[K]): Set[K] = fromSpecificIterable(this).diff(that)
```
Ichoran: Can you explain how this works and what it's supposed to do?  I'm puzzled.

src/main/scala/strawman/collection/Map.scala
```diff
@@ -67,20 +68,23 @@ trait MapOps[K, +V, +CC[X, Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]]
...
+    def empty: Set[K] = iterableFactory.empty
+  }
+
+  /** A generic trait that is reused by keyset implementations */
+  protected trait GenKeySet extends Serializable { this: Set[K] =>
```
Ichoran: Do you mean `self` instead of `this`?  `this` is confusing.

src/main/scala/strawman/collection/SortedMap.scala
```diff
@@ -22,6 +23,29 @@ trait SortedMapOps[K, +V, +CC[X, Y] <: Map[X, Y] with SortedMapOps[X, Y, CC, _],
...
+  override def keySet: SortedSet[K] = new KeySortedSet
+
+  /** The implementation class of the set returned by `keySet` */
+  protected class KeySortedSet extends SortedSet[K] with GenKeySet with GenKeySortedSet {
+    def iterableFactory: IterableFactory[Set] = Set
+    def sortedIterableFactory: SortedIterableFactory[SortedSet] = SortedSet
+    protected[this] def fromSpecificIterable(coll: Iterable[K]): SortedSet[K] = sortedFromIterable(coll)
+    protected[this] def newSpecificBuilder(): Builder[K, SortedSet[K]] = sortedIterableFactory.newBuilder()
+    protected[this] def sortedFromIterable[B: Ordering](it: Iterable[B]): SortedSet[B] = sortedFromIterable(it)
+    def diff(that: Set[K]): SortedSet[K] = fromSpecificIterable(this).diff(that)
```
Ichoran: Same thing--I don't understand how this works.

src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -36,31 +24,214 @@ class LazyList[+A](expr: => LazyList.Evaluated[A])
...
+        }
       }
-    else "LazyList(?)"
+    that match {
+      case that: LazyList[_] => lazyListEq(this, that)
+      case _ => super.sameElements(that)
+    }
+  }
+
+  /** Apply the given function `f` to each element of this linear sequence
```
Ichoran: This is too much functionality for me to review by eye and have much confidence I've caught a reasonable fraction of any errors.  Hopefully I'll get collections-laws running soon.

src/main/scala/strawman/collection/immutable/NumericRange.scala
```diff
@@ -169,25 +174,25 @@ final class NumericRange[T](
...
-//        def simpleOf(x: Any): String = x.getClass.getName.split("\\.").last
-//        val stepped = simpleOf(underlyingRange.step)
-//        s"${super.toString} (using $underlyingRange of $stepped)"
-//      }
-//    }
-//  }
+  private[immutable] def mapRange[A](fm: T => A)(implicit unum: Integral[A]): NumericRange[A] = {
+    val self = this
+
+    // XXX This may be incomplete.
```
Ichoran: How do we find out if it is?

src/main/scala/strawman/collection/Map.scala
julienrf: Why is `this` confusing? This is a self-type annotation.

src/main/scala/strawman/collection/immutable/NumericRange.scala
julienrf: I don’t know :(

I copied this from the existing code.

src/main/scala/strawman/collection/Map.scala
julienrf: Basically we implement the `diff` operation by delegating to the `IterableFactory` defined above.

`fromSpecificIterable(this)` builds a `Set[K]` by using the `IterableFactory` defined a few lines above. This creates a copy of this keyset. Then we call its `diff` method.

julienrf:
Thanks @Ichoran I know its difficult to review… Maybe I should split this PR in several smaller ones?

src/main/scala/strawman/collection/Map.scala
Ichoran: Oh, I see now that there's no default implementation of `diff` on `Set`; I thought there was, which is why I was so confused.

Still, this doesn't seem very efficient.  Wouldn't something like `fromSpecificIterable(view.filterNot(that))` make more sense, so you don't have to create the set twice?

src/main/scala/strawman/collection/Map.scala
Ichoran: Yes, but if you have a nested class you then have to wonder whether the implicitly defined local `this` shadows the explicitly defined outer `this`.  I guess if there isn't a nested class, it's fine; or we can expect people to know the rules regardless.  Maybe this is a personal style I adopted (don't use `this` to name self-types; use `_` if you don't need to name it, or some other name if you need to use it nested) that isn't widely followed.

Ichoran:
I think the big PR is okay; we just need decent testing at some point.  By the end of next weekend I should have collections-laws running.

src/main/scala/strawman/collection/immutable/NumericRange.scala
Ichoran: I guess we need to add this case into collections-laws testing.  Can you make a note somewhere that mapRange-generated ranges should be laws-tested, in case I forget?

julienrf:
FTR I’m also porting the scalacheck tests (which cover more operations than the junit tests). Do you think we should just switch to collections-laws?

Ichoran:
Collections-laws are fairly comprehensive but do not do a good job probing the corner cases that are appropriate for various different collections types.  The unit tests do the latter well at least where breakages have occurred or been anticipated.  Also, collections-laws were running too slowly to be run every compile. Don't worry too much about tests that are hard to get ported, though.

julienrf:
I do understand the complementary value of unit tests and laws, I was more asking whether the [scalacheck](https://github.com/scala/scala/tree/2.13.x/test/scalacheck) tests are still worth porting (since they may overlap with collections-laws).

Ichoran:
My recollection is that the scalacheck tests also covered some corner cases and also were not perfectly covered by collections-laws (that is, there wasn't really any major distinction between junit and not save for when they were written), but maybe put porting them on hold until we see how collections-laws do?

julienrf:
OK, that works for me :)

julienrf:
I just pushed 2 commits fixing your remarks, @Ichoran.

julienrf:
@Ichoran Is this good to merge for you?

Ichoran:
Sorry--I'm behind as you've noticed.  This was fine!

PR #202, opened 2017-08-18
Simplify benchmarks setup by removing excessive data, also add slice benchmarks.

odd:
In response to the comments below from @julienrf concerning why a fresh collection is built inside some of the individual benchmark methods instead of using `xs` as was done previously, here is an excerpt of a comment I made on this subject in the Spandex (#52) branch: 

_Also, please note that many benchmarks include the creation time of the tested collection since they use a fresh collection each time. This is done to ensure that no unwanted sharing is possible between invocations, but it also skews the benchmarks a bit since collections that are expensive to create but offers cheaper operations will get worse results than the operation would in isolation (to mitigate the effect, those benchmarks use the @OperationsPerInvocation annotation to perform the operation multiple times for each fresh collection)._ 

Another way of putting it is that using the `xs` field means that any structural modifications made to `xs` in one invocation will be shared with the following invocation (potentially leading to skewed results). This can effect the timings for mutable or ostensibly immutable collection implementations (but not fully immutable implementations). 

Perhaps a `create` benchmark should be added to be able to compare the creation times of different collection implementations.

### DIFF
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
@@ -5,7 +5,7 @@ import java.util.concurrent.TimeUnit
 import org.openjdk.jmh.annotations._
 import org.openjdk.jmh.infra.Blackhole
 
-import scala.{Any, AnyRef, Int, Long, Unit}
+import scala.{Any, AnyRef, Int, Long, Unit, math}
 import scala.Predef.intWrapper
 
 @BenchmarkMode(scala.Array(Mode.AverageTime))
@@ -15,55 +15,66 @@ import scala.Predef.intWrapper
 @Measurement(iterations = 12)
 @State(Scope.Benchmark)
 class HashSetBenchmark {
-
-  @Param(scala.Array("0", "1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "73121", "7312102"))
+  @Param(scala.Array(/*"0", */"1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "4096", "131070", "7312102"))
   var size: Int = _
 
   var xs: HashSet[Long] = _
-  var xss: scala.Array[HashSet[Long]] = _
+  var zs: HashSet[Long] = _
   var zipped: HashSet[(Long, Long)] = _
   var randomIndices: scala.Array[Int] = _
-  var randomIndices2: scala.Array[Int] = _
-  var randomXss: scala.Array[HashSet[Long]] = _
+  def fresh(n: Int) = HashSet((1 to n).map(_.toLong): _*)
 
   @Setup(Level.Trial)
-  def initData(): Unit = {
-    def freshCollection() = HashSet((1 to size).map(_.toLong): _*)
-    xs = freshCollection()
-    xss = scala.Array.fill(1000)(freshCollection())
-    zipped = xs.map(x => (x, x))
+  def initTrial(): Unit = {
     if (size > 0) {
       randomIndices = scala.Array.fill(1000)(scala.util.Random.nextInt(size))
-      randomIndices2 = scala.Array.fill(1000)(scala.util.Random.nextInt(size))
-      randomXss = scala.Array.fill(1000)(freshCollection().take(scala.util.Random.nextInt(size)))
+    }
+  }
+
+  @Setup(Level.Invocation)
+  def initInvocation(): Unit = {
+    xs = fresh(size)
+    zs = fresh((size / 1000) max 2).map(-_)
+    zipped = xs.map(x => (x, x))
+  }
+
+  @Benchmark
+  @OperationsPerInvocation(100)
+  def create(bh: Blackhole): Unit = {
+    var i = 0L
+    while (i < 100) {
+      bh.consume(fresh(size))
+      i += 1
     }
   }
 
   @Benchmark
-  def incl(bh: Blackhole): Unit = {
-    var ys = HashSet.empty[Long]
+  @OperationsPerInvocation(1000)
+  def expand_incl(bh: Blackhole): Unit = {
+    var ys = xs
     var i = 0L
-    while (i < size) {
-      ys += i
+    while (i < 1000) {
+      ys += -i
       i += 1
     }
     bh.consume(ys)
   }
 
   @Benchmark
-  def concat(bh: Blackhole): Unit = bh.consume(xs ++ xs)
+  @OperationsPerInvocation(100)
+  def expand_concat(bh: Blackhole): Unit = {
+    var i = 0L
+    while (i < 100) {
+      bh.consume(xs ++ zs)
+      i += 1
+    }
+  }
 
   @Benchmark
-  def tail(bh: Blackhole): Unit = bh.consume(xs.tail)
+  def traverse_foreach(bh: Blackhole): Unit = xs.foreach(x => bh.consume(x))
 
   @Benchmark
-  def init(bh: Blackhole): Unit = bh.consume(xs.init)
-
-  @Benchmark
-  def loop_foreach(bh: Blackhole): Unit = xs.foreach(x => bh.consume(x))
-
-  @Benchmark
-  def loop_headTail(bh: Blackhole): Unit = {
+  def traverse_headTail(bh: Blackhole): Unit = {
     var ys = xs
     while (ys.nonEmpty) {
       bh.consume(ys.head)
@@ -72,7 +83,7 @@ class HashSetBenchmark {
   }
 
   @Benchmark
-  def loop_initLast(bh: Blackhole): Unit = {
+  def traverse_initLast(bh: Blackhole): Unit = {
     var ys = xs
     while (ys.nonEmpty) {
       bh.consume(ys.last)
@@ -81,7 +92,7 @@ class HashSetBenchmark {
   }
 
   @Benchmark
-  def loop_iterator(bh: Blackhole): Unit = {
+  def traverse_iterator(bh: Blackhole): Unit = {
     val it = xs.iterator()
     while (it.hasNext) {
       bh.consume(it.next())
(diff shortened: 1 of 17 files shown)


benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
```diff
@@ -15,43 +15,40 @@ import scala.Predef.intWrapper
...
   @Param(scala.Array(/*"0", */"1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "73121", "7312102"))
   var size: Int = _
 
   var xs: HashSet[Long] = _
-  var xss: scala.Array[HashSet[Long]] = _
   var zipped: HashSet[(Long, Long)] = _
   var randomIndices: scala.Array[Int] = _
-  var randomIndices2: scala.Array[Int] = _
-  var randomXss: scala.Array[HashSet[Long]] = _
+  def fresh(n: Int) = HashSet((1 to size).map(_.toLong): _*)
```
julienrf: Should be `1 to n` instead of `1 to size`

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
```diff
@@ -15,43 +15,40 @@ import scala.Predef.intWrapper
...
-      randomIndices2 = scala.Array.fill(1000)(scala.util.Random.nextInt(size))
-      randomXss = scala.Array.fill(1000)(freshCollection().take(scala.util.Random.nextInt(size)))
     }
   }
 
   @Benchmark
+  @OperationsPerInvocation(1000)
   def incl(bh: Blackhole): Unit = {
-    var ys = HashSet.empty[Long]
+    var ys = fresh(size)
```
julienrf: I think we should not build a new collection within the benchmark. What do @Ichoran or @DarkDimius think?

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
```diff
@@ -15,43 +15,40 @@ import scala.Predef.intWrapper
...
       i += 1
     }
     bh.consume(ys)
   }
 
   @Benchmark
-  def concat(bh: Blackhole): Unit = bh.consume(xs ++ xs)
+  def concat(bh: Blackhole): Unit = {
+    var ys = fresh(size)
+    bh.consume(ys ++ ys)
```
julienrf: What’s the advantage of building `ys` within the benchmark?

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
julienrf: In most cases we could just do `var ys = xs`.

benchmarks/time/src/main/scala/strawman/collection/immutable/ImmutableArrayBenchmark.scala
```diff
@@ -74,19 +96,40 @@ class ImmutableArrayBenchmark {
   }
 
   @Benchmark
-  def prependAll(bh: Blackhole): Unit = bh.consume(xs ++: xs)
+  @OperationsPerInvocation(1000)
+  def prependAll(bh: Blackhole): Unit = {
+    var ys = fresh(size)
+    val zs = fresh((size / 1000) max 1)
```
julienrf: We could initialize `ys` like so `var ys = xs`. For `zs`, I think it’s better to create the collection outside of the benchmark, with the `xss` thing that you deleted.

odd: Besides the problem I commented on above concerning sharing of the `xs` field, the problem with the `xss` array was that it took a really long time to create for the larger benchmark sizes. Perhaps a single `zs` field could be used (but this might also be effected by the sharing problem).

Ichoran:
Building new collections within a benchmark makes it practically impossible to compare anything across benchmarks unless you subtract the time it takes to build (and even then, your accuracy goes down because you have to subtract potentially large numbers from each other).

Unfortunately, I agree with @julienrf that this is not the way to do standard benchmarking.  For targeted cases (e.g. head to head of one algorithm vs another with the same underlying data structure), this is fine, but that isn't how these benchmarks are always being used.  If there are some cases where having a fresh copy is especially important, those can be pulled out into additional "duplicate" (i.e. variant) benchmarks.

odd:
In commit 9245192 the collection creation is moved out of the individual benchmarks and instead placed in the `initInvocation` method which is annotated with `@Setup(Level.Invocation)`. This has the same positive effect of avoiding unwanted sharing of any collection instances between invocations without having to include the collection creation inside the benchmark methods.

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
```diff
@@ -15,55 +15,68 @@ import scala.Predef.intWrapper
...
   @Benchmark
-  def init(bh: Blackhole): Unit = bh.consume(xs.init)
+  @OperationsPerInvocation(100)
+  def expand_concat(bh: Blackhole): Unit = {
+    var ys = xs
+    var i = 0L
+    while (i < 100) {
+      ys = ys ++ zs.map(i * _)
+      i += 1
+    }
```
julienrf: Why do we need to make a loop rather than just calling `xs ++ zs`? Why do we need to `map` `zs` elements?

odd: All elements of `zs` will already be present in `xs` so for `Set` based collections the concat will not add any elements unless `zs` is transformed in some way. I now realize that a better way would be to instead make all elements of `zs` negative at initialization (so that `zs` will never contain any elements already contained in `xs`). 

The looping is strictly not necessary but I added it so that the cost of the benchmark becomes higher compared to the cost of running `initInvocation` before each invocation (where `xs` and `zs` are created) which should improve the quality of the result.

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
```diff
@@ -15,55 +15,67 @@ import scala.Predef.intWrapper
...
+  }
 
   @Benchmark
-  def init(bh: Blackhole): Unit = bh.consume(xs.init)
+  @OperationsPerInvocation(100)
+  def expand_concat(bh: Blackhole): Unit = {
+    val ys = xs
+    var i = 0L
+    while (i < 100) {
+      bh.consume(ys ++ zs)
```
julienrf: Why do we need to create a `val ys = xs` but don’t need it for `zs`?

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
```diff
@@ -15,55 +15,67 @@ import scala.Predef.intWrapper
...
+  @OperationsPerInvocation(1000)
+  def expand_incl(bh: Blackhole): Unit = {
+    var ys = xs
+    var i = 0L
+    while (i < 1000) {
+      ys += -i
+      i += 1
+    }
+    bh.consume(ys)
+  }
```
julienrf: Since we are redesigning the benchmarks I was wondering: de we really need a loop here? Why not the following?

~~~ scala
@Benchmark
def expand_incl(bh: Blackhole): Unit = bh.consume(xs += 42)
~~~

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
odd: We don't, I'll remove `ys` from here (and everywhere else where it is not a `var`).

benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
odd: The reason for using a loop is to make the benchmark method invocation take more time relative to the time taken by the `initInvocation`-method (which avoids the latter overshadowing the former).

julienrf: Are you sure the `@Setup` time is part of the `@Benchmark` time?

odd: No, generally it is not, but as stated in the JavaDoc for `Level.Invocation` (complete text below) benchmark methods taking too little time will skew the result:
>This level is only usable for benchmarks taking more than a millisecond per single benchmark method invocation. It is a good idea to validate the impact for your case on ad-hoc basis as well.
>
>WARNING 1: Since we have to subtract the setup/teardown costs from the benchmark time, on this level, we have to timestamp *each* benchmark invocation. If the benchmarked method is small, then we saturate the system with timestamp requests, which introduce artificial latency, throughput, and scalability bottlenecks.


Complete JavaDoc for `Level.Invocation`:
> Invocation level: to be executed for each benchmark method execution.
>
> WARNING: HERE BE DRAGONS! THIS IS A SHARP TOOL.
> MAKE SURE YOU UNDERSTAND THE REASONING AND THE IMPLICATIONS OF THE WARNINGS BELOW BEFORE EVEN CONSIDERING USING THIS LEVEL.
>
>This level is only usable for benchmarks taking more than a millisecond per single benchmark method invocation. It is a good idea to validate the impact for your case on ad-hoc basis as well.
>
>WARNING 1: Since we have to subtract the setup/teardown costs from the benchmark time, on this level, we have to timestamp *each* benchmark invocation. If the benchmarked method is small, then we saturate the system with timestamp requests, which introduce artificial latency, throughput, and scalability bottlenecks.
>
>WARNING 2: Since we measure individual invocation timings with this level, we probably set ourselves up for (coordinated) omission. That means the hiccups in measurement can be hidden from timing measurement, and can introduce surprising results. For example, when we use timings to understand the benchmark throughput, the omitted timing measurement will result in lower aggregate time, and fictionally *larger* throughput.
>
>WARNING 3: In order to maintain the same sharing behavior as other Levels, we sometimes have to synchronize (arbitrage) the access to state objects. Other levels do this outside the measurement, but at this level, we have to synchronize on *critical path*, further offsetting the measurement.
>
>WARNING 4: Current implementation allows the helper method execution at this Level to overlap with the benchmark invocation itself in order to simplify arbitrage. That matters in multi-threaded benchmarks, when one worker thread executing benchmark method may observe other worker thread already calling `TearDown` for the same object.

julienrf:
Thanks @odd I squashed everything into a single commit and merged it!

PR #204, opened 2017-08-28
Add TrieMap

julienrf:
This is a straightforward port of the existing `TrieMap` collection.


src/main/scala/strawman/collection/concurrent/TrieMap.scala
```diff
@@ -0,0 +1,1146 @@
...
+  final def copy = new TNode(k, v, hc)
+  final def copyTombed = new TNode(k, v, hc)
+  final def copyUntombed = new SNode(k, v, hc)
+  final def kvPair = (k, v)
+  final def cachedSize(ct: AnyRef): Int = 1
+  final def string(lev: Int) = ("  " * lev) + "TNode(%s, %s, %x, !)".format(k, v, hc)
+}
+
+
+private[collection] final class LNode[K, V](final val listmap: immutable.ListMap[K, V])
```
Ichoran: This whole thing is broken because it doesn't respect custom equals.  We need to replace it with a simple `List` that has the appropriate operations but defers to the appropriate `equals`.

src/main/scala/strawman/collection/concurrent/TrieMap.scala
```diff
@@ -0,0 +1,1146 @@
...
+  def inserted(k: K, v: V) = new LNode(listmap + ((k, v)))
+  def removed(k: K, ct: TrieMap[K, V]): MainNode[K, V] = {
+    val updmap = listmap - k
+    if (updmap.size > 1) new LNode(updmap)
+    else {
+      val (k, v) = updmap.iterator().next()
+      new TNode(k, v, ct.computeHash(k)) // create it tombed so that it gets compressed on subsequent accesses
+    }
+  }
+  def get(k: K) = listmap.get(k)
```
Ichoran: This line in particular fails to honor a custom `equal`.

src/main/scala/strawman/collection/concurrent/TrieMap.scala
```diff
@@ -0,0 +1,1146 @@
...
+}
+
+
+private[collection] final class LNode[K, V](final val listmap: immutable.ListMap[K, V])
+  extends MainNode[K, V] {
+  def this(k: K, v: V) = this(immutable.ListMap(k -> v))
+  def this(k1: K, v1: V, k2: K, v2: V) = this(immutable.ListMap(k1 -> v1, k2 -> v2))
+  def inserted(k: K, v: V) = new LNode(listmap + ((k, v)))
+  def removed(k: K, ct: TrieMap[K, V]): MainNode[K, V] = {
+    val updmap = listmap - k
```
Ichoran: This line uses the default equality and thus is wrong.

src/main/scala/strawman/collection/mutable/Map.scala
```diff
@@ -65,6 +65,18 @@ trait MapOps[K, V, +CC[X, Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]]
...
 
+  /** Removes a key from this map, returning the value associated previously
+    *  with that key as an option.
+    *  @param    key the key to be removed
+    *  @return   an option value containing the value associated previously with `key`,
+    *            or `None` if `key` was not defined in the map before.
+    */
+  def remove(key: K): Option[V] = {
+    val r = get(key)
+    this -= key
```
Ichoran: This is not a very efficient way to do this.  You should at least check whether `r` is empty before subtracting `key`, and better is to do it in a single operation.  (That's kind of the point of this method--it's supposed to be the efficient low-level operation in Java maps.)

julienrf: Do you have an idea of how to implement it generically? I think we want to override this default implementation in each specific Map implementation.

julienrf:
Indeed I copy-pasted the current implementation and modified very few things to make it fit into the strawman.

I added a separate commit to fix scala/bug#10481.

julienrf:
I rebased to master and enabled the existing junit tests for `TrieMap`

PR #205, opened 2017-08-28
Push ++: and :++ down from collection.Iterable to collection.Seq and pull +: and :+ up from immutable.Seq to collection.Seq.

odd:


src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -783,60 +783,10 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
       else View.Empty
     }
 
-  /** Alias for `appendAll` */
-  @`inline` final def concat[B >: A](suffix: Iterable[B]): CC[B] = appendAll(suffix)
+  @`inline` final def concat[B >: A](suffix: Iterable[B]): CC[B] = fromIterable(View.Concat(toIterable, suffix))
```
julienrf: Why is it `final` and `@inline`?

src/main/scala/strawman/collection/Seq.scala
```diff
@@ -28,6 +28,126 @@ trait SeqOps[+A, +CC[X], +C] extends Any
...
+  /** Returns a new $coll containing the elements from the left hand operand followed by the elements from the
+    *  right hand operand. The element type of the $coll is the most specific superclass encompassing
+    *  the element types of the two operands.
+    *
+    *  @param suffix the traversable to append.
+    *  @tparam B     the element type of the returned collection.
+    *  @return       a new collection of type `CC[B]` which contains all elements
+    *                of this $coll followed by all elements of `suffix`.
+    */
+  def appendAll[B >: A](suffix: Iterable[B]): CC[B] = concat(suffix)
```
julienrf: I would make this one `@inline` and `final` since it’s an alias for `concat`.

julienrf:
The problem is less important with `Seq` than it is with `Set` or `Map`, because in `immutable.Seq` the addition operations are named `:+` and `+:`, whereas in `mutable.Seq` we only have `+=`, so there is no confusion possible with `:+=` (which would desugar to `xs = xs :+ y`).

In `immutable.Set` and `immutable.Map`, however, we have `+` and `-`, and in `mutable.Set` and `mutable.Map` we have `+=` and `-=`. If `mutable.Set` and `mutable.Map` also had `+` and `-`, then `+=` and `-=` would be confusing: is it the actual `+=` and `-=` operators or the sugar for `xs = xs + y` and `xs = xs - y`?

odd:
I removed the `@inline final` from `concat` as requested. I also made another PR (#206) that pushes `++:` and `:++` down to `immutable.Seq` instead of to `collection.Seq` as this PR does (no need to move `+:` and `:+` in that PR since those are already defined in `immutable.Seq`).

szeiger:
I think `collection.Seq` is the right place for those. I don't know how much will break by removing `++` and `++:` from `Iterable` but all the `:+` and `:+` uses I've seen so far were for `collection.Seq`.

julienrf:
> all the :+ and :+ uses I've seen so far were for collection.Seq.

But if we do `scala.Seq[A] = scala.collection.immutable.Seq[A]` then at least existing code that was using `+:` and `:+` on the default `Seq` will still work if we move these operations to the immutable `Seq`.

szeiger:
You often use it on collection.Seq because you want it to work on all Seqs. Switching the default may work for some cases but I expect the majority will still require these methods (or be reimplemented in possibly very inefficient ways).

Ichoran:
I also think the utility of these methods makes them worth having even for collections typed purely as mutable.  The argument that they're inefficient with mutable collections and you want to discourage their use isn't a very good one, because changing a mutable collection into an immutable one is also not very efficient.  It's a bit of a learning barrier, but for something so central to the use of the language I don't think we should sacrifice much in the way of trained usage for the benefit of learners.

julienrf:
@odersky do you want to weigh in on one side or the other?

julienrf:
Hi @odd, in our last meeting we finally agreed on putting these methods on `collection.Seq` but to rename them to kind of convey an immutable meaning: instead of `append` and `prepend` we would have `appended` and `prepended` (that’s consistent with `Map` which already has `updated`)

odd:
Thanks for the update(d), I suppose the symbolic aliases will still be the same though.

test/junit/src/test/scala/strawman/collection/SeqViewTest.scala
```diff
@@ -11,6 +11,7 @@ class SeqViewTest {
   @Test
   def test_SI8691(): Unit = {
     // Really just testing to make sure ++: doesn't throw an exception
-    assert( (Seq(1,2) ++: Seq(3,4).view).to(Seq) == Seq(1,2,3,4) )
+    assert( (Seq(1,2) :++ Seq(3,4).view).to(Seq) == Seq(1,2,3,4) )
+    assert( (Seq(1,2).view ++: Seq(3,4)).to(Seq) == Seq(1,2,3,4) )
```
julienrf: `SeqView` doesn’t have `++:` anymore?

odd: There is no `SeqView`. The `view` method only ever returns `View` (in `Iterable`),  `IndexedView` (in `IndexedSeqOps`), `ArrayView` (in `ArrayOps`), `StringView` (in `StringOps`),  `ArrayBufferView` (in `ArrayBuffer`) and of course `View`(in `View`).

Before this PR `++:` was available on `Iterable` and thus also on `View`, to make it available again we would need a `SeqView` that extends `SeqOps` and which is returned by the `view` method in `Seq`.

Do you want me to add `SeqView` in this PR (as it is a consequence of moving `prepend` & friends from `Iterable` to `Seq`)?

By the way, I also noticed that `IndexedView` does not extend `IndexedSeqOps` but instead `SeqOps` directly (it does not make much difference however since `IndexedSeqOps` contains no new operations, only overrides for performance, but it does look a bit unregular). Is this intended or should it be changed?

julienrf: Oh, I thought we had `SeqView` :( We should probably add it, so that views of collections that inherit from `Seq` have the same operations.

julienrf: You are right, `IndexedView` should extend `IndexedSeqOps`.

src/main/scala/strawman/collection/View.scala
```diff
@@ -9,14 +9,14 @@ import scala.Predef.{<:<, intWrapper}
...
-  def iterableFactory = View
+  def iterableFactory: IterableFactory[View] = View
 
   protected[this] def fromSpecificIterable(coll: Iterable[A]): View[A] = fromIterable(coll)
 
   protected[this] def newSpecificBuilder(): Builder[A, View[A]] =
     immutable.IndexedSeq.newBuilder().mapResult(_.view)
 
-  final protected[this] def coll: this.type = this
+  protected[this] def coll: this.type = this
```
julienrf: Why isn’t that `final` anymore?

src/main/scala/strawman/collection/Seq.scala
```diff
@@ -12,7 +12,7 @@ import scala.util.hashing.MurmurHash3
 
 /** Base trait for sequence collections */
 trait Seq[+A] extends Iterable[A] with SeqOps[A, Seq, Seq[A]] {
-  final protected[this] def coll: this.type = this
+  protected[this] def coll: this.type = this
```
julienrf: Why remove the `final`?

src/main/scala/strawman/collection/View.scala
```diff
@@ -183,6 +184,12 @@ object View extends IterableFactory[View] {
       else -1
   }
 
+  /** A view that Computes the union between a set and another set.
+   */
+  case class Union[A](prefix: Iterable[A], suffix: Iterable[A]) extends View[A] {
+    def iterator() = prefix.iterator() ++ suffix.iterator()
+  }
```
julienrf: What’s the advantage over `Concat`?

src/main/scala/strawman/collection/View.scala
```diff
@@ -266,22 +264,102 @@ trait ArrayLike[+A] extends Any {
   def apply(i: Int): A
 }
 
+trait SeqView[+A] extends View[A] with Seq[A] with SeqOps[A, SeqView, SeqView[A]] { self =>
```
julienrf: We don’t want to make a diamond with `View` and `Seq`, they really have to stay distinct in the hierarchy. You should just inherit from `View` and `SeqOps`.

src/main/scala/strawman/collection/View.scala
```diff
@@ -33,6 +33,7 @@ object View extends IterableFactory[View] {
   /** Avoid copying if source collection is already a view. */
   def fromIterable[E](it: Iterable[E]): View[E] = it match {
     case it: View[E] => it
+    case it: Iterable[E] => it.view
     case _ => View.fromIterator(it.iterator())
   }
```
julienrf: The wildcard pattern is unreachable.

I think we should just implement it as `it.view` (without doing pattern matching).

src/main/scala/strawman/collection/View.scala
```diff
@@ -266,22 +264,102 @@ trait ArrayLike[+A] extends Any {
...
+  override def view: SeqView[A] = this
+
+  override def iterableFactory: SeqFactory[SeqView] = SeqView
+
+  override protected[this] def fromSpecificIterable(it: Iterable[A]): SeqView[A] =
+    it match {
+      case v: SeqView[A] => v
+      case i: Seq[A] => i.view
+      case _ => it.to(Seq).view
+    }
```
julienrf: You should delegate to `fromIterable`

src/main/scala/strawman/collection/View.scala
```diff
@@ -266,22 +264,102 @@ trait ArrayLike[+A] extends Any {
...
+  case object Empty extends SeqView[Nothing] {
+    override def length = 0
+    def iterator() = Iterator.empty
+    override def knownSize = 0
+  }
+
+  def newBuilder[A](): Builder[A, SeqView[A]] = ArrayBuffer.newBuilder[A]().mapResult(fromIterable)
+
+  /** A view that appends an element to its elements */
+  case class Append[A](underlying: Iterable[A] with ArrayLike[A], elem: A) extends SeqView[A] {
```
julienrf: I guess here we can not write `SeqOps[A, _, _]` because of the bounds on type parameters `CC[_]` and `C`?

src/main/scala/strawman/collection/View.scala
```diff
@@ -266,22 +264,102 @@ trait ArrayLike[+A] extends Any {
...
+  case class PrependAll[A](prefix: Iterable[A], underlying: Iterable[A] with ArrayLike[A]) extends SeqView[A] {
+    override val length = prefix.size + underlying.length
+    def iterator(): Iterator[A] = prefix.iterator() ++ underlying.iterator()
+    override def knownSize: Int = length
+  }
+}
+
 /** View defined in terms of indexing a range */
-trait IndexedView[+A] extends View[A] with ArrayLike[A] with SeqOps[A, View, IndexedView[A]] { self =>
+trait IndexedSeqView[+A] extends SeqView[A] with IndexedSeq[A] with IndexedSeqOps[A, IndexedSeqView, IndexedSeqView[A]] { self =>
```
julienrf: Same here, you should only inherit from `Ops`, not `IndexedSeq`

src/main/scala/strawman/collection/View.scala
odd: It needs to be overridden in `SeqView` and `IndexedSeqView` to provide a more narrow return type.

src/main/scala/strawman/collection/View.scala
odd: Yes, that sounds better, I'll change that.

src/main/scala/strawman/collection/View.scala
odd: One reason is that the known size of a set union is not the same as the size of a concatenation (where it is always the sum of the sizes of the arguments) but for unions this does not hold since the size can be smaller if the same element exists in both arguments (so `knownSize` is not overridden in `Union` but instead inherited to be `-1`).

julienrf: Thanks for the explanation, good catch!

src/main/scala/strawman/collection/View.scala
```diff
@@ -290,22 +287,93 @@ trait ArrayLike[+A] extends Any {
...
+  override def apply(i: Int) = iterator().drop(i - 1).next()
+  override def prepended[B >: A](elem: B): SeqView[B] = SeqView.Prepend(elem, this)
+  override def appended[B >: A](elem: B): SeqView[B] = SeqView.Append(this, elem)
+  override def prependedAll[B >: A](prefix: Iterable[B]): SeqView[B] = SeqView.PrependAll(prefix, this)
+  override def appendedAll[B >: A](suffix: Iterable[B]): SeqView[B] = SeqView.AppendAll(this, suffix)
+}
+
+object SeqView extends SeqFactory[SeqView] {
+  def fromIterator[A](it: => Iterator[A]): SeqView[A] = new SeqView[A] {
+    override def length = it.size
```
julienrf: That looks wrong

odd: Indeed, I guess we could buffer `it` when `size` is called (and use the buffer for any succequent `apply` or `iterator` calls, or we could simply remove `fromIterator` from the `SeqView` companion. Which do you prefer?

odd: I have replaced `SeqView.fromIterator` with `SeqView.fromIteratorProvider` and `IndexedSeqView` with `IndexedSeqView.fromIteratorProvider` in commits 8b3aabb and de8ab24 respectively.

src/main/scala/strawman/collection/View.scala
```diff
@@ -138,12 +138,12 @@ object View extends IterableFactoryLike[View] {
...
      *  that satisfy `p`.
      */
-    val first = Partitioned(this, true)
+    val first = Partitioned(this, cond = true)
 
     /** The view consisting of all elements of the underlying collection
      *  that do not satisfy `p`.
      */
-    val second = Partitioned(this, false)
+    val second = Partitioned(this, cond = false)
```
julienrf: Just a minor remark: in general I would appreciate if this kind of code style improvements were in separate commits. That would make the review process easier.

src/main/scala/strawman/collection/View.scala
```diff
@@ -290,22 +287,107 @@ trait ArrayLike[+A] extends Any {
   def apply(i: Int): A
 }
 
+trait SeqView[+A] extends View[A] with SeqOps[A, SeqView, SeqView[A]] { self =>
+  override def view: SeqView[A] = this
+
+  override def toSeq = immutable.IndexedSeq.from(this)
```
julienrf: In general I prefer omitting the `override` keyword when we only _implement_ a member.

julienrf: I would use the default `Seq` here rather than the default `IndexedSeq`

src/main/scala/strawman/collection/View.scala
```diff
@@ -290,22 +287,107 @@ trait ArrayLike[+A] extends Any {
...
+
+  override def iterableFactory: IterableFactoryLike[SeqView] = SeqView
+
+  override protected[this] def fromSpecificIterable(coll: Iterable[A]): SeqView[A] = fromIterable(coll)
+
+  override protected[this] def newSpecificBuilder(): Builder[A, SeqView[A]] =
+    IndexedSeq.newBuilder().mapResult(_.view)
+
+  override def toString = "SeqView(?)"
```
julienrf: We could implement it once in `View` like so:

~~~ scala
override def toString = s"$className(?)"
~~~

src/main/scala/strawman/collection/View.scala
```diff
@@ -290,22 +287,107 @@ trait ArrayLike[+A] extends Any {
...
+  override def className = "SeqView"
+
+  override def apply(i: Int) = iterator().drop(i - 1).next()
+  override def prepended[B >: A](elem: B): SeqView[B] = SeqView.Prepend(elem, this)
+  override def appended[B >: A](elem: B): SeqView[B] = SeqView.Append(this, elem)
+  override def prependedAll[B >: A](prefix: Iterable[B]): SeqView[B] = SeqView.PrependAll(prefix, this)
+  override def appendedAll[B >: A](suffix: Iterable[B]): SeqView[B] = SeqView.AppendAll(this, suffix)
+}
+
+object SeqView extends SeqFactory[SeqView] {
```
julienrf: Unfortunately in `SeqFactory` the `Source` type is fixed to `IterableOnce`, which is incompatible with views. We should extend `IterableFactoryLike` instead.

I’m wondering if we should make the `Source[A]` type member in `IterableFactoryLike` not lower bounded so that we could fix it to `Seq[A]` here. But we could try that in a different PR.

odd:
That makes sense to me. The work on `SeqView` & Co could then be moved into a separate PR.

julienrf:
@odd I’ve merged the two first commits into master. I had to rebase because I fixed a couple of compilation errors. I’m closing this PR.

PR #206, opened 2017-08-28
Push ++: and :++ down from collection.Iterable to immutable.Seq.

odd:


src/main/scala/strawman/collection/immutable/Seq.scala
```diff
@@ -80,6 +80,55 @@ trait SeqOps[+A, +CC[_], +C] extends collection.SeqOps[A, CC, C] {
...
+  /** Returns a new $coll containing the elements from the left hand operand followed by the elements from the
+    *  right hand operand. The element type of the $coll is the most specific superclass encompassing
+    *  the element types of the two operands.
+    *
+    *  @param suffix the traversable to append.
+    *  @tparam B     the element type of the returned collection.
+    *  @return       a new collection of type `CC[B]` which contains all elements
+    *                of this $coll followed by all elements of `suffix`.
+    */
+  def appendAll[B >: A](suffix: collection.Iterable[B]): CC[B] = fromIterable(View.Concat(toIterable, suffix))
```
julienrf: I would make that one `final` and forward to `concat` so that subclasses override only one method (`concat`) and get the benefit of it regardless that they use `++` or `:++`.

julienrf:
I left a comment about the implementation but first we have to agree on where we put these methods. The fact that it was provided on `Iterable` didn’t make a lot of sense because this collection type does not guarantee elements ordering. That being said that was not hard to implement and might be still useful to some users…

src/main/scala/strawman/collection/immutable/Seq.scala
odd: I pushed df3a9aa which makes `concat` an alias of `appendAll` in `immutable.Seq`. This way any inheritors of `immutable.Seq` can override `prependAll` and `appendAll` if needed for performance (instead of `prependAll` and `concat` which would have been the case if `appendAll` was instead made an alias of `concat`).

szeiger:
I think they belong in `collection.Seq` as implemented in #205, not `immutable.Seq`. This would cause all kinds of breakage.

Ichoran:
I concur (for reasons explained on the other PR).

julienrf:
I’m closing this one because we are going to merge #205 instead.

PR #210, opened 2017-08-29
Handle remaining failing tests.

olafurpg:


olafurpg:
I'm not entirely happy with some parts of the rewrite, I think it will take a few more iterations on the scalafix API to make it nicer. For example, I'd like to upstream SymbolMatcher and Symbol.name to scalafix. This is a rough sketch and does mostly the right things

Issue #163, opened 2017-07-19
Incorrect benchmarks charts


julienrf:
The horizontal scale messes with zero:

![concat](https://user-images.githubusercontent.com/332812/28355418-3271d504-6c64-11e7-9d86-bad7a3b42c5f.png)

marcelocenerine:
I think I have a fix for this. Can I have this issue assigned to me? I will be posting a PR shortly.

There are methods that can throw exceptions when the collection size is == 0 (e.g. tail, head, ...). In that case JMH fails the particular benchmark and does not report measurements. How would you like me to handle that case?

julienrf:
In the case of `concat` (shown above) there should be no problem with empty collections. Are you sure this is the origin of the problem?

marcelocenerine:
No, it is not related. It's just that the measurement for `tail` when size is 0 would be missing in the chart. I just wanted to confirm that this is the  expected outcome (although the answer is likely obvious ;))

julienrf:
Closed by #212

PR #212, opened 2017-08-30
#163 Fix incorrect benchmarks charts

marcelocenerine:
The issue in #163 is that potentially zero values are being plotted in a logarithmic axis (the logarithm of zero is undefined/infinity). The `org.jfree.chart.axis.LogAxis` class does not seem to allow zero/negative values (see impl. of `setSmallestValue`) or accepts them silently and the result is a messed up chart like the one below:

concat_current2

I have thought of a few options to solve this:

1. **remove zeroes:** I don't think we want this as zero is a valid input size for the benchmarks
2. **replace zeroes with a very small positive number (cheating):** the chart is still messed up if the number is too small.
3. **use a linear axis instead:** this one would plot the zeroes. However, if there is a large range of input values (as is the case for HashSetBenchmark) the ones at the beginning would be squeezed (as they are more condensed) and the higher values would be too distant leaving huge gaps in between:
concat_numberaxis

4. **use `LogarithmicAxis`:** this class is similar to `LogAxis` but can allow zero/negative values. I think the result looks pretty good. The only con is that the default base is 10 (the current charts use base 2 in the x-axis) and the api does not let it be changed.

concat_logarithmicaxis
_(p.s.: the execution times for the last 2 sizes are not real. I got OOM errors when running the benchmarks through sbt and did not find out how to set -Xms/-Xmx flags for JMH)_

The last option seemed to be the most appropriate and is the one implemented for this PR. Please let me know if it looks good enough or if there's any other alternative I should consider.

(Note: I did sign the CLA already)

### DIFF
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/HashSetBenchmark.scala
@@ -16,7 +16,7 @@ import scala.Predef.intWrapper
 @State(Scope.Benchmark)
 class HashSetBenchmark {
 
-  @Param(scala.Array(/*"0", */"1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "73121", "7312102"))
+  @Param(scala.Array("0", "1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "73121", "7312102"))
   var size: Int = _
 
   var xs: HashSet[Long] = _
--- a/benchmarks/time/src/main/scala/strawman/collection/immutable/ScalaHashSetBenchmark.scala
+++ b/benchmarks/time/src/main/scala/strawman/collection/immutable/ScalaHashSetBenchmark.scala
@@ -16,7 +16,7 @@ import scala.Predef.intWrapper
 @State(Scope.Benchmark)
 class ScalaHashSetBenchmark {
 
-  @Param(scala.Array(/*"0", */"1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "73121", "7312102"))
+  @Param(scala.Array("0", "1", "2", "3", "4", "7", "8", "15", "16", "17", "39", "282", "73121", "7312102"))
   var size: Int = _
 
   var xs: scala.collection.immutable.HashSet[Long] = _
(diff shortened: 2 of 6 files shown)


Ichoran:
Let's have logarithmic axes both on the horizontal and vertical.  Realistically, no timing should be less than about 0.1 ns.  Anyway, the vertical axis is messed up still.

julienrf:
Thanks for the investigation @marcelocenerine!

For the sake of readability I would either skip values that are less than 10⁻³ or just replace them with 10⁻³. How would the current charts look like if we do that?

marcelocenerine:
@Ichoran I thought that too. I updated the PR and this is how the chart for concat looks like:

concat

@julienrf do you still think I should do what you suggested?

julienrf:
What you did looks super nice and I find these `LogarithmicAxis` more readable than the previous ones :)

I’m surprised to see the error bars going that down, though…

marcelocenerine:
The error bars are actually correct given the input values that were used (see below). The reason for them to go too down is that the distance between the ticks corresponding to the powers of ten is exact the same, even though they are 10x bigger than the preceding ones. This gives this awkward visual perception when the  bars reach numbers close to zero (positive/negative). In order to correctly interpret the information in the chart it's very important to read the tick labels.

```
"size" : "2"
"score" : 63.673358947431716
"scoreError" : 88.77307770042611
"scoreConfidence" : [
    -25.0997187529944,
    152.44643664785784
]


"size" : "16"
"score" : 399.112543226808
"scoreError" : 512.6664719301193
"scoreConfidence" : [
    -113.55392870331133,
    911.7790151569272
]


"size" : "39"
"score" : 1886.1148195021515
"scoreError" : 5565.476584355723
"scoreConfidence" : [
    -3679.361764853572,
    7451.591403857875
]


"size" : "282"
"score" : 14755.610807043742,
"scoreError" : 21855.991922009503,
"scoreConfidence" : [
    -7100.381114965761,
    36611.602729053244
]
```

@julienrf / @Ichoran, thanks for reviewing the PR! Please let me know if you want me to do any other tweak before merging it.

julienrf:
> The reason for them to go too down is that the distance between the ticks corresponding to the powers of ten is exact the same, even though they are 10x bigger than the preceding ones

The error bars do not take into account the fact that the scale is logarithmic?

marcelocenerine:
Oh, sorry. I think I used incorrect input values to generate the chart (maybe because I set the number of benchmark iterations to 3?). The `scoreError`s make no sense. The error bars do take into account the fact that the scale is logarithmic, thus the chart is correct given the input I provided. Here is the chart again but this time running with 12 iterations (I skipped the 73121 and 7312102 sizes as I get OOM errors when running the benchmarks through sbt):

concat

julienrf:
Awsome, thanks!

PR #216, opened 2017-08-31
small size specialized implementations of immutable Set and Map

julienrf:
I copied what was done in the standard library.

I’ll post the benchmark results once it’s finished.

@EPronovost I started from your work but finally didn’t keep it because I needed it for `immutable.Set` (instead of `immutable.HashSet`), I’m sorry for the confusion.

### DIFF
--- a/src/main/scala/strawman/collection/immutable/HashSet.scala
+++ b/src/main/scala/strawman/collection/immutable/HashSet.scala
@@ -88,20 +88,20 @@ object HashSet extends IterableFactory[HashSet] {
 
     protected def get0(elem: Any, hash: Int, level: Int) = false
 
-    protected def updated0(elem: Any, hash: Int, level: Int) = new HashSet1(elem, hash)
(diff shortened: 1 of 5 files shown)


julienrf:
Here are the results of the benchmarks. `HashSetBenchmark` uses the strawman’s `HashSet`, `SetBenchmark` uses the strawman’s `Set` and `ScalaHashSetBenchmark` uses Scala’s `HashSet`. I tested with sizes between 0 and 17. (the size-specialized Sets go to 4 elements max)

We don’t clearly see the advantage of the size-specialized Sets (excepted for `tail` but I doubt this operation is commonly used on `Set`).

incl

loop_iterator

concat

map

tail

julienrf:
Allright, I pushed a small improvements (I avoid calling the `Iterator.apply` varargs factory method because it’s less efficient than building a `List` and getting its `Iterator`) and the results look slightly better for iteration:

loop_iterator

I also ran the `foreach` benchmark (which was missing in my previous comment):

loop_foreach

For me, the improvements (up to 2x faster) are engough to justify this specialization.

PR #217, opened 2017-08-31
Upgrade to Dotty 0.3.0-RC1

smarter:


test/scalacheck/src/test/scala/Ctrie.scala
```diff
@@ -42,19 +41,23 @@ object CtrieTest extends Properties("concurrent.TrieMap") {
       }
     }
 
+    val threads = for (idx <- 0 until totalThreads) yield new ParThread(idx)
```
julienrf: Interesting. Is that a change required by Dotty or just a coding style?

smarter: Structural type support is different in Dotty: http://dotty.epfl.ch/docs/reference/changed/structural-types.html In general it's better to not use them if possible in Scala 2 since the generated code uses Java reflection.

smarter:
> Don’t worry about the junit and scalacheck tests, currently we don’t run them from Dotty. (But that’s nice that you tried to make them more Dotty compatible ^^)

This is require for [Dotty IDE support](http://dotty.epfl.ch/docs/usage/ide-support.html) to work since it will only run when everything compiles.

test/scalacheck/src/test/scala/Ctrie.scala
julienrf: Oh indeed I didn’t notice the use of structural typing

PR #220, opened 2017-09-01
Add cross-compilation with Scala.js 0.6.20.

sjrd:


build.sbt
```diff
@@ -64,8 +71,18 @@ val collections =
       ).toList
     )
 
+val collectionsJVM = collections.jvm
+val collectionsJS = collections.js
+
+val root = project.in(file("."))
```
julienrf: Interesting, we can have two sbt projects in the same file? Well, the `collections` project is not really an sbt project since it’s a cross project, that might be the reason why…

build.sbt
```diff
@@ -64,8 +71,18 @@ val collections =
...
+  .settings(commonSettings ++ disablePublishing)
```
julienrf: So now the new way to publish is to invoke `; collectionsJVM/publishSigned; collectionsJS/publishSigned`? Would it be possible to make a “short-cut” like you did for `test in Test`?

build.sbt
sjrd: Indeed, `collections` is *really not* an sbt project. It's just a `val` like any other.

build.sbt
sjrd: Sure, we can add all the short-cuts you want. (at least for regular tasks; for input tasks it's more complicated)

sjrd: Done.

sjrd:
Rebased after #222

PR #224, opened 2017-09-02
Intersperse

EPronovost:


julienrf:
Why not keep it on IterableOps? We already have mkString there.

EPronovost:
I don't know if it makes sense for Sets or Maps thought.

julienrf:
Yeah actually I agree with you. Sorry for the noise...

src/main/scala/strawman/collection/View.scala
```diff
@@ -86,6 +86,19 @@ object View extends IterableFactory[View] {
...
+    def iterator(): Iterator[B] = new Iterator[B] {
+      val underlyingIterator = underlying.iterator()
+      var intersperseNext = false
+      override def hasNext = intersperseNext || underlyingIterator.hasNext
+      override def next() = {
+        val n = if (intersperseNext) sep else underlyingIterator.next()
+        intersperseNext = !intersperseNext && underlyingIterator.hasNext
+        n
+      }
+    }
```
julienrf: Could you move this implementation to the `Iterator` class (and just call it from here)?

src/main/scala/strawman/collection/Seq.scala
```diff
@@ -99,6 +99,32 @@ trait SeqOps[+A, +CC[X], +C] extends Any
...
+    *
+    * @param start the element to prepend
+    * @param sep the element to intersperse
+    * @param end the element to append
+    * @tparam B the element type of the returned $coll
+    * @return a new collection of type `$Coll` consisting of all elements of this $coll
+    *         interspersed with the element `sep`, beginning with `start` and ending with `end`
+    */
+  def intersperse[B >: A](start: B, sep: B, end: B): CC[B] =
+    fromIterable(View.Append(View.Prepend(start, View.Intersperse(toIterable, sep)), end))
```
julienrf: I’d be curious to know if it is worth writing a specialized method on `Iterator` rather than combining `View`s (from a performance point of view). Could you check that?

src/main/scala/strawman/collection/View.scala
EPronovost: Do you mean to `SeqOps`?

EPronovost:
@julienrf  How do I run the `junit` tests?

src/main/scala/strawman/collection/Seq.scala
EPronovost: How should I test performance?

EPronovost:
The junit test failed.  Should I add `import immutable.List` to the test.  I interpreted the `strawman` package as using our new collections.  If not, then what's the difference between that and the `scala` package?  What's the prior test in the same file doing then?

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -383,6 +383,42 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+    var intersperseNext = false
+    override def hasNext = intersperseNext || self.hasNext
+    override def next() = {
+      val n = if (intersperseNext) sep else self.next()
+      intersperseNext = !intersperseNext && self.hasNext
+      n
+    }
+  }
+  
+  def intersperse[B >: A](start: B, sep: B, end: B): Iterator[B] = new Iterator[B] {
```
EPronovost: @julienrf How can I test the performance of this?

test/junit/src/test/scala/strawman/collection/SeqLikeTest.scala
```diff
@@ -23,4 +23,14 @@ class SeqLikeTest {
     assertEquals(List(List(1, 2), List(1, 3), List(2, 3)), List(1, 2, 3).combinations(2).toList)
     assertEquals(List(List(1, 2, 3)), List(1, 2, 3).combinations(3).toList)
   }
+  
+  @Test def intersperse(): Unit = {
+    import strawman.collection.immutable.List
```
EPronovost: Is it okay to import our own List implementation here for now?

src/main/scala/strawman/collection/Iterator.scala
julienrf: You can have a look at the benchmarks/time project and follow the same pattern. Create two benchmark classes with one benchmark method that has the same name (e.g. `intersperse`). In one class, call the the `intersperse` implementation based on combining `View`s, in the other class, call the `intersperse` implemntation implemented on `Iterator`. Does that make sense? I won’t have time to write more documentation this week.

test/junit/src/test/scala/strawman/collection/SeqLikeTest.scala
julienrf: Yes

EPronovost:
iteratorintersperse
viewintersperse

Big difference for n = 2

julienrf:
Did you also notice that there is a factor 10 between the two results :)

EPronovost:
For n = 2? Yes

EPronovost:
@julienrf is anything else blocking this PR?

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -383,6 +383,42 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
   }
 
   final def size: Int = length
+  
+  def intersperse[B >: A](sep: B): Iterator[B] = new Iterator[B] {
+    var intersperseNext = false
+    override def hasNext = intersperseNext || self.hasNext
+    override def next() = {
+      val n = if (intersperseNext) sep else self.next()
```
julienrf: Could you rename `n` to `elem`, for readability?

src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -383,6 +383,42 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+
+    override def next(): B =
+      if (!started) {
+        started = true
+        start
+      } else if (intersperseNext) {
+        intersperseNext = false
+        sep
+      } else if (self.hasNext) {
+        val n = self.next()
```
julienrf: Same here: I prefer using `elem` as an identifier.

src/main/scala/strawman/collection/Seq.scala
```diff
@@ -99,6 +99,32 @@ trait SeqOps[+A, +CC[X], +C] extends Any
    *          that the resulting collection has a length of at least `len`.
    */
   def padTo[B >: A](len: Int, elem: B): CC[B] = fromIterable(View.PadTo(toIterable, len, elem))
+  
+  /** Adds the element `sep` between each element of the sequence.
+    * e.g. `List(1, 2, 3, 4).intersperse(0) = List(1, 0, 2, 0, 3, 0, 4)`
+    * If the sequence has less then two elements, the collection is unchanged.
```
julienrf: Typo: “less th**a**n”

src/main/scala/strawman/collection/Seq.scala
```diff
@@ -99,6 +99,32 @@ trait SeqOps[+A, +CC[X], +C] extends Any
    *          that the resulting collection has a length of at least `len`.
    */
   def padTo[B >: A](len: Int, elem: B): CC[B] = fromIterable(View.PadTo(toIterable, len, elem))
+  
+  /** Adds the element `sep` between each element of the sequence.
+    * e.g. `List(1, 2, 3, 4).intersperse(0) = List(1, 0, 2, 0, 3, 0, 4)`
```
julienrf: Could you use triple braces?

~~~
{{{
  List(1, 2, 3, 4).intersperse(0) == List(1, 0, 2, 0, 3, 0, 4)
}}}
~~~

src/main/scala/strawman/collection/Seq.scala
```diff
@@ -99,6 +99,32 @@ trait SeqOps[+A, +CC[X], +C] extends Any
...
+    * @param sep the element to intersperse
+    * @tparam B the element type of the returned $coll
+    * @return a new collection of type `$Coll` consisting of all elements of this $coll
+    *         interspersed with the element `sep`
+    */
+  def intersperse[B >: A](sep: B): CC[B] = fromIterable(View.Intersperse(toIterable, sep))
+  
+  /** Adds the element `sep` between each element of the sequence,
+    * prepending `start` and appending `end`.
+    * e.g. `List(1, 2, 3, 4).intersperse(-1, 0, 5) => List(-1, 1, 0, 2, 0, 3, 0, 4, 5)`
```
julienrf: Same here, could you use triple braces for the usage example?

src/main/scala/strawman/collection/Seq.scala
```diff
@@ -99,6 +99,32 @@ trait SeqOps[+A, +CC[X], +C] extends Any
...
+    * If the sequence has less than two elements, returns `start :: this :: end`.
```
julienrf: “returns `start +: this :+ end`”

src/main/scala/strawman/collection/View.scala
```diff
@@ -86,6 +86,20 @@ object View extends IterableFactory[View] {
     def iterator(): Iterator[A] = Iterator.iterate(start)(f).take(len)
     override def knownSize: Int = 0 max len
   }
+  
+  case class Intersperse[A, B >: A](underlying: Iterable[A], sep: B) extends View[B] {
```
julienrf: You should use the type parameter `A` everywhere instead of having both an `A` and a `B`.

src/main/scala/strawman/collection/View.scala
```diff
@@ -86,6 +86,20 @@ object View extends IterableFactory[View] {
...
+    def iterator(): Iterator[B] = underlying.iterator().intersperse(sep)
+    override def knownSize: Int = if (underlying.knownSize > 0) (2 * underlying.knownSize - 1) else underlying.knownSize
+  }
+  
+  case class IntersperseAppend[A, B >: A](underlying: Iterable[A], start: B, sep: B, end: B) extends View[B] {
```
julienrf: I don’t think `IntersperseAppend` is a good name. What do you think of `IntersperseAndSurround`, or just `IntersperseSurround`?

julienrf: Also, you can just use a single type parameter `A` everywhere.

EPronovost:
@julienrf Do those changes work?

julienrf:
@EPronovost sorry for being silent, I need more time before deciding whether we include `intersperse` or not.

julienrf:
@EPronovost we decided to include `intersperse` as a decorator, in `collections-contrib`. Could you move your implementation there?

julienrf:
Thanks @EPronovost, I’ve rebased and moved to the `collections-contrib` project.

Issue #186, opened 2017-08-07
Consider adding a lazyFoldRight method on LazyList


julienrf:

Ichoran:
To make it more grammatical, I'd suggest either `lazyFoldRight` or `foldRightLazily`.  I almost never use `LazyList` so I can't comment on whether this is valuable functionality.

julienrf:
Yes, `lazyFoldRight` looks good. The motivation is that it’s a good way to lazily consume the elements of a `LazyList`, with the ability to short-circuit the termination (so, we can even use it on infinite lists).

EPronovost:
Would this be along the lines of http://voidmainargs.blogspot.com/2011/08/folding-stream-with-scala.html ?

julienrf:
Yes.

EPronovost:
Is someone working on this already?

EPronovost:
How does this look? https://github.com/EPronovost/collection-strawman/commit/924fafc21050c49193817710168e1d04f6e9a1f1

PR #225, opened 2017-09-04
LazyFoldRight

EPronovost:
Per https://github.com/scala/collection-strawman/issues/186

### DIFF
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -3,7 +3,7 @@ package collection
 
 import scala.annotation.unchecked.uncheckedVariance
 import scala.reflect.ClassTag
-import scala.{Any, AnyRef, Array, Boolean, `inline`, Int, None, Numeric, Option, Ordering, PartialFunction, StringContext, Some, Unit}
+import scala.{Any, AnyRef, Array, Boolean, Either, `inline`, Int, None, Numeric, Option, Ordering, PartialFunction, StringContext, Some, Unit}
 import java.lang.{String, UnsupportedOperationException}
 import scala.Predef.<:<
 
@@ -115,10 +115,10 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
   def foldLeft[B](z: B)(op: (B, A) => B): B = toIterable.iterator().foldLeft(z)(op)
 
   /** Fold right */
-  def foldRight[B](z: B)(op: (A, B) => B): B = toIterable.iterator().foldRight(z)(op)
+  def foldRight[B](z: B)(op: (A, B) => B): B = reversed.iterator().foldLeft(z)((b, a) => op(a, b))
   
   /** Lazy fold right */
-  def lazyFoldRight[B](z: B)(op: (A, => B) => B): B = toIterable.iterator().lazyFoldRight(z)(op)
+  def lazyFoldRight[B](z: B)(op: A => Either[B, B => B]): B = toIterable.iterator().lazyFoldRight(z)(op)
 
   /** Reduces the elements of this $coll using the specified associative binary operator.
    *
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -116,6 +116,9 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
 
   /** Fold right */
   def foldRight[B](z: B)(op: (A, B) => B): B = toIterable.iterator().foldRight(z)(op)
+  
+  /** Lazy fold right */
+  def lazyFoldRight[B](z: B)(op: (A, => B) => B): B = toIterable.iterator().lazyFoldRight(z)(op)
 
   /** Reduces the elements of this $coll using the specified associative binary operator.
    *
--- a/src/main/scala/strawman/collection/Iterator.scala
+++ b/src/main/scala/strawman/collection/Iterator.scala
@@ -2,7 +2,7 @@ package strawman.collection
 
 import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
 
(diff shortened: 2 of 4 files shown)


julienrf:
This looks correct but you should add tests. Also, I think it would be interesting to have this method on `IterableOps` instead of only `LazyList`.

We should also make it stack safe (and we should do that for `foldRight` too, but that could be done in a separate PR)

EPronovost:
How would I ensure it's stack safe?  Both `foldRight` and `lazyFoldRight` should be.

Ichoran:
Just include a test that is 100,000 elements long.  If it's not stack-safe, it'll blow the stack and the test will fail.

EPronovost:
So the current implementations are not stack safe.  Looking into similar implementations in cats, it looks like significantly more machinery is required for such an implementation.  Is that the route we want to go?

julienrf:
Hey @EPronovost,

I think we should implement `foldRight` in `Iterator` as follows:

~~~ scala
def foldRight[B](z: B)(op: (A, B) => B): B =
  reversed.foldLeft(z)((b, a) => op(a, b))
~~~

Where `reversed` is copy-pasted from `IterableOps`. At some point we might want to reduce the duplication between `Iterator` and `Iterable`, but that’s a different work :)

We should also re-implement `foldRight` in `IterableOps` like that (to benefit from the fact that `reversed` is overriden in `IndexedSeq` and is O(1)) instead of delegating to `iterator().foldRight…`.

For `lazyFoldRight`, it seems that we indeed have to change its signature in order to implement it in stack-safe way :(

julienrf:
Hi @EPronovost! First, thanks for your patience because we are groping together :) What do you think of having a more meaningful type rather than `Either[B, B => B]`?

What do you think of the following?

~~~ scala
/**
  * A folding step consists either in stopping the traversal (`Break`) or
  * getting the result of folding the tail of the collection (`Pull`).
  */
sealed trait FoldStep[A]

object FoldStep {
  case class Break[A](value: A) extends FoldStep[A]
  case class Pull[A](f: A => A) extends FoldStep[A]
}
~~~

And then:

~~~ scala
def lazyFoldRight[B](z: B)(f: A => FoldStep[B]): B = {
  def foldSteps(x: B, fs: List[B => B]): B = fs.foldLeft(x)((b, f) => f(b))

  def loop(fs: List[B => B]): B =
    if (hasNext) {
      f(next()) match {
        FoldStep.Done(b) => foldSteps(b, fs)
        FoldStep.Pull(f) => loop(f :: fs)
      }
    } else foldSteps(z, fs)

  loop(Nil)
}
~~~

Example of use:

~~~ scala
def forall[A](as: Iterable[A], p: A => Boolean): Boolean =
  as.lazyFoldRight(true)(a => if (p(a)) FoldStep.Break(false) else FoldStep.Pull(b => b))
~~~

EPronovost:
@julienrf That's definitely a more meaningful type usage.  My potential concern is in introducing a whole new trait for this one specific use case.  Is it worth it?

Ichoran:
What about `lazyFoldRight[B](zero: B)(indep: A => Option[B])(accum: (A, B) => B)`?  You then run `indep` until either you get a `Some(b)` or you start with `zero`; and then you do a normal right fold with `accum` on that fragment?  The algorithm is then easy:

```scala
var xs: List[A] = Nil
var z: Option[B] = None
while (hasNext && (z eq None)) {
  val a = next
  z = indep(a)
  if (z.isEmpty) xs = a :: xs
}
xs.foldLeft(z getOrElse zero)(accum)
```

Ichoran:
As an aside, I'm not really convinced that this method is general-purpose enough to include.

You can `xs.takeWhile(dependent).foldRight(zero)(op)` as long as you're dealing with a monoid.  Unless it really is the ability to synthesize of chain of different `B => B`s that is critical, and in that case the usage is even more specialized.

julienrf:
Yeah, it’s right that we can always use `takeWhile` to not consume the whole collection. Do you have examples of things that you can not express with `takeWhile` + `foldRight` but can with `lazyFoldRight`?

EPronovost:
I don't think there's anything you can do with one and not the other, but `takeWhile` + `foldRight` would require two iterations, and thus be less performant.  I can't think of a way to do "foldRight on an unknown-length prefix of the collection" in a single iteration with the existing methods

Ichoran:
Traversing twice is almost always faster than building a list of closures (two mandatory object creations per element).

EPronovost:
@Ichoran @julienrf so then what do we want to do for `lazyFoldRight()`?

julienrf:
On `View` and `LazyList`, `takeWhile` doesn’t build a new collection. So :

~~~ scala
(1 :: 2 :: 3 :: Nil).view.takeWhile(p).foldRight(z)(op)
~~~

and

~~~ scala
(1 :: 2 :: 3 :: Nil).lazyFoldRight(z)(x => if (!p(x)) Break(z2) else Pull(y => op(x, y)))
~~~

Would be equivalent. A key difference between both is that with `lazyFoldRight` you can provide a value for the element that doesn’t verify the predicate `p`. We could probably achieve the same with a `takeUntil` method, though.

Let’s compare the implementation of `forall` with `takeUntil` + `foldRight`, the `lazyFoldRight` signature I proposed and the one proposed by @Ichoran:

~~~ scala
// With takeWhile + foldRight
def forall[A](xs: Iterable[A])(p: A => Boolean): Boolean =
  xs.takeUntil(a => !p(a)).foldRight(true)((a, b) => b && p(a))

// lazyFoldRight
def forall[A](xs: Iterable[A])(p: A => Boolean): Boolean =
  xs.lazyFoldRight(true)(a => if (!p(a)) FoldStep.Break(false) else FoldStep.Pull(b => b))

// Ichoran’s lazyFoldRight
def forall[A](xs: Iterable[A])(p: A => Boolean): Boolean =
  xs.lazyFoldRight(true)(a => if (p(a) None else Some(false))((a, b) => b && p(a))
~~~

I would argue that the `lazyFoldRight` form that I proposed might be preferable because we both compute the `B` and control the traversal at the same time (`p(a)` is evaluated only once). In the alternatives we might have to perform redundant computations.

julienrf:
Superseeded by #295

Issue #107, opened 2017-06-09
FoldWhile


Lasering:
Recently I came across [this code](https://github.com/cvogt/scala-extensions/blob/master/src/main/scala/collection.scala#L139-L167):

```scala
  implicit class GenTraversableOnceExtensions[A](val coll: GenTraversableOnce[A]) extends AnyVal{
    /**
    Fold while accumulation function returns Some. Stops on first None.
    @param initial initual element to start the accumulation on
    @param accumulate accumulation function
    */
    def foldWhile[A1 >: A](initial: A1)
      (accumulate: (A1, A1) => Option[A1]): A1
      = foldLeftWhile[A1](initial)(accumulate)

    /**
    Fold while accumulation function returns Some. Stops on first None.
    @param initial initual element to start the accumulation on
    @param accumulate accumulation function
    */
    def foldLeftWhile[B](initial: B)
      (accumulate: (B, A) => Option[B]): B = {
      @tailrec
      def loop(it: Iterator[A], prev: B): B = {
        if(it.hasNext){
          val next = it.next
          val current = accumulate(prev,next)
          if (current.isEmpty) prev
          else loop(it,current.get)
        } else prev
      }
      loop(coll.toIterator, initial)
    }
  }
```

I'm written multiple variations of an abort early fold like this one.

I think it is a sufficiently common use case with collections that should merit its inclusion the new redesign. Unfortunately I have no data to prove it is in fact a common use case.

julienrf:
Indeed I think that a “breakable” fold could be useful.

EPronovost:
Is it more performant to use `@tailrec` or a while loop on `it.hasNext`?

EPronovost:
https://github.com/scala/collection-strawman/pull/228

PR #228, opened 2017-09-09
Fold left while

EPronovost:


EPronovost:
From https://github.com/scala/collection-strawman/issues/107

Ichoran:
I'm not sure this is the most appropriate signature for a function of this name; everywhere else, `While` always refers to a predicate-based check.  On the other hand, using `Option` provides the greatest functionality.  If we can't think of another equally clear name, then at least the documentation should very clearly note the difference.  (Right now it doesn't have any documentation at all.)

julienrf:
Both `lazyFoldRight` and `foldLeftWhile` have a similar purpose but very different signature and name. I wonder if we can make something more symmetric.

Ichoran:
Maybe `foldSomeLeft` would be a better name for this version?  I'm not very hopeful about unifying the lazy FP potentially-infinite-right-fold and the eager-early-terminated-left-fold concepts.  They are two different ways to do a similar thing, but as you say the signatures are very different (and necessarily so).

julienrf:
`foldSomeLeft` is a good suggestion :)

odd:
How about `foldLeftWhileDefined` or `foldLeftWhileSome`? Or perhaps the other way around `foldLeftUntilEmpty` or `foldLeftUntilNone`?

Ichoran:
I think it is reasonable to introduce `Some` as a generic way to indicate that an option is being used for early termination.  So you could also have `collectSome`, `spanSome`, `mapSome` and so on.

odd:
In that case, I prefer `foldLeftSome` over `foldSomeLeft`.

Ichoran:
@odd - It would make sense if `foldLeft` was a verb, but it doesn't feel very verby, which makes `foldSomeLeft` (with `fold` being the verb) seem more natural.  So, `foldSome`, `foldSomeLeft`, `foldSomeRight` (with directionality being the secondary modifier and `Some` being primary).  But I don't care much either way.

odd:
@Ichoran I'm not a native English speaker so please don't take my naming suggestions too seriously 🤓

EPronovost:
So do we want to implement `foldSome` and `foldSomeRight` as well?

Ichoran:
Yes, I think it would be good to have the various types available.

nafg:
Maybe it would be better to use a special ADT and give it a better name? Like
```scala
xs.foldLeftWhile(start){ (state, x) =>
  if(cond(state)) Continue(state + x) else Break
}
```

EPronovost:
@julienrf What do you think?  This parallels the discussion in #225

julienrf:
In the case of `foldSomeLeft`, the signature isn’t that complicated, I would say that we can just use `Option` instead of a special ADT.

EPronovost:
@julienrf Agreed I think using `Option` is fine for this.  Is this ready to merge then?

odersky:
I'd like us to push the pause button on this. There's a good argument for every single method to be added but if we follow this path we end up in a horrible mess. We need an overall strategy what operations we want and the less the better! Generally, if you don't find immediately a clear name and experiment with many awkward naming attempts like in this thread that's a sign that something is wrong and maybe you should not go ahead in the first place!

I believe for now and in the future, _before_ we add a new method implementation, we need 

 - an extensive documentation what the method does,
 - a clear motivation why we should add this, including where else the method as has been introduced, and what the experiences were. 

I am sorry to be a spoilsport here, but implementing a new method is super easy. Coming up with a minimal set of operations that's powerful and easy to remember is the hard part.

julienrf:
Hi @EPronovost, sorry for the delay. I’ve created a `collections-contrib` sub-project, could you move your method implementation to a decorator on `Iterable`, as I’ve done for `Map` [here](https://github.com/scala/collection-strawman/blob/32fb029af2b00ad464acb04046dc3a6bd403216e/collections-contrib/src/main/scala/strawman/collection/decorators/MapDecorator.scala#L12)? Also, I think we should not have the `foldSomeRight` version which seems confusing to me because even though iteration can be stopped before the whole collection is traversed, it will fail on infinite collections because it needs to `reverse` the collection first. I think that `lazyFoldRight` is the right tool to do a fold with early termination.

danielwegener:
I hope i am not too late to the party but I just came across this addition in 

https://gist.github.com/julienrf/79cc1937530a3b9e26de930794a16042

> foldSomeLeft is a good suggestion :)

I think the important property of "stopping at some point" gets lost in `foldSomeLeft`. At the first glance I would have guessed that `foldSomeLeft` simply skips steps that return a `None` (although that would already be easy to do by simply returning the old accu).
`foldLeftWhile`, `foldLeftWhileSome` or similar captures the property "it is a fold left until a certain point" somewhat better. But it is just a gut feeling and I just wanted to share my first impression.

nafg:
I strongly agree with @danielwegener 

I had made a suggestion to use a different ADT than Option that was rejected, but the point is that the fact that it uses `None | Some a` vs. `Stop | Keep a` or something like that is incidental, an optimization. Putting `Some` in the name doesn't explain what the method is for. It is simply a clue as to how to use it.

I think `foldLeftWhileSome` is the only name that is fully self-explanatory. It's a bit verbose, but I think it's worth it.

EPronovost:
@julienrf in moving it to the decorator, would we still keep the core logic in `Iterator`?

julienrf:
Good point. Indeed we shouldn’t have it in `Iterator` anymore.

We also have to agree on the name. I agree that `foldSomeLeft` is not extremely meaningful. I like `foldLeftWhileDefined` more, but it’s a bit longer…

ShaneDelmore:
Folds that stop remind me of the work done in https://github.com/atnos-org/origami
I would love to have a breakable fold in the stdLib but I am not a fan of foldSome or it’s variations.  Sure, it stops at None but so does unfold and we don’t call it unfoldSome.

nafg:
If we aren't going to have a foldRight version of it then how about just
`foldWhileDefined`?

julienrf:
Superseded by #295

PR #231, opened 2017-09-11
Use custom start/step in 'zipWithIndex'

Dveim:
Motivation -- sometimes I want to use other than 0 start index in `zipWithIndex`. Some alternatives -- implicit class or manual addition, like `for ((x, _i) <- list.zipWithIndex; i = _i + start) ...`. However, probably would be useful to have this functionality built in stdlib, to simplify user code.

Example:
```
scala> List(1, 2, 3).zipWithIndex(start = 42, step = -1)
res0: strawman.collection.immutable.List[(Int, Int)] = List((1,42), (2,41), (3,40))
```

Goals:
1. Already existing syntax `list.zipWithIndex ...` must be preserved, so no user code should be changed.
2. New method must use the same name, so it looks like `list.zipWithIndex(42, 1) ...`.
3. `list.zipWithIndex() ...` must behave exactly the same as `list.zipWithIndex ...`.

To achieve first goal, I had to create explicit overloaded `zipWithIndex` method. That slightly complicates collection's code. Alternative is to provide scalafix rule that replaces `list.zipWithIndex` with `list.zipWithIndex()`. Not sure what's better, decided to start without scalafix.

Any comments and suggestions are welcome.

### DIFF
--- a/src/main/scala/strawman/collection/Iterable.scala
+++ b/src/main/scala/strawman/collection/Iterable.scala
@@ -841,11 +841,14 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
   /** Zips this $coll with its indices.
     *
     *  @return        A new collection of type `That` containing pairs consisting of all elements of this
-    *                 $coll paired with their index. Indices start at `0`.
+    *                 $coll paired with their index. Indices start at `start` and are increased by `step`.
     *  @example
     *    `List("a", "b", "c").zipWithIndex == List(("a", 0), ("b", 1), ("c", 2))`
     */
-  def zipWithIndex: CC[(A @uncheckedVariance, Int)] = fromIterable(View.ZipWithIndex(toIterable))
+  def zipWithIndex(start: Int = 0, step: Int = 0): CC[(A @uncheckedVariance, Int)] = fromIterable(View.ZipWithIndex(toIterable, start, step))
+
+  /** Alias for default zipWithIndex. As a separate method, to support old syntax */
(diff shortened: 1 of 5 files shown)


Dveim:
```
[error] 181 |  override final def zipWithIndex: LazyList[(A, Int)] = zipWithIndex(start = 0, step = 1)
[error]     |                     ^
[error]     |error overriding method zipWithIndex in trait IterableOps of type => strawman.collection.immutable.LazyList[(A @uncheckedVariance, scala.Int)];
[error]     | method zipWithIndex of type => strawman.collection.immutable.LazyList[(A, scala.Int)] cannot override final member method zipWithIndex in trait IterableOps
```

Looks like dotty somehow thinks that alias `zipWithIndex` in `IterableOps` is final. Is it a bug or I missed something?

marcelocenerine:
You can probably achieve this goal with the existing `zip` operation:
```
List(1, 2, 3).zip(LazyList.from(start = 42, step = -1))
```

Dveim:
@marcelocenerine Yes, for `LazyList`. And there many variations with the same result for other collections. But what I propose is unified method with obvious signature.

julienrf:
I agree with @marcelocenerine, in your case I'd rather use zip with a Range or a LazyList.

For me, xs.zipWithIndex should be equivalent to xs.zip(xs.indices).

SethTisue:
I agree with Marcelo and Julien.

julienrf:
@Dveim I’m closing this PR. Thanks for the suggestion but I don’t think the use cases you describe are general enough to be supported in the core collections.

Issue #233, opened 2017-09-12
List.fill performance problem


shawjef3:
#232 shows that List.fill has a severe performance problem for large values of `n`. I suspect other builder functions have the same problem.

```
[info] Benchmark                 (size)  Mode  Cnt           Score            Error  Units
[info] ListBenchmark.fill       7312102  avgt   10  1016141204.700 ± 3015303989.604  ns/op
[info] ScalaListBenchmark.fill  7312102  avgt   10    72150762.767 ±    8988583.569  ns/op
```

EPronovost:
From what I looked at, it appears that the generic `List.fill` falls back to `SeqFactor.fill`, which calls `fromIterable`.  This, in turn, uses a `ListBuilder` and `+= elem` _n_ times.  For Lists, this is a lot of extra baggage.  I wrote an optimized version of `fill` for `List` and was able to increase performance.  Here's the commit: https://github.com/EPronovost/collection-strawman/commit/aa5829efc6628041b5c7df6abf96e502e730ad2b

```
[info] Benchmark                    (size)  Mode  Cnt          Score          Error  Units
[info] FillBenchmark.fillExisting  7312102  avgt   12  131716487.886 ± 36529591.411  ns/op
[info] FillBenchmark.fillNew       7312102  avgt   12   77262934.910 ± 28352202.580  ns/op
[info] FillBenchmark.fillScala     7312102  avgt   12   92711301.221 ± 27794612.476  ns/op
```

shawjef3:
That's a great result. We'll want to use ListBuffer though, so that `fill` appends rather than prepends.

shawjef3:
I'm just learning this code, but it looks like the intended way to optimize this is to use SeqFactory.Delegate. However, an object can't be its own class constructor argument. IE we can't do `object List extends SeqFactory.Delegate[List](this)`. Maybe it's time to consider using a type class instead.

shawjef3:
`fill` uses `ListBuffer.+=`, which should be quite fast. It's how the current collections work, also.

shawjef3:
Attaching VisualVM cpu profiling results for `strawman.collection.immutable.List.fill(7312102)(())`.
![list fill](https://user-images.githubusercontent.com/1625822/30433280-0f266678-9932-11e7-83e3-22d705fbfc24.png)

shawjef3:
[...]

PR #236, opened 2017-09-14
Optimized `List.fill`

EPronovost:
To address https://github.com/scala/collection-strawman/issues/233

Once given the okay, I'll remove `FillBenchmark`


julienrf:
Is `fill` only overriden in `List`’s companion? Do other collections benefit from an optimization?

src/main/scala/strawman/collection/immutable/List.scala
```diff
@@ -364,7 +364,17 @@ object List extends SeqFactory[List] {
   def empty[A]: List[A] = Nil
 
   private[collection] val partialNotApplied = new Function1[Any, Any] { def apply(x: Any): Any = this }
-
+  
+  def fillNew[A](n: Int)(elem: => A): List[A] = {
+    var filled = empty[A]
+    var i = 0
+    while (i < n) {
+      filled = ::(elem, filled)
```
Ichoran: This fills it backwards, which breaks the (at least implied) contract of `fill` for state-dependent operations.

PR #237, opened 2017-09-14
Make factory methods take an IterableOnce instead of an Iterable

### LINKED ISSUE #199: Consider making fromIterable method of factories take an IterableOnce rather than an Iterable
Sometimes we want to build a collection from just an `IterableOnce`. In such a case, we can wrap the `IterableOnce` in a `View` (which is an `Iterable`) to pretend it _is_ an `Iterable` but this is not actually the case and will break if one attempts to perform multiple traversals.

One way to solve this problem would be to change the signature of `fromIterable` as follows:

~~~ diff
- def fromIterable[A](it: Iterable[A]): CC[A]
+ def fromIterable[A](it: IterableOnce[A]): CC[A]
~~~

Thus it would be clear for factory implementations that they should not perform multiple traversals. (they could still check at runtime if the given `IterableOnce` is an instance of `Iterable` if needed).

We might also consider moving the `knownSize` to `IterableOnce` (because one reason to perform several traversals is to compute the size of the collection, which could be avoided in some cases).

As suggested by @szeiger we could also make it more explicit that the `View.fromIterator` takes a thunk as parameter:

~~~ diff
- def fromIterator[A](it: => Iterator[A]): View[A]
+ def fromIteratorProvider[A](it: () => Iterator[A]): View[A]
~~~

julienrf:
Fixes #199.

Changes the signature of factory methods from:

~~~ scala
def fromIterable[A](it: Iterable[A]): CC[A]
~~~

to:

~~~ scala
def from[A](it: IterableOnce[A]): CC[A]
~~~

This makes it simpler to build a collection from an `Iterator`, for instance.

Since this change doesn’t work for `View` (they have to be built from an `Iterable`, not an `IterableOnce`), the source type is still fixed to `Iterable`. To achieve that I introduced an `IterableFactoryLike`, which generalizes `IterableFactory` by making the type of the source collection abstract. This type is then refined to be `IterableOnce` in `IterableFactory` and `Iterable` in `View` companion. I wish I had thought of a better name than `IterableFactoryLike`. Any suggestion?

I also changed the signature of `View.fromIterator[A](it: => Iterator[A]): View[A]` to `View.fromIteratorProvider[A](it: () => Iterator[A]): View[A]` to make it clearer that one should not pass a “provider” that returns the same `Iterator` (because the resulting `View` would be effectively iterable only once, in such a case).

### DIFF
--- a/src/main/scala/strawman/collection/View.scala
+++ b/src/main/scala/strawman/collection/View.scala
@@ -26,14 +26,14 @@ trait View[+A] extends Iterable[A] with IterableOps[A, View, View[A]] {
 /** This object reifies operations on views as case classes */
 object View extends IterableFactory[View] {
 
-  def fromIterator[A](it: => Iterator[A]): View[A] = new View[A] {
-    def iterator() = it
+  def fromIteratorProvider[A](it: () => Iterator[A]): View[A] = new View[A] {
+    def iterator() = it()
   }
 
   /** Avoid copying if source collection is already a view. */
   def from[E](it: IterableOnce[E]): View[E] = it match {
     case it: View[E]     => it
-    case it: Iterable[E] => View.fromIterator(it.iterator())
+    case it: Iterable[E] => View.fromIteratorProvider(() => it.iterator())
     case _ => scala.sys.error("One should not build a View from an IterableOnce instance")
   }
 
--- a/src/main/scala/strawman/collection/View.scala
+++ b/src/main/scala/strawman/collection/View.scala
@@ -2,7 +2,7 @@ package strawman.collection
 
 import strawman.collection.mutable.{ArrayBuffer, Builder}
 
-import scala.{Any, Boolean, Equals, Int, NoSuchElementException, Nothing, annotation, IndexOutOfBoundsException, throws}
+import scala.{Any, Boolean, Equals, IllegalArgumentException, Int, NoSuchElementException, Nothing, annotation, IndexOutOfBoundsException, throws}
 import scala.Predef.{<:<, intWrapper}
 
 /** Concrete collection type: View */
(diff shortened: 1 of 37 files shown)


src/main/scala/strawman/collection/View.scala
```diff
@@ -26,19 +26,20 @@ trait View[+A] extends Iterable[A] with IterableOps[A, View, View[A]] {
...
   }
 
   /** Avoid copying if source collection is already a view. */
-  def fromIterable[E](it: Iterable[E]): View[E] = it match {
-    case it: View[E] => it
-    case _ => View.fromIterator(it.iterator())
+  def from[E](it: IterableOnce[E]): View[E] = it match {
+    case it: View[E]     => it
+    case it: Iterable[E] => View.fromIteratorProvider(() => it.iterator())
+    case _ => scala.sys.error("One should not build a View from an IterableOnce instance")
```
julienrf: The problem that this PR _creates_ is that since `View[A]` extends `Iterable[A]`, it has to have an accompanying `IterableFactory[View]`, which itself has to provide an implementation of `from[E](it: IterableOnce[E]): View[E]`. However, we don’t really want to support building a `View` from an `Iterable`, because the resulting `View` would be effectively iterable only once. Hence, in such a case I throw an exception.

Alternatively, we could just silently copy the `IterableOnce` into an `ArrayBuffer` and call `.view` on it. But I’m not sure it’s better than throwing an exception because we want that building a `View` remains a cheap operation that doesn’t evaluate the underlying collection elements, right?

Ichoran: This should fail at compile-time if at all possible.  I would rather silently copy than throw an exception if compile-time failure is not possible.

julienrf: I wasn't able to find a way to make it fail at compile time if View is an Iterable. The only solution that could work would be to make View something in between IterableOnce and Iterable.

Ichoran: Can we
```scala
  type From[X] <: IterableOnce[X]
  from[E](it: From[E]): CC
```
in the factory, and then `type From[X] = IterableOnce[X]` in everything that's going to make a copy anyway and `type From[X] = Iterable[X]` in everything that refuses to make a copy?

Or does this propagate through other usages too far?

julienrf: This propagates quite a lot but maybe it’s not such a big deal actually… I’ll try again.

julienrf: No actually your trick worked fine with few changes. I just pushed a commit with that change.

julienrf: I updated the description of the PR with last changes (the introduction of the `IterableFactoryLike` generalization).

PR #238, opened 2017-09-14
Fix the comment for `fill`.

shawjef3:
I copied the comment from 2.12.3's documentation, which is accurate for strawman also.


julienrf:
Thanks :)

Issue #239, opened 2017-09-14
Add distinctBy operation


julienrf:
Similar to `maxBy` or `sortBy`.

Related: https://github.com/scala/scala/pull/3850

PR #244, opened 2017-09-14
Add distinctBy operation

marcelocenerine:
Addresses #239:
- Add `distinctBy` operation and re-write `distinct` in terms of `distinctBy`
- Add unit tests and benchmarks for `distinctBy`

### DIFF
--- a/src/main/scala/strawman/collection/Iterator.scala
+++ b/src/main/scala/strawman/collection/Iterator.scala
@@ -3,7 +3,7 @@ package strawman.collection
 import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
 
 import scala.{Any, Array, Boolean, IllegalArgumentException, Int, NoSuchElementException, None, Nothing, Option, PartialFunction, Some, StringContext, Unit, `inline`, math, throws}
-import scala.Predef.{intWrapper, require, String}
+import scala.Predef.{identity, intWrapper, require, String}
 import strawman.collection.mutable.{ArrayBuffer, StringBuilder}
 
 import scala.annotation.tailrec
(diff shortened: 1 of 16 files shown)


src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -469,13 +469,25 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
-    private val traversedValues = mutable.HashSet.empty[A]
+  /**
+    *  Builds a new iterator from this one without any duplicated elements as determined by `==` after applying
+    *  the transforming function `f`.
+    *
+    *  @param f The transforming function whose result is used to determine the uniqueness of each element
+    *  @tparam B the type of the elements after being transformed by `f`
+    *  @return iterator with distinct elements
+    *
+    *  @note   Reuse: $consumesIterator
```
marcelocenerine: does this @note have any relevance?

jozic:
finally this is happening :)

marcelocenerine:
* current impl. of `distinct`:

transform_distinct



* `distinct` implemented in terms of `distinctBy`:

transform_distinctold

julienrf:
@marcelocenerine Do you also have the actual numbers of the benchmarks? It is hard to read them on the images. Alternatively, you could generate one image per collection type (`List`, `Vector`, etc.) in which you would have two lines, one for the current implementation and one for the implementation of this PR.

Also, the title of the second image (“transform_distinctOld”) makes me think that it shows the performance of the current implementation (not the one of this PR), can you double-check?

marcelocenerine:
@julienrf, as you requested I attached a [document](https://github.com/scala/collection-strawman/files/1326568/distinct.benchmarks.pdf) showing the actual numbers as well as individual charts for each collection type. The classes with the "Current" prefix refers to the current implementation of `distinct` whereas the ones without it refers to the implementation of `distinct` in terms of `distinctBy`. The numbers are pretty much identical for all collections types.

julienrf:
Thanks a lot @marcelocenerine!

PR #245, opened 2017-09-15
Replace Level.Invocation with Level.Trial for all benchmarks and remove looping from create benchmark.

odd:


shawjef3:
@odd I assume that tests that alter the state of the collection would be exempt from this. Perhaps mutable and LazyList benchmarks could be put into separate classes for benchmarks that alter the state and those that don't.

odd:
@shawjef3 No, all benchmarks (including mutable and lazy) use the same setup level. This will mean that modifications performed on a mutable collection in one invocation (or even iteration when the level is set to trial) will show through to the next. If this is undesirable (because of result skewing) that benchmark class could either use `Level.Invocation` as was tried before (but that could then once more lead to GC interference problems) or take other measures to avoid the skewing problem (such as resetting the state of the collection via some inexpensive operation if one is available for the collection type being tested).

PR #248, opened 2017-09-21
Add LinkedHashSet and LinkedHashMap

### LINKED ISSUE #211: Provide a LinkedHashMap implementation
The compiler uses `LinkedHashMap` in several places. The most notorious is `Implicits.scala` https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/typechecker/Implicits.scala#L151-L153.
  
Is there any plan to add `LinkedHashMap` to the strawman? Is there a current replacement for it?
  
The difference between a `HashMap` and a `LinkedHashMap` is that the latter preserves order of insertion.

julienrf:
Fixes #211

### DIFF
--- a/src/main/scala/strawman/collection/Map.scala
+++ b/src/main/scala/strawman/collection/Map.scala
@@ -214,6 +214,8 @@ trait MapOps[K, +V, +CC[X, Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]]
 
   override def hashCode(): Int = Set.unorderedHash(toIterable, "Map".##)
 
+  override def toString(): String = super[IterableOps].toString()
+
   override def mkString(start: String, sep: String, end: String): String =
     toIterable.iterator().map { case (k, v) => s"$k -> $v" }.mkString(start, sep, end)
 
--- a/src/main/scala/strawman/collection/mutable/LinkedHashMap.scala
+++ b/src/main/scala/strawman/collection/mutable/LinkedHashMap.scala
@@ -1,0 +1,178 @@
+package strawman
+package collection
+package mutable
+
+import scala.{Unit, None, Option, Serializable, SerialVersionUID, Some, transient}
+
+/** $factoryInfo
+ *  @define Coll `LinkedHashMap`
+ *  @define coll linked hash map
+ */
+object LinkedHashMap extends MapFactory[LinkedHashMap] {
+
+  def empty[A, B] = new LinkedHashMap[A, B]
+
+  def from[K, V](it: collection.IterableOnce[(K, V)]) =
+    it match {
+      case lhm: LinkedHashMap[K, V] => lhm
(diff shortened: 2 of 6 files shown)


src/main/scala/strawman/collection/Map.scala
```diff
@@ -214,6 +214,8 @@ trait MapOps[K, +V, +CC[X, Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]]
 
   override def hashCode(): Int = Set.unorderedHash(toIterable, "Map".##)
 
+  override def toString(): String = super[IterableOps].toString()
```
julienrf: At the same time I fixed `toString` of `Map` collections. Previously, the `toString` implementation of `PartialFunction` was chosen.

src/main/scala/strawman/collection/mutable/LinkedHashMap.scala
```diff
@@ -0,0 +1,178 @@
...
+@SerialVersionUID(1L)
+class LinkedHashMap[K, V]
+  extends Map[K, V]
+    with MapOps[K, V, LinkedHashMap, LinkedHashMap[K, V]]
+    with StrictOptimizedIterableOps[(K, V), Iterable, LinkedHashMap[K, V]]
+    with Serializable {
+
+  type Entry = LinkedHashMap.LinkedEntry[K, V]
+
+  private[this] val table: HashTable[K, V, Entry] =
```
julienrf: One difference with the current standard collections is that they use inheritance whereas we use delegation to manage the hash table.

src/main/scala/strawman/collection/mutable/LinkedHashMap.scala
```diff
@@ -0,0 +1,178 @@
...
+    def next() =
+      if (hasNext) { val res = (cur.key, cur.value); cur = cur.later; res }
+      else Iterator.empty.next()
+  }
+
+  protected class LinkedKeySet extends KeySet {
+    override def iterableFactory: IterableFactory[collection.Set] = LinkedHashSet
+  }
+
+  override def keySet: collection.Set[K] = new LinkedKeySet
```
julienrf: In the current collection the actual `Set` returned by the `keySet` operation of `LinkedHashMap` is a `LinkedHashSet`. I don’t know if this is specified anywhere but I replicated the same behaviour here.

test/junit/src/test/scala/strawman/collection/mutable/LinkedHashMapTest.scala
```diff
@@ -1,15 +1,13 @@
-package scala.collection.mutable
+package strawman.collection.mutable
```
julienrf: I enabled these tests but they cover almost nothing. I’ve had a look at the scalacheck tests but they don’t seem to cover mutable map either…

PR #249, opened 2017-09-21
Separate BuildFrom and CanBuild.

### LINKED ISSUE #198: Build from dynamic type
Removing the `BuildFrom` implicits in #168 (see https://github.com/scala/collection-strawman/pull/168/files#r129011332) breaks building based on the dynamic type of the source collection. The new `CanBuild` implicits are not an adequate replacement for this.

julienrf:
Fixes #198

These changes re-introduce the `BuildFrom` behaviour as it was before #168: the dynamic factory of the source collection is reused to build the target collection.

The PR contains several commits. The [first one](https://github.com/scala/collection-strawman/pull/249/commits/b7195c5704c75f3c5ed26627b3d608e9adbb23b5) contains the essence of the fix and the second one only performs renamings (`CanBuild` is renamed to `Factory`, and `BuildFrom` is moved to a separate file). I suggest to review the first commit before looking at the rest.

Last, I also pushed an alternative implementation [here](https://github.com/julienrf/collection-strawman/commit/d2cfa497fbab6612eceb7b4a32ab089f58fd8227) that makes the conversion between factories and `BuildFrom` instances explicit. I explained the rationale for this change in the commit message. Let me know what you think. [Edit: actually I merged the commit mentioned above in the current branch]

### DIFF
--- a/src/main/scala/strawman/collection/BuildFrom.scala
+++ b/src/main/scala/strawman/collection/BuildFrom.scala
@@ -1,0 +1,50 @@
+package strawman.collection
+
+import scala.{Any, Ordering}
+
+import strawman.collection.mutable.Builder
+
+/** Builds a collection of type `C` from elements of type `A` when a source collection of type `From` is available.
+  * Implicit instances of `BuildFrom` are available for all collection types.
+  *
+  * @tparam From Type of source collection
+  * @tparam A Type of elements (e.g. `Int`, `Boolean`, etc.)
+  * @tparam C Type of collection (e.g. `List[Int]`, `TreeMap[Int, String]`, etc.)
+  */
+trait BuildFrom[-From, -A, +C] extends Any {
+  def fromSpecificIterable(from: From)(it: Iterable[A]): C
+
+  /** Get a Builder for the collection. For non-strict collection types this will use an intermediate buffer.
+    * Building collections with `fromSpecificIterable` is preferred because it can be lazy for lazy collections. */
+  def newBuilder(from: From): Builder[A, C]
+}
+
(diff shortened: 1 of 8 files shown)


szeiger:
I don't like the removal of the implicit conversion. Requiring an extra `.toBuildFrom` IMHO does not make the situation any clearer. On the contrary, I find it more obvious that passing a companion object builds that specific type (rather than something dynamically determined) when you don't have to mention `BuildFrom` by name.

I'm not quite happy with the new names, either. `IterableFactoryLike` vs `IterableFactory` doesn't say anything about the difference. Do we need to make a distinction at all? Also, `Factory` is a bit confusing. The name makes sense on its own (and I wouldn't argue that `CanBuild` is better) but having it alongside types like `IterableFactory` and `SpecificIterableFactory` doesn't make the relationship of these types very clear.

julienrf:
> I'm not quite happy with the new names, either. IterableFactoryLike vs IterableFactory doesn't say anything about the difference

Yeah, I totally agree. The difference between them is only about the type of the `Source` collection that can be used to build a new target collection. In one case (`IterableFactoryLike`) this type is unknown, in the other case (`IterableFactory`), it is `IterableOnce`.

> Also, Factory is a bit confusing. The name makes sense on its own (and I wouldn't argue that CanBuild is better)

`strawman.collection.Factory` does make sense because this type models a general factory for collections. It provides the two basic methods to build a collection (either using a builder or from a source collection). It is also a generalization of `IterableFactory`, `MapFactory`, etc. because all these factories can be modelled as a `Factory` once you fix their element type.

PR #252, opened 2017-09-27
Add collections-contrib sub-project

julienrf:
The collections-contrib project contains decorators introducing new operations on the core collections.

As an example, I added a `zipByKeyWith` operation on `Map` collections.


odersky:
I think it could be released separately, as part of the Scala platform, not the core libraries. Or maybe shepard with the aim to release as a completely independent project. 

My concern is that we need to find a balance to encourage the community to contribute and to face to the hard reality that most of these additions will need an extensive vetting period before they can be added (and a lot of them will likely not be added after all, because we need to keep the footprint small).

Ichoran:
I agree that the cognitive footprint needs to be small.  The actual number of methods doesn't need to be _that_ small since there are already lots, and since Scaladoc is a great tool.  Vetting is useful for three things: usefulness, correctness, and style.  Features that are routinely asked as questions on Stack Overflow or elsewhere can be presumed useful, though; testing can ensure correctness; and that leaves only style, which can partially be taken care of with checking against criteria.

So I don't think we need to invent a new library to get methods that pretty obviously fit into the core functionality.  I do agree that for more complex or esoteric operations, an add-on library can be helpful as an independent project.

PR #255, opened 2017-09-28
Move all the tests into the test/junit sub-project

julienrf:


sjrd:
We should then definitely cross-compile the `junit` project as well. Otherwise `collectionsJS` is completely non-tested (the linker will dce away everything).

PR #258, opened 2017-10-02
Optimize List#updated

### LINKED ISSUE #256: Optimize `updated` on `List`
The result and the receiver should share the suffix of the list.

julienrf:
Don’t recompute the shared tail.

Fixes #256 

Benchmark on call to `updated` with random indices, for various sizes of lists (lower is better):

transform_updaterandom

The `ListUpdatedBenchmark` line (in blue) contains the specialized implementation.


Atry:
`LazyList#update` is still neither optimized nor lazy.

julienrf:
Do you mind opening another issue for that?

PR #259, opened 2017-10-03
Add more join operations

### LINKED ISSUE #71: Map functional combinators
Last year I opened a PR to Scala 2.12 aimed to include a join like operation subset within `Map` operations set: 

https://github.com/scala/scala/pull/5353  

After some discussion it was clear that a more functional interface would fit better with the current map collections interface as `join` methods belong to the foreign (and not so abstract) realm of databases operations.

That was thoroughly syntetized by @szeiger 's  comment, who proposed the following operation set:

```scala
def zipByKeyWith[W, X](that: Map[K, W])(f: (V, W) => X): Map[K, X]
def zipByKey[W](that: Map[K, W]): Map[K, (V, W)] = zipByKeyWith(that)(_ -> _)
```

And, to be able to handle non matching - in both maps - keys:

```scala
def mergeByKeyWith[W, X](that: Map[K, W](f: PartialFunction[(Option[V], Option[W]), X]: Map[K, X]
def mergeByKey[W](that: Map[K, W]): Map[K, (Option[V], Option[W])] = mergeByKeyWith(that)(PartialFunction(identity))
```
`merge*` would then be the way to provide left/right/full-outer join logical operations.

**How could join operations be expressed using this API?**

(**NOTE:**  This is an application example, not part of the requested functionality)

- Inner join:

```scala
def join[K, VA, VB](a: Map[K, VA], b: Map[K, VB]): Map[K, (VA, VB)] =
    a zipByKey b
```
- Left outer join: 

```scala
def leftOuterJoin[K, VA, VB](a: Map[K, VA], b: Map[K, VB]): Map[K, (VA, Option[VB])] =
    a.mergeByKeyWith(b) { case (Some(va), optVb) => va -> optVb }
```

- Right outer join: 

```scala
def rightOuterJoin[K, VA, VB](a: Map[K, VA], b: Map[K, VB]): Map[K, (Option[VA], VB)] = 
    a.mergeByKeyWith(b) { case (optVa, Some(vb)) => optVa -> vb }
```

- Full outer join:

```scala
def fullOuterJoin[K, VA, VB](a: Map[K, VA], b: Map[K, VB]): Map[K, (Option[VA], Option[VB])] = 
    a.mergeByKey(b)
```
I think `mergeByKey*` operations are important as they cover all outer cases. In fact, `zipByKeyWith` implementation could be: 

```scala
def zipByKeyWith[W, X](that: Map[K, W])(f: (V, W) => X): Map[K, X] = 
    mergeByKeyWith(that) { case (Some(va), Some(vb)) => f(va, vb) }
```

But I wouldn't use that implementation since, provided you're zipping only matching keys, you could avoid scrutinising `that` key set.

julienrf:
Add zipByKey/join, mergeByKeyWith, mergeByKey/fullOuterJoin, leftOuterJoin,
rightOuterJoin

Fixes #71

/cc @pfcoperez

### DIFF
--- a/collections-contrib/src/main/scala/strawman/collection/decorators/MapDecorator.scala
+++ b/collections-contrib/src/main/scala/strawman/collection/decorators/MapDecorator.scala
@@ -8,6 +8,15 @@ trait MapDecorator[K, V] {
   /**
     * Combines entries of `this` Map with entries of `that` Map that have the same key,
     * using the combination function `f`
+    *
+    * @param that Other Map
+    * @param f Combination function
+    * @param bf Builder driven by the type of `this` Map
+    * @tparam W Type of values of the other Map (e.g. `Int`, `String`)
+    * @tparam X Type of values of the resulting Map
+    * @tparam C Type of the resulting Map
+    * @return A Map that associates all the keys `k` contained by both `this` and `that` to the result of
+    *         the application of `f` to the values `v` and `w` respectively associated by `this` and `that` to `k`
     */
   def zipByKeyWith[W, X, C](that: Map[K, W])(f: (V, W) => X)(implicit bf: BuildFrom[`this`.type, (K, X), C]): C = {
     val b = bf.newBuilder(`this`)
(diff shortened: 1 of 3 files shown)


collections-contrib/src/main/scala/strawman/collection/decorators/MapDecorator.scala
```diff
@@ -20,4 +29,108 @@ trait MapDecorator[K, V] {
...
+    *         `f`, when this one is defined.
+    *
+    * @param that Map to merge
+    * @param f Combination function
+    * @param bf Builder driven by the type of `this` Map
+    * @tparam W Type of values of the other Map (e.g. `Int`, `String`)
+    * @tparam X Type of values of the resulting Map
+    * @tparam C Type of the result (e.g. `Map[Int, (String, Option[Boolean])]`)
+    */
+  def mergeByKeyWith[W, X, C](that: Map[K, W])(f: PartialFunction[(Option[V], Option[W]), X])(implicit bf: BuildFrom[`this`.type, (K, X), C]): C = {
```
julienrf: I considered introducing a type `EitherOrBoth[A, B]` like so:

~~~ scala
type EitherOrBoth[+A, +B] = Either[(A, B), Either[A, B]]
~~~

But at use site it was less convenient than a pair of options.

LPTK: > it was less convenient

@julienrf How so?

julienrf: First, the goal of `EitherOrBoth` is to not represent meaningless states like `(None, None)`, which could be useful to avoid non exhaustiveness warnings. But in practice we always take as parameter a `PartialFunction`, so exhaustiveness doesn’t matter.

Also, with `(Option[V], Option[W])` we can deal with the cases where `V` is defined in one case: `case (Some(v), _) => …`. Whereas with `EitherOrBoth[V, W]` you would have to write two cases: `case Left((v, _)) => … case Right(Left(v)) => …`.

LPTK: Letting users write `{case (None,None) => ...}` as if it made sense is not great either.

`EitherOrBoth` should have an extractor that looks for a `Left` value whether there is a `Right` one or not: something like `AtLeastLeft(x)`, though the `Left` extractor itself could also have this semantics (would it be too surprising?).

It looks like you have tried encoding `EitherOrBoth[A,B]` as `Either[(A,B),Either[A,B]]`. It's not surprising to me that this was annoying to use, compared to a proper `EitherOrBoth` data type 😉

julienrf: Yeah, good point :)

PR #260, opened 2017-10-03
Add NonEmpty collection type

### LINKED ISSUE #98: Add NonEmpty data structure to the collections
There [exists third-party code that integrates with the previous Scala collections to provide a non empty collection](https://github.com/tarao/nonempty-scala) that proves its contents are non-zero. This is useful in lots of cases and since now only Scalaz has provided this functionality out of the box.
  
I think this may shine on public APIs, providing more type safety. I have just found a use case for this in the sbt API (https://github.com/sbt/sbt/pull/3230) but we cannot use it because it's not in the official collections. We may end up porting the implementation over to sbt/sbt.
  
@julienrf @szeiger What do you think?

julienrf:
Fixes #98.

The approach taken here is similar to @tarao’s approach here: https://github.com/tarao/nonempty-scala

I added a `NonEmpty` wrapper around a collection and make sure that the only way to build an instance of such a wrapper is to provide at least one element.

I provide an implicit conversion from `NonEmpty` to the wrapped instance, making it possible to use all the operations of the specific collection directly on the `NonEmpty` instance, as if it was effectively an instance of that specific collection.

Operations that preserve non-emptiness (e.g. `map`, `++`) are defined directly on `NonEmpty` to return another `NonEmpty` instance.

Current limitations:
 - doesn’t work with `String` or `Array`,
 - doesn’t work with `View` (but that would be easily doable at the cost of slightly more cryptic method signatures).

### DIFF
--- a/collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
+++ b/collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
@@ -19,9 +19,6 @@ final class NonEmpty[+A, C <: Iterable[A]] private (val coll: C)
   def map[B, C2 <: Iterable[B]](f: A => B)(implicit bf: BuildFrom[C, B, C2]): NonEmpty[B, C2] =
     new NonEmpty[B, C2](bf.fromSpecificIterable(coll)(coll.toIterable.map(f)))
 
-  def flatMap[B, C2 <: Iterable[B]](f: A => IterableOnce[B])(implicit bf: BuildFrom[C, B, C2]): NonEmpty[B, C2] =
-    new NonEmpty[B, C2](bf.fromSpecificIterable(coll)(coll.toIterable.flatMap(f)))
-
   def concat[B >: A, C2 <: Iterable[B]](that: Iterable[B])(implicit bf: BuildFrom[C, B, C2]): NonEmpty[B, C2] =
     new NonEmpty[B, C2](bf.fromSpecificIterable(coll)(coll.toIterable ++ that))
 
@@ -84,7 +81,29 @@ object NonEmpty {
     * @tparam A Type of the elements
     * @tparam C Type of the wrapped collection
     */
-  def apply[A, C <: Iterable[A]](elem: A, coll: C)(implicit bf: BuildFrom[C, A, C]): NonEmpty[A, C] =
-    new NonEmpty[A, C](bf.fromSpecificIterable(coll)(View.Prepend(elem, coll.toIterable)))
+  def cons[A, C <: Iterable[A]](elem: A, coll: C)(implicit bf: BuildFrom[C, A, C]): NonEmpty[A, C] =
+    coll match {
+      // Take advantage of `Seq` to not copy the entire tail
+      case seq: Seq[A] => new NonEmpty[A, C]((elem +: seq).asInstanceOf[C])
+      case _ => new NonEmpty[A, C](bf.fromSpecificIterable(coll)(View.Prepend(elem, coll.toIterable)))
+    }
+
+  /**
+    * Example {{{
+    *   val xs = NonEmpty[List[Int]](1, 2, 3)
+    * }}}
+    *
+    * @return A non empty collection of type `C`
+    * @tparam C Type of the non empty collection
+    */
+  def apply[C]: NonEmptyBuilder[C] = new NonEmptyBuilder[C]
+
+  /**
+    * Internal class whose purpose is only to partially apply the type parameter `C`
+    */
+  class NonEmptyBuilder[C] {
+    def apply[A, C1 <: C with Iterable[A]](elem: A, elems: A*)(implicit factory: Factory[A, C1]): NonEmpty[A, C1] =
+      new NonEmpty[A, C1](factory.fromSpecific(elem +: elems.toStrawman))
+  }
 
 }
--- a/collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
+++ b/collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
@@ -1,0 +1,90 @@
+package strawman
+package collection
+
+import collection.immutable.List
+
+/**
+  * A wrapper around a collection that is guaranteed to have at least one element.
+  *
+  * A `NonEmpty` instance is implicitly enriched with the same operations as the
+  * wrapped collection and also has its own operations that preserve non-emptiness.
+  *
+  * @param coll Wrapped collection
+  * @tparam A Type of elements (e.g. `Int`, `String`)
+  * @tparam C Type of collection (e.g. `List[Int]`, `Map[Int, Boolean]`)
+  */
+final class NonEmpty[+A, C <: Iterable[A]] private (val coll: C)
+  extends AnyVal {
+
+  def map[B, C2 <: Iterable[B]](f: A => B)(implicit bf: BuildFrom[C, B, C2]): NonEmpty[B, C2] =
+    new NonEmpty[B, C2](bf.fromSpecificIterable(coll)(coll.toIterable.map(f)))
+
+  def flatMap[B, C2 <: Iterable[B]](f: A => IterableOnce[B])(implicit bf: BuildFrom[C, B, C2]): NonEmpty[B, C2] =
+    new NonEmpty[B, C2](bf.fromSpecificIterable(coll)(coll.toIterable.flatMap(f)))
+
+  def concat[B >: A, C2 <: Iterable[B]](that: Iterable[B])(implicit bf: BuildFrom[C, B, C2]): NonEmpty[B, C2] =
+    new NonEmpty[B, C2](bf.fromSpecificIterable(coll)(coll.toIterable ++ that))
+
+  @`inline` def ++ [B >: A, C2 <: Iterable[B]](that: Iterable[B])(implicit bf: BuildFrom[C, B, C2]): NonEmpty[B, C2] =
+    concat(that)
+
+  def zip[A0 >: A, B, C2 <: Iterable[(A0, B)]](that: NonEmpty[B, _ <: Iterable[B]])(implicit bf: BuildFrom[C, (A0, B), C2]): NonEmpty[(A0, B), C2] =
+    new NonEmpty[(A0, B), C2](bf.fromSpecificIterable(coll)(coll.zip(that.coll)))
+
+  def zipWithIndex[A0 >: A, C2 <: Iterable[(A0, Int)]](implicit bf: BuildFrom[C, (A0, Int), C2]): NonEmpty[(A0, Int), C2] =
+    new NonEmpty[(A0, Int), C2](bf.fromSpecificIterable(coll)(coll.zipWithIndex))
+
+  def prepended[B >: A, C2 <: Seq[B]](elem: B)(implicit bf: BuildFrom[C, B, C2], ev: C <:< Seq[B]): NonEmpty[B, C2] =
+    new NonEmpty[B, C2](bf.fromSpecificIterable(coll)(elem +: coll))
+
+  @`inline` def +: [B >: A, C2 <: Seq[B]](elem: B)(implicit bf: BuildFrom[C, B, C2], ev: C <:< Seq[B]): NonEmpty[B, C2] =
+    prepended(elem)
+
+  @`inline` def :: [B >: A, C2 <: List[B]](elem: B)(implicit bf: BuildFrom[C, B, C2], ev: C <:< List[B]): NonEmpty[B, C2] =
+    prepended(elem)
+
+  def appended[B >: A, C2 <: Iterable[B]](elem: B)(implicit bf: BuildFrom[C, B, C2], ev: C <:< Seq[B]): NonEmpty[B, C2] =
+    new NonEmpty[B, C2](bf.fromSpecificIterable(coll)(coll :+ elem))
+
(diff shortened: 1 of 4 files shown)


tpolecat:
Now that there's a way to represent a non-empty `Iterable` why not remove all the empty-unsafe methods like `reduce`, `head`, `tail`, `min`, etc., and define them on `NonEmpty`?

Ichoran:
@tpolecat - Because that's a gigantic migration hassle for everyone who uses existing collections.

tpolecat:
@Ichoran all such code is a bug anyway so I would think people would be grateful!

julienrf:
It's not that simple : IMHO reduce and max should have returned an Option from the beginning, head and tail are useful for performance reasons. But I do agree that for some methods it would be nice to use NonEmpty (eg groypBy).

Also, the NonEmpty thing provided by this PR is a wrapper around another collection. It's usage is not as easy or straightforward as actual collections. That's why I put it in the collections-contrib artifact only. A consequence of that is that we cannot currently use it in the core collections. We'll see if time tells us that we should move it to the core! (well, this PR must be merged first...)

Ichoran:
@tpolecat - All such code _might_ be a bug, but since Scala doesn't have the kind of context-dependent type refinement that e.g. Kotlin has, and even if it did it might not be hooked up to a logic engine capable of working through all cases, it's not _necessarily_ a bug.  For instance,

```
def firstNumber(xs: Seq[Int]) =
  if (xs.isEmpty) 0 else xs.head
```

has no business _not_ compiling, but we have no mechanism to make it compile, and no alternatives that avoid syntactic and/or runtime overhead (the latter of which can be mitigated by extensive use of macros, but that brings its own problems).

A library that made safety an overriding priority would not do it this way, but the existing collections library is not such a library, and all changes need to be compatible with the existing library modulo small and/or automatic rewrites.

tpolecat:
I would prefer that it not compile. This is exactly the kind of code that becomes problematic as it evolves: if you factor the `else` consequent into a method you now have an unchecked entry condition. It's very easy for blocks to get unmoored from their guards this way.

sjrd:
> and all changes need to be compatible with the existing library modulo small and/or automatic rewrites.

IMO, even more important than this criteria is the following: there must be a way to write code that will cross-compile against 2.12 and 2.13. If that is not the case, the whole library ecosystem breaks down, because no one can afford to drop their existing 2.12 artifacts, so no one can afford to support 2.13.

Therefore, removing `head` from potentially empty collections is a no go, because if you do that I cannot write code that will cross-compile for 2.12 (where `NonEmpty` does not exist) and 2.13 (where `head` does not exist except on `NonEmpty`). And that's whether or not I am willing to write *safe* code.

dwijnand:
> there must be a way to write code that will cross-compile against 2.12 and 2.13

I was hoping that the tooling community would have developed some tools and examples of using Scalafix to fix code "in flight" while cross-compiling. That way we wouldn't need to have the exact same source code. This was was being explored in https://github.com/typelevel/catz-cradle, using a scalaz to cats or a cats to scalaz rewrites, but was abandoned.

Ichoran:
@sjrd - That is already not the case; plenty of things have changed in minor source-incompatible ways.

For example, `.to[]` with a type argument has changed to `.to()` with a companion argument.

Barring some work with Scalafix along the lines of what @dwijnand was saying, cross-compilation will typically not be possible even as currently structured.

ShaneDelmore:
The rewriting situation is easier these days that it was when those experiments were performed.  At the time SBT/scalameta held us back but these days there are proof of concepts using sbt and cbt (for example here is an old one https://github.com/cvogt/cbt/pull/466/files).  Seems everyone who has worked on it, including me, has fallen victim to getting a new job and running out of time to carry it over the finish line but don’t let that discourage anyone.

sjrd:
Even with a good story for on-the-fly rewriting, this is not viable for projects that sit near the roots of the dependency graph in the eco-system. For example, scalafix itself depends on Scala.js and ScalaTest (to name only two big dependencies). This means that neither ScalaTest nor Scala.js can rely on on-the-fly fixing by scalafix, lest we introduce cyclic dependencies (the ultimate nightmare). If those two projects don't cross-compile for 2.12 and 2.13, guess what? virtually none of the ecosystem cross-compiles.

I would probably not trust on-the-fly rewriting anyway. I wouldn't be able to see the diffs it applies on my codebase in code reviews, for example.

`.to[]` versus `.to()` is bad enough, but it might be viable, since it's quite possible to avoid using those methods altogether (provided the common ones like `.toList` are still there). `git grep '\.to\['` says I only use that in tests. It is however not possible to live without `head`.

dwijnand:
I don't think we'd need Scalafix on 2.13 to cross build Scala.js on 2.12 and 2.13. Much like we didn't need sbt on 2.12 (outside of the compiler-bridge) to build the 2.12 ecosystem. Scalafix would be just a build-level dependency.

And for reviewing purposes I can image it would be possible to get CI to show diff of the results of rewriting the before and after code...

julienrf:
@szeiger Thanks for the review!

> I don't like the complexity of this design.

Do you have any suggestion for a simpler design?

> The automatic unwrapping of the underlying collection from a NonEmpty […] blurs the lines between safe and unsafe operations

Yeah, that’s a good point. Should the conversion be explicit?

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
```diff
@@ -0,0 +1,90 @@
...
+  *
+  * @param coll Wrapped collection
+  * @tparam A Type of elements (e.g. `Int`, `String`)
+  * @tparam C Type of collection (e.g. `List[Int]`, `Map[Int, Boolean]`)
+  */
+final class NonEmpty[+A, C <: Iterable[A]] private (val coll: C)
+  extends AnyVal {
+
+  def map[B, C2 <: Iterable[B]](f: A => B)(implicit bf: BuildFrom[C, B, C2]): NonEmpty[B, C2] =
+    new NonEmpty[B, C2](bf.fromSpecificIterable(coll)(coll.toIterable.map(f)))
```
Ichoran: Why does this not just defer directly to the `map` method on `C`?

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  def flatMap[B, C2 <: Iterable[B]](f: A => IterableOnce[B])(implicit bf: BuildFrom[C, B, C2]): NonEmpty[B, C2] =
+    new NonEmpty[B, C2](bf.fromSpecificIterable(coll)(coll.toIterable.flatMap(f)))
```
Ichoran: This is unsafe.  `flatMap` might return no elements.

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
```diff
@@ -0,0 +1,90 @@
...
+    concat(that)
+
+  def zip[A0 >: A, B, C2 <: Iterable[(A0, B)]](that: NonEmpty[B, _ <: Iterable[B]])(implicit bf: BuildFrom[C, (A0, B), C2]): NonEmpty[(A0, B), C2] =
+    new NonEmpty[(A0, B), C2](bf.fromSpecificIterable(coll)(coll.zip(that.coll)))
+
+  def zipWithIndex[A0 >: A, C2 <: Iterable[(A0, Int)]](implicit bf: BuildFrom[C, (A0, Int), C2]): NonEmpty[(A0, Int), C2] =
+    new NonEmpty[(A0, Int), C2](bf.fromSpecificIterable(coll)(coll.zipWithIndex))
+
+  def prepended[B >: A, C2 <: Seq[B]](elem: B)(implicit bf: BuildFrom[C, B, C2], ev: C <:< Seq[B]): NonEmpty[B, C2] =
+    new NonEmpty[B, C2](bf.fromSpecificIterable(coll)(elem +: coll))
```
Ichoran: This is too safe.  Prepending to an existing collection is a way to get a NonEmpty.

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  @`inline` def +: [B >: A, C2 <: Seq[B]](elem: B)(implicit bf: BuildFrom[C, B, C2], ev: C <:< Seq[B]): NonEmpty[B, C2] =
+    prepended(elem)
+
+  @`inline` def :: [B >: A, C2 <: List[B]](elem: B)(implicit bf: BuildFrom[C, B, C2], ev: C <:< List[B]): NonEmpty[B, C2] =
```
Ichoran: This method shouldn't be here.  It's `List` only.

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
Ichoran: (Same comment  about "why not directly" for everything else.)

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
```diff
@@ -0,0 +1,90 @@
...
+}
+
+object NonEmpty {
+
+  import scala.language.implicitConversions
+
+  implicit def toCollection[A, C <: Iterable[A]](nonEmpty: NonEmpty[A, C]): C =
+    nonEmpty.coll
+
+  def fromIterable[A, C <: Iterable[A]](it: C): Option[NonEmpty[A, C]] =
```
Ichoran: `fromIterable` seems like more typing than is warranted.

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
```diff
@@ -0,0 +1,90 @@
...
+    * Create a `NonEmpty` instance that contains `elem` and all the elements of `coll` (which
+    * might be empty).
+    *
+    * If the wrapped collection is a sequence, then `elem` is prepended to `coll`.
+    *
+    * @param bf Builder for the wrapped collection
+    * @tparam A Type of the elements
+    * @tparam C Type of the wrapped collection
+    */
+  def apply[A, C <: Iterable[A]](elem: A, coll: C)(implicit bf: BuildFrom[C, A, C]): NonEmpty[A, C] =
```
Ichoran: That this would be the apply method is unexpected.  I would think `NonEmpty("herring", "cod", "salmon")` would be the most natural usage.  `NonEmpty.prepended(x, xs)` or somesuch would be better for this method.

Ichoran:
I have serious misgivings about this approach.  The forwarding to the underlying collection seems very heavyweight, and having to maintain every reasonable method manually seems like a pain.  Furthermore, it can't detect when you take an existing collection and do something to it that would make it nonempty.

I'm not sure I have a substantially better suggestion (without understanding why you chose the form you did for the forwarding), but I'm not terribly enthusiastic about us spending more time on this without a really clear use case from somewhere explaining how much this implementation solves some problem.

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
julienrf: Oops, indeed!

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
julienrf: Indeed we could omit the `.toIterable` call. That’s a left over of a previous iteration…

julienrf: The `bf.fromSpecificIterable` call is needed because it allows us to compute the resulting collection type `C2` (otherwise we would get an `Iterable[B]`).

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
julienrf: What do you mean?

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
julienrf: Hence the implicit `ev: C <:< List[B]` parameter. The goal is to be able to have one `NonEmpty` class that can wrap any other collection type.

collections-contrib/src/main/scala/strawman/collection/NonEmpty.scala
julienrf: That makes sense.

julienrf:
Thanks the reviews!

I’ve added a commit fixing the issues you reported (I removed `flatMap` and renamed the factory method to `cons`).

I’ve added a better factory method that makes it easier to build a `NonEmpty` collection of any type, as you can see in [the tests](https://github.com/scala/collection-strawman/pull/260/files#diff-bb8618b45c2c17688e3cf448a6bcdbd3R62).

There is still an important issue with equality: currently `NonEmpty` is defined as a value class and a consequence of that is that it is only comparable with other `NonEmpty` instances. I would love to be able to transparently compare a non empty list and a list but that seems a bit hard to achieve. If we make `NonEmpty` a regular class, then we could override the `equals` and `hashCode` to just forward to the implementation of the wrapped collection, but that would only work when the non empty collection is on the left hand side of a comparison…

julienrf:
I’m closing this one because of the problem with equality.

PR #262, opened 2017-10-04
Add mutable.MultiMap

julienrf:
This is an experiment to add a `mutable.MultiMap` collection type. That collection type does already exist in the current collections. I tried to implement it differently.

In the current collections, `MultiMap[K, V]` is simply a trait that you can mix in any `Map[K, Set[V]]` implementation. This trait adds new methods to the underlying `Map`, notably `addBinding(key: K, value: V)` and `removeBinding(key: K, value: V)`. When you use such a `MultiMap` you have to take care of using these new methods to add or remove entries because `+=` and `-=` methods inherited from `Map` still have the `Map` semantics (which is good, BTW!): adding an entry replaces the previous entry, if any, whereas `addBinding` adds the new value to the pre-existing `Set[V]`, if any.

I wanted to find a design where users could still use the usual `+=` and `-=` operators to add or remove elements to a `MultiMap`. As a consequence, I can not inherit from `mutable.Map[K, Set[V]]` because the signature of `+=` would be `def += (key: K, value: Set[V])` instead of the desired `def += (key: K, value: V)`. Therefore, this `MultiMap` implementation only inherit from the generic `collection.Map[K, Set[V]]`.

As you can see in the tests, the API is probably more straightforward.

However, one limitation of this PR is that transformation operations such as `map` or `++` will return a `collection.Map` instead of a `MultiMap`. We could add overloads for those methods:

~~~ scala
def map[V2](f: ((K, V)) => (K, V2)): MultiMap[K, V2]
// etc.
~~~

But I’m not even sure the type system will be able to disambiguate calls with the `map` method inherited from `Map`:

~~~ scala
def map[V2](f: ((K, Set[V])) => (K, V2)): Map[K, V2]
~~~

I think this limitation already exists in the current collections (since I think there is no `CanBuildFrom` instance that builds a `MultiMap` when the type of values is a `Set[V]`).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@ import org.scalajs.sbtplugin.cross.CrossProject
 
 // Convenient setting that allows writing `set scalaVersion := dotty.value` in sbt shell to switch from Scala to Dotty
 val dotty = settingKey[String]("dotty version")
-dotty in ThisBuild := "0.3.0-RC2"
+dotty in ThisBuild := "0.4.0-bin-20171005-4aa13b0-NIGHTLY"
 
 val commonSettings = Seq(
   organization := "ch.epfl.scala",
--- a/collections/src/main/scala/strawman/collection/Map.scala
+++ b/collections/src/main/scala/strawman/collection/Map.scala
@@ -14,7 +14,7 @@ trait Map[K, +V] extends Iterable[(K, V)] with MapOps[K, V, Map, Map[K, V]] {
 }
 
 /** Base Map implementation type */
-trait MapOps[K, +V, +CC[X, Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]]
+trait MapOps[K, +V, +CC[X, Y] <: MapOps[X, Y, CC, _], +C]
   extends IterableOps[(K, V), Iterable, C]
     with PartialFunction[K, V]
     with Equals {
--- a/collections/src/main/scala/strawman/collection/mutable/MultiMap.scala
+++ b/collections/src/main/scala/strawman/collection/mutable/MultiMap.scala
@@ -1,0 +1,124 @@
+package strawman
+package collection.mutable
+
+import scala.{Any, Boolean, None, Option, Some, Unit}
+import strawman.collection.{BuildFrom, Factory, IterableOnce, MapFactory, View}
+
+/** A class for mutable maps with multiple values assigned to a key.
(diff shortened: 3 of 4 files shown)


Ichoran:
I haven't reviewed this in depth yet, but last time I looked deeply into MultiMaps, I thought the same thing: they really need to be their own collection type, a map-like Iterable that is not exactly a Map.  A supertype is possible (for things like `.keys`) but maybe not desirable.
That was in the context of the old collections, however.  I'll re-evaluate in this context in the next day or two.

Ichoran:
I think this is going to be really confusing to use because it's hard to intuit where you can use `K, V` and where you have to use `K, Set[V]`.  Especially in cases where operations available on `Set` are available on `V`, this could lead to all sorts of problems.  Even more confusingly, it's got an immutable interface to `Map[K, Set[V]]` and doesn't have the normal mutable map methods (yet, anyway) like `getOrElseUpdate`.

I like the idea of having a `MultiMap` and I'm surprised how well this one works as designed (e.g. that there aren't catastrophic collisions between Growable and collection.Map), but I still suspect it's too confusing as-is.

I can think of two alternative approaches that may work better.

1. Have a `MultiMap` be its own thing, neither `Seq` or `Map`, but providing access to the underlying `Map[K, Set[V]]`.  In particular, if it was an `Iterable[(K, V)]` it would give you the same key multiple times as it traversed.
2. Don't have a `MultiMap` at all explicitly, but rather enrich a `Map[K, mutable.Set[V]]` with new operations that do the multi-thing.  You could enrich a value class wrapping method `multi` which then allowed syntax like `m.multi += "fish" -> 3`.

julienrf:
Thanks for your reviews.

@Ichoran I think solution 1 is preferable than solution 2. With solution 2, if you forget to call `.multi` and just write `m += "fish" -> 3` then you will loose the previous values.

I think I should make another experiment with `MultiMap` being its own branch of the hierarchy and compare that to this PR. This PR didn’t do that because I wanted to see if it was possible to define `MultiMap` as a very thin layer on top of `Map[K, Set[V]]`.

Ichoran:
If you forget `multi`, you will get a compile time error because `"fish" -> 3` is not a `(String, Set[Int])`, which is the type of the unenriched collection.

That said, I think the interface that can be provided by an "its own thing" implementation will provide superior use cases.

julienrf:
Superseded by #314

PR #263, opened 2017-10-05
Add mutable.MultiSet collection and updateWith operation on mutable Maps

### LINKED ISSUE #133: Add `MultiSet`
And maybe `MultiMap`?

### LINKED ISSUE #251: Compute family of functions for mutable and immutable Maps
Consider adding

```scala
// Possible implementations in terms of public-facing API.

def compute(key: K, f: Option[V] => Option[V]): Map[K, V] = 
  f(this.get(key)) match { 
    case None => this.remove(key)
    case Some(x) => this.updated(key, x)
  }

def computeIfAbsent(key: K, f: => V): Map[K, V] = 
  this.get(key) match {
    case None => this.updated(key, f)
    case Some(_) => this
  }

def computeIfPresent(key: K, f: V => Option[V]): Map[K, V] =
  this.get(key) match {
    case None => this
    case Some(x) => f(x) match {
      case None => this.remove(key)
      case Some(y) => this.updated(key, y)
    }
  }
```

to immutable Maps and similar operations to mutable Maps (see https://docs.oracle.com/javase/8/docs/api/java/util/Map.html).

These patterns occur fairly often in Java (probably 90% of operations on `java.util.Map` in my experience).

julienrf:
Fixes #133 and fixes #251 (cc @alexknvl).

In `collections-contrib`, decorate `mutable.Map` collections with a new operation, `updateWith`, similar to `java.util.Map#compute`: it creates a new binding or updates or removes an existing binding. I felt like we don’t need an equivalent for `computeIfAbsent` because we already have `getOrElseUpdate`, which does exactly the same thing. The case of `computeIfPresent` is also supported by `updateWith(key) { case Some(value) => … }`.

Add a `mutable.MultiSet` collection, whose implementation internally uses a mutable Map and the aforementioned `updateWith` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,12 @@ The following operations are provided:
 - `Map`
     - `zipByKey` / `join` / `zipByKeyWith`
     - `mergeByKey` / `fullOuterJoin` / `mergeByKeyWith` / `leftOuterJoin` / `rightOuterJoin`
+- `mutable.Map`
+    - `updateWith`
+
+The following collections are provided:
+
+- `mutable.MultiSet`
 
 ## Roadmap
 
--- a/collections-contrib/src/main/scala/strawman/collection/decorators/MutableMapDecorator.scala
+++ b/collections-contrib/src/main/scala/strawman/collection/decorators/MutableMapDecorator.scala
@@ -1,0 +1,43 @@
+package strawman
+package collection
+package decorators
+
+class MutableMapDecorator[K, V](`this`: mutable.Map[K, V]) {
+
+  /**
+    * Updates an existing binding or create a new one according to the
+    * result of the application of `f`.
+    *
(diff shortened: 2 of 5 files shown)


Ichoran:
These both look fairly reasonable, but I think the two changes should be two separate PRs.  `updateWith` is a really obvious missing feature.  A counted set is nice, but it's less clear we need it.  In particular, I'm really uncertain whether it should be an `Iterable[(K, Int)]` or an `Iterable[K]` that you can get at least an immutable interface to a `Map[K, Int]` from.  Either way, the strategy should match the `MultiMap` strategy: if `MultiMap[K, V]` is an `Iterable[(K, V)]` then `MultiSet` should be an `Iterable[K]`; if the groupedness is visible in the `Iterable` then it should be in both (for consistency).

Issue #265, opened 2017-10-14
NumericRange.iterator() throws java.util.NoSuchElementException when empty


marcelocenerine:
```scala
@RunWith(classOf[JUnit4])
class NumericRangeTest {

  @Test
  def iteratorFromEmptyNumericRange(): Unit = {
    assertFalse(NumericRange(start = 0, end = 0, step = 1).iterator().hasNext) 
    // java.util.NoSuchElementException: head of empty list
  }

  @Test
  def iteratorFromEmptyRange(): Unit = {
    assertFalse(Range(0, 0, 1).iterator().hasNext) // OK
  }

  @Test
  def iteratorFromEmptyScalaRange(): Unit = {
    assertFalse((0 until 0 by 1).iterator.hasNext) // OK
  }
}
```

PR #266, opened 2017-10-14
Fix empty iterator in NumericRange

marcelocenerine:
Fix for #265

### DIFF
--- a/collections/src/main/scala/strawman/collection/immutable/NumericRange.scala
+++ b/collections/src/main/scala/strawman/collection/immutable/NumericRange.scala
@@ -41,9 +41,21 @@ sealed class NumericRange[T](
   extends IndexedSeq[T]
     with IndexedSeqOps[T, IndexedSeq, IndexedSeq[T]]
     with StrictOptimizedSeqOps[T, IndexedSeq, IndexedSeq[T]]
-    with Serializable {
+    with Serializable { self =>
 
-  override def iterator(): Iterator[T] = new NumericRangeIterator[T](start, step, last, isEmpty)
+  override def iterator() = new Iterator[T] {
+    private var _hasNext = !self.isEmpty
+    private var _next: T = start
+    private val lastElement: T = if (_hasNext) last else start
+    def hasNext: Boolean = _hasNext
+    def next(): T = {
+      if (!_hasNext) Iterator.empty.next()
+      val value = _next
+      _hasNext = value != lastElement
+      _next = num.plus(value, step)
+      value
+    }
+  }
 
   def iterableFactory: SeqFactory[IndexedSeq] = IndexedSeq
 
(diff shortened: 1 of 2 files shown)


Ichoran:
I don't understand what the problem was with the old logic.  Can you figure it out?  The new approach seems like it would perform more poorly, and if the old approach didn't work, there was probably something else wrong which is still wrong and is a bug.

In particular, the iterator has to detect _anyway_ when it becomes empty, so if it starts off empty it should be able to start in that state, not have to return an entirely different iterator type (i.e. Iterator.empty).  Yes, there's an advantage in memory allocation from returning `Iterator.empty`, but it's probably more than offset by needing multiple dispatch on the iterator.

marcelocenerine:
@Ichoran, the problem was that the old logic calls `last` when instantiating `NumericRangeIterator`:

```scala
override def iterator(): Iterator[T] = new NumericRangeIterator[T](start, step, last, isEmpty)

override def last: T =
  if (length == 0) Nil.head
  else locationAfterN(length - 1)
```

Another option would be to turn `lastElement` a by-name parameter of the `NumericRangeIterator` constructor. It is accompanied by the `initiallyEmpty` flag which is checked first and would avoid the exception.

Ichoran:
What about `if (isEmpty) NumericRangeIterator(start, step, start, false) else NumericRangeIterator(start, step, last, true)` or something like that?

marcelocenerine:
@Ichoran, that would work too. However, I'm not seeing much advantage compared to the 1st approach, given that with your suggestion both `iterator()` and `NumericRangeIterator(...)` would have to concern about the initial empty state rather than just one.
The 1st solution does not perform more poorly given that the old approach calls `Nil.head`, which is the source of the problem. In the 2nd solution (by-name parameter) there wouldn't be an additional dispatch given that the function wrapping the parameter would never be called (although it needs to be created). This shouldn't be a big deal as it only happens when the iterator is created. Both 1 & 2 look more readable too.
I'm OK with your approach too. Just wanted to share my thoughts.

marcelocenerine:
Alternatively....

```scala
  self =>

   override def iterator() = new Iterator[T] {
    private var _hasNext = !self.isEmpty
    private var _next: T = start
    private val lastElement: T = if (_hasNext) last else start
    def hasNext: Boolean = _hasNext
    def next(): T = {
      if (!_hasNext) Iterator.empty.next()
      val value = _next
      _hasNext = value != lastElement
      _next = num.plus(value, step)
      value
    }
  }
```

marcelocenerine:
@Ichoran I changed to the approach above which is similar to what you proposed, but having the emptiness dealt with in one single place.

Ichoran:
LGTM, thanks @marcelocenerine !

PR #269, opened 2017-10-18
Add MultiMap branch. Add mutable.MultiMap and immutable.MultiMap leaves.

julienrf:
This is another idea to provide a `MultiMap` collection type. Here, `MultiMap[K, V]` is a new “branch” of the hierarchy, it inherits from `Iterable[(K, Set[V])]`. The PR provides two implementations, a mutable one and an immutable one.

Cons:
- the fact that `MultiMap` has its own branch makes it incomparable with `Map` collections.
- if we want to make it work with an underlying `SortedMap` we have to create a new branch for that in the hierarchy.

Pros:
- `map`, `flatMap` and `concat` return a `MultiMap[K, V]` instead of a `Map[K, Set[V]`
- maybe cleaner hierarchy?

Also, when we manipulate a `MultiMap[K, V]`, sometimes we want to see the entries as `(K, Set[V])` pairs, but sometimes we want to see them as `(K, V)` pairs… I don’t see a reason to not support both, actually. So, what I’ve done is that several methods have two variants: `map` shows the entries as `(K, Set[V])` pairs, and `mapByValue` shows the entries as `(K, V)` pairs ; `iterator()` returns an `Iterator[(K, Set[V])]` and `iteratorByValue()` returns an `Iterator[(K, V)]`. Basically I used the suffix “byValue” when an operation sees its entries as `(K, V)` pairs.

### DIFF
--- a/collections-contrib/src/main/scala/strawman/collection/MultiMap.scala
+++ b/collections-contrib/src/main/scala/strawman/collection/MultiMap.scala
@@ -1,0 +1,79 @@
+package strawman
+package collection
+
+trait MultiMap[K, V]
+  extends Iterable[(K, Set[V])]
+    with MultiMapOps[K, V, MultiMap, MultiMap[K, V]]
+    with Equals {
+
+  final protected[this] def coll: this.type = this
+
+  def canEqual(that: Any): Boolean = true
+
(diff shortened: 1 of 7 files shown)


julienrf:
Superseded by #314

PR #273, opened 2017-10-19
A collection of smaller changes from my integration branch

szeiger:
for better compatibility with the old collections library

### DIFF
--- a/collections/src/main/scala/strawman/collection/BuildFrom.scala
+++ b/collections/src/main/scala/strawman/collection/BuildFrom.scala
@@ -1,6 +1,6 @@
 package strawman.collection
 
-import scala.{Any, Ordering}
+import scala.{Any, Ordering, deprecated, `inline`}
 
 import strawman.collection.mutable.Builder
 
@@ -17,6 +17,9 @@ trait BuildFrom[-From, -A, +C] extends Any {
   /** Get a Builder for the collection. For non-strict collection types this will use an intermediate buffer.
     * Building collections with `fromSpecificIterable` is preferred because it can be lazy for lazy collections. */
   def newBuilder(from: From): Builder[A, C]
+
+  @deprecated("Use newBuilder() instead of apply()", "2.13.0")
+  @`inline` def apply(from: From): Builder[A, C] = newBuilder(from)
 }
 
 object BuildFrom extends BuildFromLowPriority {
--- a/collections/src/main/scala/strawman/collection/Iterable.scala
+++ b/collections/src/main/scala/strawman/collection/Iterable.scala
@@ -3,7 +3,7 @@ package collection
 
 import scala.annotation.unchecked.uncheckedVariance
 import scala.reflect.ClassTag
-import scala.{Any, AnyRef, Array, Boolean, `inline`, Int, None, Numeric, Option, Ordering, PartialFunction, StringContext, Some, Unit}
+import scala.{Any, Array, Boolean, `inline`, Int, None, Numeric, Option, Ordering, PartialFunction, StringContext, Some, Unit, deprecated, IllegalArgumentException, Function1, AnyRef}
 import java.lang.{String, UnsupportedOperationException}
 import scala.Predef.<:<
 
@@ -11,10 +11,13 @@ import strawman.collection.mutable.{ArrayBuffer, Builder, StringBuilder}
 import java.lang.String
 
 /** Base trait for generic collections */
-trait Iterable[+A] extends IterableOnce[A] with IterableOps[A, Iterable, Iterable[A]] {
+trait Iterable[+A] extends IterableOnce[A] with IterableOps[A, Iterable, Iterable[A]] with Traversable[A] {
 
   /** The collection itself */
   final def toIterable: this.type = this
+
+  //TODO scalac generates an override for this in AbstractMap; Making it final leads to a VerifyError
+  protected[this] def coll: this.type = this
 
 }
 
@@ -33,7 +36,7 @@ trait Iterable[+A] extends IterableOnce[A] with IterableOps[A, Iterable, Iterabl
   *  @tparam C  type of the collection (e.g. `List[Int]`, `String`, `BitSet`). Operations returning a collection
   *             with the same type of element (e.g. `drop`, `filter`) return a `C`.
   */
-trait IterableOps[+A, +CC[X], +C] extends Any {
+trait IterableOps[+A, +CC[_], +C] extends Any with IterableOnce[A] {
 
   /**
     * @return This collection as an `Iterable[A]`. No new collection will be built if `this` is already an `Iterable[A]`.
@@ -116,6 +119,12 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
 
   /** Fold right */
   def foldRight[B](z: B)(op: (A, B) => B): B = toIterable.iterator().foldRight(z)(op)
+
+  @deprecated("Use foldLeft instead of /:", "2.13.0")
+  @`inline` final def /: [B](z: B)(op: (B, A) => B): B = foldLeft[B](z)(op)
+
+  @deprecated("Use foldRight instead of :\\", "2.13.0")
+  @`inline` final def :\ [B](z: B)(op: (A, B) => B): B = foldRight[B](z)(op)
 
   /** Reduces the elements of this $coll using the specified associative binary operator.
    *
@@ -225,6 +234,12 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
   /** The first element of the collection. */
   def head: A = toIterable.iterator().next()
 
+  /** The first element of the collection. */
+  def headOption: Option[A] = {
+    val it = toIterable.iterator()
+    if(it.hasNext) Some(it.next()) else None
+  }
+
   /** Selects the last element.
     * $orderDependent
     * @return The last element of this $coll.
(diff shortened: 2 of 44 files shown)


collections/src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -266,6 +284,19 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
...
+  def toList: immutable.List[A] = immutable.List.from(toIterable)
+
+  def toVector: immutable.Vector[A] = immutable.Vector.from(toIterable)
+
+  def toMap[T, U](implicit ev: A <:< (T, U)): immutable.Map[T, U] =
+    immutable.Map.from(toIterable.asInstanceOf[Iterable[(T, U)]])
+
+  def toSet[B >: A]: immutable.Set[B] = immutable.Set.from(toIterable)
+
+  def toSeq[B >: A]: immutable.Seq[B] = immutable.Seq.from(toIterable)
```
julienrf: Is it useful to have this `B` type parameter? Why not just return an `immutable.Seq[A]`?

collections/src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -266,6 +284,19 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
     */
   def to[C1](factory: Factory[A, C1]): C1 = factory.fromSpecific(toIterable)
 
+  def toList: immutable.List[A] = immutable.List.from(toIterable)
+
+  def toVector: immutable.Vector[A] = immutable.Vector.from(toIterable)
+
+  def toMap[T, U](implicit ev: A <:< (T, U)): immutable.Map[T, U] =
```
julienrf: Minor remark about naming: we should try to stick to the same conventions everywhere. `Map` names its type parameter `K` and `V`, we should reuse the same names here.

collections/src/main/scala/strawman/collection/immutable/package.scala
```diff
@@ -0,0 +1,8 @@
+package strawman.collection
+
+package object immutable {
+  type StringOps = strawman.collection.StringOps
+  val StringOps = strawman.collection.StringOps
+  type StringView = strawman.collection.StringView
+  val StringView = strawman.collection.StringView
```
julienrf: Are these aliases used somewhere?

collections/src/main/scala/strawman/collection/mutable/Builder.scala
```diff
@@ -45,6 +45,9 @@ trait Builder[-A, +To] extends Growable[A] { self =>
     }
   }
 
+  @deprecated("Use sizeHint(coll, 0) instead of sizeHint(coll)", "2.13.0")
+  @`inline` final def sizeHint(coll: strawman.collection.Iterable[_]): Unit = sizeHint(coll, 0)
```
julienrf: What do you think of simply having a default value `0` for the `delta` parameter?

collections/src/main/scala/strawman/collection/BuildFrom.scala
```diff
@@ -17,6 +17,9 @@ trait BuildFrom[-From, -A, +C] extends Any {
   /** Get a Builder for the collection. For non-strict collection types this will use an intermediate buffer.
     * Building collections with `fromSpecificIterable` is preferred because it can be lazy for lazy collections. */
   def newBuilder(from: From): Builder[A, C]
+
+  @deprecated("Use newBuilder() instead of apply()", "2.13.0")
+  @`inline` def apply(from: From): Builder[A, C] = newBuilder(from)
```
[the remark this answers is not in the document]
szeiger: Not yet but it might make sense to add

collections/src/main/scala/strawman/collection/immutable/List.scala
```diff
@@ -345,20 +345,76 @@ sealed trait List[+A]
...
+    */
+  @`inline` final def mapConserve[B >: A <: AnyRef](f: A => B): List[B] = {
+    // Note to developers: there exists a duplication between this function and `reflect.internal.util.Collections#map2Conserve`.
+    // If any successful optimization attempts or other changes are made, please rehash them there too.
+    //@tailrec
+    def loop(mappedHead: List[B] = Nil, mappedLast: ::[B], unchanged: List[A], pending: List[A]): List[B] = {
+      if (pending.isEmpty) {
+        if (mappedHead eq null) unchanged
+        else {
+          mappedLast.next = (unchanged: List[B])
```
szeiger: This did not compile on Dotty without the explicit widening. I get a poor error message without a position:
```
[info] Compiling 76 Scala sources and 5 Java sources to /Users/szeiger/code/scala/collection-strawman/collections/.jvm/target/scala-0.3/classes...
[error]
[error] covariant type A occurs in contravariant position in type strawman.collection.immutable.List[A] of value x$0
[error] one error found
```

collections/src/main/scala/strawman/collection/immutable/List.scala
```diff
@@ -345,20 +345,75 @@ sealed trait List[+A]
...
+    *
+    *  @param f      the function to apply to each element.
+    *  @tparam B     the element type of the returned collection.
+    *  @return       a list resulting from applying the given function
+    *                `f` to each element of this list and collecting the results.
+    *
+    *  @usecase def mapConserve(f: A => A): List[A]
+    *    @inheritdoc
+    */
+  @`inline` final def mapConserve[B >: A <: AnyRef](f: A => B): List[B] = {
```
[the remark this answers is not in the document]
szeiger: It's useful for performance and it relies on mutating the list, which requires package-private access.

collections/src/main/scala/strawman/collection/Iterable.scala
szeiger: I thought it's currently done this way (for some reason) but it turns out it is not.

collections/src/main/scala/strawman/collection/Seq.scala
```diff
@@ -252,6 +249,16 @@ trait SeqOps[+A, +CC[X], +C] extends Any
...
 
+  /** Tests whether this $coll contains given index.
+    *
+    *  The implementations of methods `apply` and `isDefinedAt` turn a `Seq[A]` into
+    *  a `PartialFunction[Int, A]`.
+    *
+    * @param    idx     the index to test
+    * @return   `true` if this $coll contains an element at position `idx`, `false` otherwise.
+    */
+  def isDefinedAt(idx: Int): Boolean = (idx >= 0) && (idx < length)
```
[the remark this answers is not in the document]
szeiger: Indeed, the old one does, too

collections/src/main/scala/strawman/collection/immutable/package.scala
szeiger: Yes, because that's where these classes used to be. I think we should move them back (and had already done so) but the code was easier to rebase with them remaining in the collection package.

collections/src/main/scala/strawman/collection/mutable/Builder.scala
szeiger: Makes sense

collections/src/main/scala/strawman/collection/immutable/SeqExtractors.scala
```diff
@@ -0,0 +1,23 @@
...
+/** An extractor used to init/last deconstruct sequences. */
+object :+ {
+  /** Splits a sequence into init :+ last.
+   * @return Some((init, last)) if sequence is non-empty. None otherwise.
+   */
+  def unapply[T, CC[_], Coll <: SeqOps[T, CC, Coll]](
+      t: Coll with SeqOps[T, CC, Coll]): Option[(Coll, T)] =
+    if(t.isEmpty) None
+    else Some((t.init, t.last))
+}
```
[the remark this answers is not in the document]
szeiger: Maybe I overlooked them or didn't properly get them into scope. I'll remove them from here for now.

szeiger:
rebased

collections/src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -266,6 +284,19 @@ trait IterableOps[+A, +CC[X], +C] extends Any {
...
+  def toVector: immutable.Vector[A] = immutable.Vector.from(toIterable)
+
+  def toMap[K, V](implicit ev: A <:< (K, V)): immutable.Map[K, V] =
+    immutable.Map.from(toIterable.asInstanceOf[Iterable[(K, V)]])
+
+  def toSet[B >: A]: immutable.Set[B] = immutable.Set.from(toIterable)
+
+  def toSeq: immutable.Seq[A] = immutable.Seq.from(toIterable)
+
+  def toIndexedSeq[B >: A]: immutable.IndexedSeq[B] = immutable.IndexedSeq.from(toIterable)
```
julienrf: In this one too, we should remove the `B` type parameter.

Also, I’d be in favor of deprecating `toIndexedSeq`, `toList` and `toVector`, but that’s another issue.

collections/src/main/scala/strawman/collection/Seq.scala
```diff
@@ -598,12 +607,126 @@ trait SeqOps[+A, +CC[X], +C] extends Any
...
+    *   @param   len   the test value that gets compared with the length.
+    *   @return  A value `x` where
+    *   {{{
+    *        x <  0       if this.length <  len
+    *        x == 0       if this.length == len
+    *        x >  0       if this.length >  len
+    *   }}}
+    *  The method as implemented here does not call `length` directly; its running time
+    *  is `O(length min len)` instead of `O(length)`. The method should be overwritten
+    *  if computing `length` is cheap.
```
julienrf: So, I think we should override it in `IndexedSeq`

julienrf:
Hey @szeiger, I’ve just added two small commits with the changes I was proposing in my last comments. Let me know if they are OK for you and I’ll squash them and merge the PR.

szeiger:
Yes, looks good.

julienrf:
Rebased.

PR #275, opened 2017-10-25
Add withDefault and withDefaultValue to Map types

szeiger:
There is no shared abstraction for mutable and immutable Map types
because of differences in variance. This is the same as in the old
collections library.


collections/src/main/scala/strawman/collection/immutable/Map.scala
```diff
@@ -10,7 +10,28 @@ import scala.{Any, Boolean, `inline`, Int, None, NoSuchElementException, Nothing
 trait Map[K, +V]
   extends Iterable[(K, V)]
      with collection.Map[K, V]
-     with MapOps[K, V, Map, Map[K, V]]
+     with MapOps[K, V, Map, Map[K, V]] {
+
+  /** The same map with a given default function.
+    *  Note: `get`, `contains`, `iterator`, `keys`, etc are not affected by `withDefault`.
```
julienrf: Instead of giving an incomplete list of methods followed by “etc” I would clearly say what’s the impact: it provides an implementation for the `default` method, which is (internally) used only by the `apply` method.

collections/src/main/scala/strawman/collection/immutable/Map.scala
```diff
@@ -84,6 +105,35 @@ trait MapOps[K, +V, +CC[X, +Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]
...
+    // valid for the underlying collection.
+    protected[this] def fromSpecificIterable(coll: collection.Iterable[(K, V)]): Map[K,V] = new WithDefault[K, V](mapFactory.from(coll), d)
+    protected[this] def newSpecificBuilder(): mutable.Builder[(K, V), Map[K,V]] = {
+      val ub = mapFactory.newBuilder[K, V]()
+      new mutable.Builder[(K, V), Map[K, V]] {
+        def clear(): Unit = ub.clear()
+        def result(): Map[K, V] = new WithDefault[K, V](ub.result(), d)
+        def add(elem: (K, V)): this.type = { ub.add(elem); this }
+      }
+    }
```
julienrf: It seems that it should be possible to use `mapFactory.newBuilder[K, V]().mapResult(WithDefault[K, V](_, d))` instead.

collections/src/main/scala/strawman/collection/immutable/Map.scala
```diff
@@ -84,6 +105,35 @@ trait MapOps[K, +V, +CC[X, +Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]
...
+    override def size = underlying.size
+    def get(key: K) = underlying.get(key)
+    def iterator() = underlying.iterator()
+    override def default(key: K): V = d(key)
+    override def empty = new WithDefault(underlying.empty, d)
+    override def updated[V1 >: V](key: K, value: V1): WithDefault[K, V1] = new WithDefault[K, V1](underlying.updated[V1](key, value), d)
+    override def remove (key: K): WithDefault[K, V] = new WithDefault(underlying - key, d)
+    override def withDefault[V1 >: V](d: K => V1): immutable.Map[K, V1] = new WithDefault[K, V1](underlying, d)
+    override def withDefaultValue[V1 >: V](d: V1): immutable.Map[K, V1] = new WithDefault[K, V1](underlying, x => d)
```
julienrf: Why do we need to override `withDefault` and `withDefaultValue`? The implementation is the same as the base implementation.

julienrf:
Could you also update the README to indicate that the `withDefault` and `withDefaultValue` operations are now supported?

szeiger:
You can't and shouldn't return a `CC`. You don't want to reimplement a `WithDefault` subclass for every concrete Map implementation. It wouldn't provide any advantage, either. `WithDefault` is more like a view (which we do not have for Maps in the general case), not a transformation. There is no rebuilding going on that could use the standard mechanisms for that. Every single affected operation needs to be reimplemented for every single subtype.

collections/src/main/scala/strawman/collection/immutable/Map.scala
szeiger: No, it's not. It avoids layering several `WithDefault` on top of each other.

szeiger:
Updated and rebased.

nafg:
I wonder if it should really be defined on PartialFunction...

julienrf:
What behavior would you recommend?

PR #276, opened 2017-10-26
Enable scalacheck tests

### LINKED ISSUE #103: Add correctness tests
A way of automatically verifying correctness of the implemented collections would be nice to have in addition to the benchmarking suite (especially during the optimization phase). 

Two existing code bases to look at are: 
https://github.com/scala/scala/tree/2.13.x/test/junit/scala/collection
https://github.com/scala/scala-collections-laws

julienrf:
Fixes #103.

I’ve adapted the scalacheck tests to use the strawman collections instead of the standard collections.

I’ve fixed a couple of issues I’ve discovered by doing so.

Two collections haven’t been adapted, `UnrolledBuffer` and `PriorityQueue`, because these classes don’t exist in the strawman. We can still enable them afterwards, if we introduce these collections.

Also, the tests against the `java.util.concurrent.ConcurrentHashMap` wrapper haven’t been adapted because the strawman version of this adapter is in #279 (which is unmerged yet).

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ For more information, see the [CONTRIBUTING](CONTRIBUTING.md) file.
 - [x] `indices`
 - [x] `isDefinedAt`
 - [x] `isEmpty` / `nonEmpty`
-- [x] `keysIteratorFrom`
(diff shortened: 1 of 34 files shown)


julienrf:
I’ve merged this for the release of today, let me know if you see a problem…

PR #278, opened 2017-10-27
Primitive array types in ArrayOps / ArrayBuilder / ImmutableArray

szeiger:
This is similar to the old implementation. Primitive array types are not
polymorphic on the JVM so we need to special-case them in order to be
able to efficiently wrap existing arrays as collections.


szeiger:
The test doesn't compile on Dotty anymore with this changes. The error message is not very helpful:
```
[info] Compiling 1 Scala source to /Users/szeiger/code/scala/collection-strawman/collections/.jvm/target/scala-0.4/test-classes...
[error]
[error] Type argument strawman.collection.immutable.IndexedSeq has not the same kind as its bound  <: [_$61] => scala.Any
[error] one error found
[error] (collectionsJVM/test:compileIncremental) Compilation failed
```

collections/src/main/scala/strawman/collection/ArrayOps.scala
```diff
@@ -54,11 +69,158 @@ class ArrayOps[A](val xs: Array[A])
...
+  def empty[T <: AnyRef]: ArrayView[T] = EmptyArrayView.asInstanceOf[ArrayView[T]]
+
+  @`inline` def apply[T](x: Array[T]): ArrayView[T] = make(x)
+
+  // If make is called explicitly we use whatever we're given, even if it's
+  // empty.  This may be unnecessary (if ArrayView is to honor the collections
+  // contract all empty ones must be equal, so discriminating based on the reference
+  // equality of an empty array should not come up) but we may as well be
+  // conservative since wrapRefArray contributes most of the unnecessary allocations.
+  def make[T](x: AnyRef): ArrayView[T] = (x match {
```
julienrf: Why is the type of `x` `AnyRef` rather than `Array[_]` (for instance)? Do we want this method to be public?

collections/src/main/scala/strawman/collection/ArrayOps.scala
```diff
@@ -54,11 +69,158 @@ class ArrayOps[A](val xs: Array[A])
...
+      case that: ofBoolean => Arrays.equals(array, that.array)
+      case _ => super.equals(that)
+    }
+  }
+
+  final class ofUnit(val array: Array[Unit]) extends ArrayView[Unit] with Serializable {
+    def elemTag = ClassTag.Unit
+    def length: Int = array.length
+    def apply(index: Int): Unit = array(index)
+    override def hashCode = arrayViewHash(array)
```
julienrf: Why do we want to override `ArrayView`’s hashcode?

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -152,7 +151,140 @@ object ImmutableArray extends StrictOptimizedSeqFactory[ImmutableArray] {
...
+    case x: Array[Long]    => new ofLong(x)
+    case x: Array[Float]   => new ofFloat(x)
+    case x: Array[Char]    => new ofChar(x)
+    case x: Array[Byte]    => new ofByte(x)
+    case x: Array[Short]   => new ofShort(x)
+    case x: Array[Boolean] => new ofBoolean(x)
+    case x: Array[Unit]    => new ofUnit(x)
+  }).asInstanceOf[ImmutableArray[T]]
+
+  final class ofRef[T <: AnyRef](val array: Array[T]) extends ImmutableArray[T] with Serializable {
```
julienrf: I think we should follow the well established naming convention where class names begin with an upper case letter.

collections/src/main/scala/strawman/collection/mutable/ArrayBuilder.scala
```diff
@@ -0,0 +1,665 @@
...
+      if (capacity < size) resize(size)
+    }
+
+    private def ensureSize(size: Int): Unit = {
+      if (capacity < size || capacity == 0) {
+        var newsize = if (capacity == 0) 16 else capacity * 2
+        while (newsize < size) newsize *= 2
+        resize(newsize)
+      }
+    }
```
julienrf: Do you have an idea of the loss of performance we would get if we defined these methods in `ArrayBuilder`?

collections/src/main/scala/strawman/collection/mutable/ArrayBuilder.scala
```diff
@@ -0,0 +1,665 @@
...
+  }
+
+  /** A class for array builders for arrays of reference types.
+   *
+   *  This builder can be reused.
+   *
+   *  @tparam T     type of elements for the array builder, subtype of `AnyRef` with a `ClassTag` context bound.
+   */
+  final class ofRef[T <: AnyRef : ClassTag] extends ArrayBuilder[T] {
```
sjrd: Can we make those `private`? Scala.js does not like them. We have an entirely different implementation of `ArrayBuilder` in Scala.js. For example see https://github.com/scala-js/scala-js/blob/7ea476aa62349f681bb2dbba16b7ffa6470bd786/scalalib/overrides-2.13/scala/collection/mutable/ArrayBuilder.scala#L32-L73

collections/src/main/scala/strawman/collection/ArrayOps.scala
szeiger: It's currently done with `AnyRef` but it should work with `Array[_]` (which also erases to `Object`). Since this is the only way of creating an `ArrayView` that takes primitive arrays into account, it needs to be public.

collections/src/main/scala/strawman/collection/ArrayOps.scala
szeiger: To make use of specialization

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
szeiger: They're public and copied directly from the current API. I wouldn't want to do any unnecessary breaking changes here.

collections/src/main/scala/strawman/collection/mutable/ArrayBuilder.scala
szeiger: I haven't benchmarked it but I wouldn't expect any performance difference. `ensureSize` and `sizeHint` don't touch the actual array. They should be OK to pull up.

collections/src/main/scala/strawman/collection/mutable/ArrayBuilder.scala
szeiger: These types are exposed as part of the API so that you can use them directly with unboxed primitive element types in all cases where you don't need polymorphism over primitive and reference types. Having to code against a non-specialized superclass means you need to box and unbox every single element you append to the Builder.

szeiger:
Updated:
- Rebased
- Workaround for Dotty bug (see comment in ArrayOps.scala)
- Removed `ArrayView` in favor of `ImmutableArray`
- More abstraction in specialized `ArrayBuilder`s

collections/src/main/scala/strawman/collection/ArrayOps.scala
```diff
@@ -1,34 +1,47 @@
...
+  with StrictOptimizedIterableOps[A, Seq, Array[A]]
+  with ArrayLike[A] {
+
+  protected def fromTaggedIterable[B: ClassTag](coll: Iterable[B]): Array[B] = coll.toArray[B]
+
+  // Note: Declaring the return type as ArrayOps.WithFilter[A] breaks Dotty 0.4.0-RC1
+  override def withFilter(p: A => Boolean): collection.WithFilter[A, immutable.IndexedSeq] = new ArrayOps.WithFilter[A](p, this)
 
-  def toIterable = ArrayView(xs)
+  def toIterable: ImmutableArray[A] = ImmutableArray.wrapArray(xs)
```
julienrf: Actually I think it is not safe to expose it as an `immutable` one because the underlying array `xs` is still available and can be mutated.

collections/src/main/scala/strawman/collection/package.scala
```diff
@@ -104,7 +104,7 @@ class LowPriority {
   import strawman.collection._
 
   /** Convert array to iterable via view. Lower priority than ArrayOps */
-  implicit def arrayToView[T](xs: Array[T]): ArrayView[T] = ArrayView[T](xs)
+  implicit def arrayToImmutableArray[T](xs: Array[T]): immutable.ImmutableArray[T] = immutable.ImmutableArray.wrapArray[T](xs)
```
julienrf: Here too, we should use `collection.IndexedSeq[T]` as a return type instead of `immutable.ImmutableArray[T]`

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -152,7 +151,140 @@ object ImmutableArray extends StrictOptimizedSeqFactory[ImmutableArray] {
...
+    protected def elements: Array[T] = array
+    def length: Int = array.length
```
julienrf: Should we pull up `length`’s implementation?

collections/src/main/scala/strawman/collection/ArrayOps.scala
szeiger: That's a general risk with `ImmutableArray` but it is unavoidable if we want to make the default `Seq` a `collection.immutable.Seq` and support varargs as `Seq`.

Or should we limit the implicit conversion from `Array` to produce a generic `Seq`? Varargs could still expose it as an immutable one. It may still be somewhat confusing once we make the immutable `Seq` the default one.

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
szeiger: No, that would require a polymorphic `array` field (which erases to `Object`) in `ImmutableArray`. All methods which operate on the actual `array` field need to be in the subclasses.

collections/src/main/scala/strawman/collection/ArrayOps.scala
julienrf: Yes, I would expose `collection.IndexedSeq` as the public result type of `toIterable`. We can still safely use `ImmutableArray` to support varargs since the reference to the underlying mutable `Array` will not leak (or will it?).

szeiger: It will but I'd classify that as an abuse of varargs. If you mutate an array after passing it to a varags method all bets are off.

szeiger:
Updated to return a generic `IndexedSeq` as the static type when implicitly wrapping an `Array` as an `ImmutableArray`.

collections/src/main/scala/strawman/collection/ArrayOps.scala
```diff
@@ -1,34 +1,47 @@
...
+
+class ArrayOps[A](val xs: Array[A]) extends AnyVal
+  with IterableOnce[A]
+  with IndexedSeqOps[A, immutable.IndexedSeq, Array[A]]
+  with StrictOptimizedIterableOps[A, Seq, Array[A]]
+  with ArrayLike[A] {
+
+  protected def fromTaggedIterable[B: ClassTag](coll: Iterable[B]): Array[B] = coll.toArray[B]
+
+  // Note: Declaring the return type as ArrayOps.WithFilter[A] breaks Dotty 0.4.0-RC1
```
julienrf: Does that mean that we won’t get back an `Array` but an `IndexedSeq` if we call `map`?

julienrf: Could you file an issue in the dotty repository (lampepfl/dotty)? These issues have high priority and might be fixed in a few days.

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -152,7 +151,140 @@ object ImmutableArray extends StrictOptimizedSeqFactory[ImmutableArray] {
       ScalaRunTime.array_update(elements, i, f(i))
       i = i + 1
     }
-    new ImmutableArray(elements)
+    ImmutableArray.wrapArray(elements)
+  }
+
+  def wrapArray[T](x: AnyRef): ImmutableArray[T] = (x match {
```
julienrf: Does this method have to be public? We should at least document what it does and also warn about the fact that it doesn’t create a copy of the given `Array`.

sjrd: It needs to be public in the bytecode so that the codegen can generate code that calls it, in cases that are known to be safe (i.e., varargs generation). However it could be private-ish at the language level. 

One way is to make it `protected[immutable]`.

collections/src/main/scala/strawman/collection/ArrayOps.scala
szeiger: Strangely enough I was unable to reproduce the error again. Let's see how CI fares.

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
szeiger: I don't think it will buy us anything to hide the method. The specialized subclasses expose the constructors and the wrapped arrays, and both are necessary to write high-performance code that uses ImmutableArray. Also exposing wrapArray on top of that does not make it any less safe. The current collections do the same and I have not seen any problems in the wild that were caused by this.

julienrf: Having it non-public would just prevent confusion.

sjrd: > The current collections do the same and I have not seen any problems in the wild that were caused by this.

Hum ... the current collections don't have `ImmutableArray`s to begin with, so I'm not sure what you're talking about.

szeiger:
Opened https://github.com/lampepfl/dotty/issues/3467 against Dotty

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
szeiger: It's called `WrappedArray` in the current collections. (`WrappedArray` is a *mutable* collection though, so maybe it would be more problematic for `ImmutableArray`)

sjrd: > WrappedArray is a mutable collection though

Ah, well, yes. That changes everything!

szeiger:
So what should we do about the `ImmutableArray` constructors / extractors? IMHO being able to wrap and unwrap arrays into collections is too important to forego.

- For decent varargs support with the default `scala.Seq` being immutable we need the full `ImmutableArray`, including specialization. Access to internals could be hidden in the `scala` namespace. But we have two additional use cases:

- Explicit conversion from arrays to collections: Can we rely on users to use this responsibly when we pretend that an `ImmutableArray` is immutable even though we expose the non-copying constructors? I think we can. It is already commonplace to treat mutable collections as effectively immutable after construction. `ImmutableArray` goes a bit further by being an `immutable.Seq` but it provides sensible semantics for this use case: The additional methods for immutable Seqs are implemented without any structural sharing. If we can't agree on this interpretation we will have to bring `WrappedArray` back. It would be exactly the same as `ImmutableArray` except that it implements `mutable.Seq` instead of `immutable.Seq` (and is therefore not as convenient to use because it is not a `scala.Seq` anymore)

- Implicit array wrapping: Pretending to have an immutable collection could be dangerous because these conversions are not always obvious. I would make this contingent on the decision for the previous case. If we add `WrappedArray`, that's the right one to use, otherwise use `ImmutableArray` but declare the return type as `collection.Seq` rather than `immutable.Seq`.

julienrf:
> Explicit conversion from arrays to collections: Can we rely on users to use this responsibly when we pretend that an ImmutableArray is immutable even though we expose the non-copying constructors? I think we can

In such a case I would make it as clear as possible in the method name. E.g. `ImmutableArray.fromArrayNoCopy`.

sjrd:
I definitely think we need `WrappedArray` to wrap an existing `Array` into a `Seq`.

The current `collection.Coll`s are read-only, not immutable. We do not have precedent to create an `immutable.Coll` that we can still mutate. `immutable.Coll`s need to stay immutable.

julienrf:
@sjrd What about the following implementation?

~~~ scala
package object collection {
  type WrappedArray[A] <: Seq[A]
  def wrap[A](array: Array[A]): WrappedArray[A] = ImmutableArray.fromArrayNoCopy(array)
}
~~~

sjrd:
In addition to the run-time type check issue, that implementation is not helpful when one wants to wrap an `Array` and give to someone who wants a `mutable.Seq`, in order to actually mutate it.

Ichoran:
@sjrd - `immutable.BitSet` has `fromBitMaskNoCopy` that results in an "immutable" object where the backing array can be mutated.  So there is precedent.  (That doesn't make it a good idea, though.)

julienrf:
@szeiger I think that bringing back `WrappedArray` would make sense. I created #300 to keep track of that work (which is out of the scope of that PR). In the meanwhile do you agree with making `fromArray` package private?

szeiger:
I still think we need non-private wrapping and unwrapping even in the presence of `WrappedArray`. This is also the only way to implement them separately but I'd still prefer to add `WrappedArray` back to this PR (it was present in the original version) directly.

szeiger:
Looking at `WrappedArray` again I wonder why we should have it at all. Why not add the manual specialization to `ArrayBuffer` instead? There doesn't seem to be any substantial difference between the current `ArrayBuffer` and `WrappedArray.ofRef`.

julienrf:
The difference is that `ArrayBuffer` is a `Growable`, whereas `WrappedArray` isn’t. Not sure if it is that important in practice, though.

sjrd:
`WrappedArray` guarantees that it is a *view* of an `Array`. Changes to one are visible from the other. `ArrayBuffer` is just a buffer that happens to *backed*, privately, by an `Array` (to provide its complexity guarantees). That's the fundamental difference (much more so than `Growable` versus not-growable).

julienrf:
Right @sjrd I missed that one and I agree it’s important.

szeiger:
We'll need to make some bigger changes for that. Currently every `mutable.Iterable` is `Growable` and furthermore every `mutable.Seq` is `Shrinkable`.

szeiger:
Everything should be ready here, including `WrappedArray` and rebasing to build on Dotty 0.5.0

julienrf:
I’m a bit lost with the builders. Why do we need `ArrayBuilder` at all? Can’t `ArrayBuffer[A]` *be* a `Builder[A, Array[A]]` (like we do with `ListBuffer`)?

collections/src/main/scala/strawman/collection/mutable/WrappedArray.scala
```diff
@@ -0,0 +1,310 @@
...
+    } else fromArrayBuffer(ArrayBuffer.from(it))
+  }
+
+  def fromArrayBuffer[A](arr: ArrayBuffer[A]): WrappedArray[A] =
+    make[A](arr.asInstanceOf[ArrayBuffer[Any]].toArray)
+
+  def newBuilder[A](): Builder[A, WrappedArray[A]] =
+    ArrayBuffer.newBuilder[A]().mapResult(fromArrayBuffer)
+
+  def newBuilder[A](implicit t: ClassTag[A]): Builder[A, WrappedArray[A]] = new WrappedArrayBuilder[A](t)
```
julienrf: Why do we need `WrappedArrayBuilder`? Can’t we use `ArrayBuilder.mapResult(fromArray)`?

collections/src/main/scala/strawman/collection/mutable/WrappedArray.scala
```diff
@@ -0,0 +1,310 @@
...
+      while (i < n) {
+        ScalaRunTime.array_update(elements, i, iterator.next())
+        i = i + 1
+      }
+      make(elements)
+    } else fromArrayBuffer(ArrayBuffer.from(it))
+  }
+
+  def fromArrayBuffer[A](arr: ArrayBuffer[A]): WrappedArray[A] =
+    make[A](arr.asInstanceOf[ArrayBuffer[Any]].toArray)
```
julienrf: Why `fromArrayBuffer` rather than `fromArray`?

Ichoran:
At least in the existing collections, `ArrayBuffer` stores elements as `AnyRef`; `ArrayBuilder` has to build an array of the correct type (including primitive type, if asked for).

collections/src/main/scala/strawman/collection/mutable/WrappedArray.scala
szeiger: We have that in the form of `make`. `fromArrayBuffer` is only used in two places. We could inline it or make the method private.

collections/src/main/scala/strawman/collection/mutable/WrappedArray.scala
szeiger: I don't see any reason why we couldn't.

szeiger:
Updated: Removed `WrappedArrayBuilder`, make `WrappedArray.fromArrayBuffer` private, rebased

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -16,52 +19,55 @@ import scala.Predef.intWrapper
...
   extends IndexedSeq[A]
     with IndexedSeqOps[A, ImmutableArray, ImmutableArray[A]]
     with StrictOptimizedSeqOps[A, ImmutableArray, ImmutableArray[A]] {
 
   def iterableFactory: SeqFactory[ImmutableArray] = ImmutableArray
 
+  /* The wrapped mutable `Array` */
+  def unsafeArray: Array[A @uncheckedVariance]
+  // uncheckedVariance should be safe: Array[A] for reference types A is covariant at the JVM level. Array[A] for
+  // primitive types A can only be widened to Array[Any] which erases to Object.
```
julienrf: No it is not safe:

~~~ scala
val rectangles: ImmutableArray[Rectangle] = ImmutableArray(rect1, rect2)
val shapes: ImmutableArray[Shape] = rectangles // assuming Rectangle is a Shape
val shapesArray: Array[Shape] = shapes.unsafeArray
shapesArray(0) = someCircle // assuming Circle is a Shape which is not a Rectangle
rectangles.foreach(rect => println(rect.height)) // ClassCastException! Circle can not be cast to Rectangle
~~~

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -150,7 +158,131 @@ object ImmutableArray extends StrictOptimizedSeqFactory[ImmutableArray] {
       ScalaRunTime.array_update(elements, i, f(i))
       i = i + 1
     }
-    new ImmutableArray(elements)
+    ImmutableArray.unsafeWrapArray(elements)
+  }
+
+  /** Wrap an existing `Array` into an `ImmutableArray` of the proper primitive specialization type without copying. */
+  def unsafeWrapArray[T](x: AnyRef): ImmutableArray[T] = (x match {
```
julienrf: Just curious here: `x: Array[_]` doesn’t work?

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -150,7 +158,131 @@ object ImmutableArray extends StrictOptimizedSeqFactory[ImmutableArray] {
...
+    case x: Array[Int]     => new ofInt(x)
+    case x: Array[Double]  => new ofDouble(x)
+    case x: Array[Long]    => new ofLong(x)
+    case x: Array[Float]   => new ofFloat(x)
+    case x: Array[Char]    => new ofChar(x)
+    case x: Array[Byte]    => new ofByte(x)
+    case x: Array[Short]   => new ofShort(x)
+    case x: Array[Boolean] => new ofBoolean(x)
+    case x: Array[Unit]    => new ofUnit(x)
+  }).asInstanceOf[ImmutableArray[T]]
```
julienrf: In the end we never statically know that we have an underlying `ofInt` instance. We always manipulate an `ImmutableArray[X]`. How does the JVM know that it doesn’t have to box/unbox elements?

collections/src/main/scala/strawman/collection/mutable/WrappedArray.scala
```diff
@@ -0,0 +1,242 @@
...
+
+  def iterableFactory: strawman.collection.SeqFactory[WrappedArray] = WrappedArray
+
+  protected[this] def fromSpecificIterable(coll: strawman.collection.Iterable[T]): WrappedArray[T] = {
+    val b = ArrayBuilder.make()(elemTag)
+    val s = coll.knownSize
+    if(s > 0) b.sizeHint(s)
+    b ++= coll
+    WrappedArray.make(b.result())
+  }
```
julienrf: Looks like `iterableFactory.from(coll)` would be a better implementation.

collections/src/main/scala/strawman/collection/mutable/WrappedArray.scala
```diff
@@ -0,0 +1,242 @@
...
+  protected[this] def newSpecificBuilder(): Builder[T, WrappedArray[T]] = ArrayBuilder.make()(elemTag).mapResult(WrappedArray.make)
```
julienrf: Why do we have different implementations here and in `WrappedArray.newBuilder`?

julienrf:
Did you measure the performance difference?

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
szeiger: Did you test this? I would expect it to fail with an `ArrayStoreException` in `shapesArray(0) = someCircle`. But both this error and the one you demonstrate are due to the inherent unsoundness of covariant arrays.

This could be mitigated in `ImmutableArray` by making it invariant, but for an *immutable* array this should not be necessary. The JVM array actually *is* covariant and.  If you use it in a mutable way all bets are off. That's why it is called *unsafe* in the first place. Any immutable use should be sound.

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
szeiger: It doesn't. Any access through `ImmutableArray` is boxed. If you want unboxed access you need to cast down or instantiate a primitive subclass directly. It's the same for `WrappedArray`. In theory a `@specialized` annotation should be beneficial for both cases but since `WrappedArray` isn't currently specialized I didn't want to do any speculative changes without more complete benchmarks.

collections/src/main/scala/strawman/collection/mutable/WrappedArray.scala
szeiger: No, the `IterableFactory` cannot make use of the existing `elemTag`. It would always have to build a boxed array.

szeiger:
Addressed review comments. I didn't do any performance tests since this only brings the implementation up to the previous standard in 2.12.

PR #279, opened 2017-10-27
Add back the non-deprecated Java collection converter APIs

szeiger:
Pretty much 1:1 from the current 2.12/2.13 impl. Deprecated parts only when necessary to implement the non-deprecated ones.

### DIFF
--- a/collections/src/main/scala/strawman/collection/convert/AsScalaConverters.scala
+++ b/collections/src/main/scala/strawman/collection/convert/AsScalaConverters.scala
@@ -1,0 +1,211 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___     Scala API                            **
+**    / __/ __// _ | / /  / _ |    (c) 2006-2016, LAMP/EPFL             **
+**  __\ \/ /__/ __ |/ /__/ __ |    http://www.scala-lang.org/           **
+** /____/\___/_/ |_/____/_/ | |                                         **
+**                          |/                                          **
+\*                                                                      */
+
+package strawman
+package collection
+package convert
+
+import scala.Predef.String
+import java.{ lang => jl, util => ju }, java.util.{ concurrent => juc }
+
+/** Defines converter methods from Java to Scala collections. */
+trait AsScalaConverters {
+  import Wrappers._
+
+  /**
+   * Converts a Java `Iterator` to a Scala `Iterator`.
+   *
+   * The returned Scala `Iterator` is backed by the provided Java `Iterator` and any side-effects of
+   * using it via the Scala interface will be visible via the Java interface and vice versa.
+   *
+   * If the Java `Iterator` was previously obtained from an implicit or explicit call of
+   * `[[JavaConverters.asJavaIterator]](scala.collection.Iterator)` then the original Scala
+   * `Iterator` will be returned.
+   *
+   * @param i The Java `Iterator` to be converted.
+   * @return  A Scala `Iterator` view of the argument.
+   */
+  def asScalaIterator[A](i: ju.Iterator[A]): Iterator[A] = i match {
+    case null                     => null
+    case IteratorWrapper(wrapped) => wrapped
+    case _                        => JIteratorWrapper(i)
+  }
+
+  /**
+   * Converts a Java `Enumeration` to a Scala `Iterator`.
+   *
+   * The returned Scala `Iterator` is backed by the provided Java `Enumeration` and any side-effects
+   * of using it via the Scala interface will be visible via the Java interface and vice versa.
+   *
+   * If the Java `Enumeration` was previously obtained from an implicit or explicit call of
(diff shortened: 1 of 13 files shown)


szeiger:
Fails to compile under Dotty:

```
[error] -- [E007] Type Mismatch Error: /Users/szeiger/code/scala/collection-strawman/collections/src/main/scala/strawman/collection/convert/Wrappers.scala:56:34
[error] 56 |    def iterableFactory = mutable.ArrayBuffer
[error]    |                          ^^^^^^^^^^^^^^^^^^^
[error]    |found:    strawman.collection.mutable.ArrayBuffer.type
[error]    |required: strawman.collection.IterableFactoryLike[strawman.collection.Iterable]
[error]    |
[error] -- [E007] Type Mismatch Error: /Users/szeiger/code/scala/collection-strawman/collections/src/main/scala/strawman/collection/convert/Wrappers.scala:65:34
[error] 65 |    def iterableFactory = mutable.ArrayBuffer
[error]    |                          ^^^^^^^^^^^^^^^^^^^
[error]    |found:    strawman.collection.mutable.ArrayBuffer.type
[error]    |required: strawman.collection.IterableFactoryLike[strawman.collection.Iterable]
[error]    |
[error] -- [E007] Type Mismatch Error: /Users/szeiger/code/scala/collection-strawman/collections/src/main/scala/strawman/collection/convert/Wrappers.scala:135:34
[error] 135 |    def iterableFactory = mutable.ArrayBuffer
[error]     |                          ^^^^^^^^^^^^^^^^^^^
[error]     |found:    strawman.collection.mutable.ArrayBuffer.type
[error]     |required: strawman.collection.SeqFactory[strawman.collection.Seq] &
[error]     |  strawman.collection.IterableFactoryLike[strawman.collection.mutable.Buffer]
[error]     |
```

`+CC[_]` is covariant in all the factory types. It's not clear to me why this doesn't compile.

julienrf:
About the type error: I think I’ve had the same issue and it works if you explicitly add the return type of the member.

szeiger:
I've tried all kinds of type annotations, nothing helps. Apparently the compiler really thinks these types are incompatible.

collections/src/main/scala/strawman/collection/convert/AsScalaConverters.scala
```diff
@@ -0,0 +1,211 @@
...
+   * @param m The Java `Map` to be converted.
+   * @return  A Scala mutable `Map` view of the argument.
+   */
+  def mapAsScalaMap[A, B](m: ju.Map[A, B]): mutable.Map[A, B] = m match {
+    case null                       => null
+    case MutableMapWrapper(wrapped) => wrapped
+    case _                          => new JMapWrapper(m)
+  }
+
+  //TODO Add concurrent maps back
```
julienrf: Isn’t `TrieMap` enough?

collections/src/main/scala/strawman/collection/convert/Wrappers.scala
```diff
@@ -0,0 +1,494 @@
...
+        case Some(v) => v
+      }
+    } catch {
+      case ex: ClassCastException => null.asInstanceOf[B]
+    }
+
+    override def clear() = underlying.clear()
+  }
+
+  trait JMapWrapperLike[A, B, +Repr <: mutable.Map[A, B] with mutable.Map[A, B]] extends mutable.Map[A, B] with mutable.MapOps[A, B, mutable.Map, mutable.Map[A, B]] {
```
julienrf: For consistency it’s better if we use `C` everywhere instead of `Repr`, and, in the case of Maps, `K` and `V` instead of `A` and `B`, respectively.

szeiger:
I just realized I was still using 0.3.0-RC2 for local testing and the CI build on 0.4.0-RC1 failed for a different reason which can be easily fixed.

collections/src/main/scala/strawman/collection/convert/AsScalaConverters.scala
szeiger: We don't need any specific implementation, just the *concept* of a `scala.collection.concurrent.Map`. Without that there's no point in treating it differently from any other map.

julienrf: But we do have [it](https://static.javadoc.io/ch.epfl.scala/collection-strawman_2.13.0-M2/0.5.0/strawman/collection/concurrent/Map.html), right?

julienrf:
Did you adapt existing tests or write your own? In the diff I see no removed/renamed file

szeiger:
They are copies of the old tests that we already had in the repo for `scala.collection.convert`

szeiger:
Squashed and rebased.

collections/src/main/scala/strawman/collection/convert/AsScalaConverters.scala
szeiger: Ah, right. It wasn't there yet when I started working on the branch. I'll reenable it.

PR #280, opened 2017-10-28
Make ListBuffer and ArrayBuffer serializable.

### LINKED ISSUE #272: Make ListBuffer serializable
I just noticed that our implementation is not `Serializable` whereas in the current collection it is.

jiminhsieh:
Fix #272.


julienrf:
Thanks ;)

PR #282, opened 2017-10-30
Simplify toXxx methods on IterableOnce

julienrf:
Since now the `from` factory method accept an `IterableOnce` we don’t need to check whether the receiver is actually an `Iterable` or not.

We could even make these `toXxx` methods available (without deprecation) on `IterableOnce`. What do you think?

### DIFF
--- a/collections/src/main/scala/strawman/collection/IterableOnce.scala
+++ b/collections/src/main/scala/strawman/collection/IterableOnce.scala
@@ -42,23 +42,17 @@ final class IterableOnceExtensionMethods[A](private val it: IterableOnce[A]) ext
     case _ => it.iterator().foreach(f)
   }
 
-  @deprecated("Use ArrayBuffer.fromIterable(it) instead of it.toBuffer on Iterable (wrap Iterators with View.fromIterator first)", "2.13.0")
-  def toBuffer[B >: A]: mutable.Buffer[B] = it match {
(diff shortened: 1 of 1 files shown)


julienrf:
@szeiger any thoughts about that one?

Ichoran:
I would recommend not deprecating any `toXxx` methods that we currently have.  Currently on `TraversableOnce` we have `toArray`, `toBuffer`, `toIndexedSeq`, `toIterable`, `toIterator`, `toList`, `toMap`, `toSeq`, `toSet`, `toStream`, `toTraversable`, and `toVector`.  Except for `toTraversable`, I don't see a compelling reason to deprecate any of the others unless the implementation can't be efficient.

julienrf:
@Ichoran just to give a more complete view of what we have: currently `Iterable` has all these methods excepted `toBuffer` and `toStream`. In `IterableOnce` they are available via a deprecated implicit conversion. My suggestion is to pull these methods up to `IterableOnce`, so that the implicit conversion is not even needed.

julienrf:
I agree that we should add `toBuffer` and `toLazyList`, for the sake of compatibility.

julienrf:
> We deliberately removed all methods other than iterator() from it to keep it minimal. Should we reverse that decision and aim for easier abstraction rather than simplicity?

Hmm, you’re right. I think it’s better to leave `IterableOnce` with no methods.

PR #285, opened 2017-11-03
Fix broken scalafix tests

marcelocenerine:
There are two scalafix tests failing:

- scala/fix/Collectionstrawman_v0_Traversable.scala:
```
[info]   =======
[info]   => Diff
[info]   =======
[info]        xs.to(strawman.collection.Map)
[info]   -    xs.iterator
[info]   +    xs.iterator()
```

`scala.collection.TraversableLike.toIterator` was being converted to `iterable` without `()`

- scala/fix/Collectionstrawman_v0.scala:
```
[info]   =======
[info]   => Diff
[info]   =======
[info]   -import scala.Predef.{ augmentString => _, intArrayOps => _, wrapString => _, _ }
[info]   +import scala.Predef.{ augmentString => _, genericArrayOps => _, genericWrapArray => _, intArrayOps => _, wrapIntArray => _, wrapString => _, _ }
```

Here some unimports from Predef were not being added to the resulting file (`genericArrayOps`, `genericWrapArray` and `wrapIntArray`).

### DIFF
--- a/scalafix/output/src/main/scala/fix/Collectionstrawman_v0.scala
+++ b/scalafix/output/src/main/scala/fix/Collectionstrawman_v0.scala
@@ -1,6 +1,6 @@
 package fix
 
-import scala.Predef.{ augmentString => _, genericArrayOps => _, genericWrapArray => _, intArrayOps => _, wrapIntArray => _, wrapString => _, _ }
+import scala.Predef.{ augmentString => _, genericArrayOps => _, intArrayOps => _, wrapString => _, _ }
 import strawman.collection.{ arrayToArrayOps, stringToStringOps }
 import strawman.collection.immutable.{ ::, HashMap, LazyList, List, Map, Nil, Range, Vector }
 import strawman.collection.immutable.LazyList.#::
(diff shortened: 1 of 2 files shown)


scalafix/output/src/main/scala/fix/Collectionstrawman_v0.scala
```diff
@@ -1,6 +1,6 @@
 package fix
 
-import scala.Predef.{ augmentString => _, genericArrayOps => _, genericWrapArray => _, intArrayOps => _, wrapIntArray => _, wrapString => _, _ }
+import scala.Predef.{ augmentString => _, genericArrayOps => _, intArrayOps => _, wrapString => _, _ }
 import strawman.collection.{ arrayToArrayOps, stringToStringOps }
```
marcelocenerine: I'm not pretty sure whether `genericWrapArray` and `wrapIntArray` should also be added. It was enough to include only `genericArrayOps` to make `output/../Collectionstrawman_v0_List` compile. Please let me know if it makes more sense to change the scalafix rule to unimport those 2 as well.

julienrf: In some tests I did, these conversions could also be picked. But if it does compile without unimporting them in that test, that’s fine. We might refine that behaviour later…

julienrf:
Thanks Marcelo ;)

Issue #241, opened 2017-09-14
Consider adding `split` operations


julienrf:
Scalaz’s `Foldable` has two variants of partitioning methods, named `splitWith` and `selectSplit`:

https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Foldable.scala#L303-L351

markus1189:
I am interested, will give it a try soon

julienrf:
FYI I’ve used `splitWith` as a test in this PR: https://github.com/scala/scala/pull/6674/files#diff-1ac5384a2a49471ae923396b6ee3f269R47

markus1189:
@julienrf Is it still supposed to be in `collections-contrib` and inside a `SeqDecorator` as it was originally?  Seems like a lot of the stuff there was deleted.  Where did you use `splitWith` in that pr?  It's not there, or I am blind :wink: 

Also, where can I find the `IsSeqLike`?

julienrf:
Oops, I’ve updated the PR yesterday and renamed it to `groupedWith`: https://github.com/scala/scala/pull/6674/files#diff-1ac5384a2a49471ae923396b6ee3f269R80

Here is the story about `IsSeqLike`. It’s part of the standard library but in collections-contrib we have something similar named `HasSeqOps`. The goal of the PR I linked is to unify them.

The idea I have in mind would be for you work to target the collections-contrib module. But you can also do a PR on scala/scala (without using `IsSeqLike` since you will be able to implement it directly in the collections, then) and start a discussion to have them in the standard library…

LPTK:
Here's another naming quibble!

I think the names `splitWith` and `selectSplit` are terrible!

 * `selectSplit` should be called `runsWhere` as it splits a sequence into runs where a predicate is true;

 * `splitWith` sounds like it would do like `String`'s `split` but using an element predicate instead of an actual character (as in `"ab$cd$efg".splitWith(notLetter) == Seq("ab","cd","efg")`); maybe a better name would be `runsUntilFlip`? I'm not sure the function is that useful though; I'd prefer a function that returns two distinct sequences, and it would be called `partitionRunsWhere`.

julienrf:
We have precedent for an operation returning consecutive subsequences: `grouped`. I think we should use `grouped` instead of `runs`.

LPTK:
[...]

PR #286, opened 2017-11-03
Add the `splitWith` operation to SeqDecorator

markus1189:
Relates to #241. This pull request adds the first of the two proposed methods.

There is still some bikeshedding to do before merging, I will comment when it's ready, but please comment!

Greetings from the Scala Center FLOSS Spree at Scala.io 2017!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -85,7 +85,10 @@ val `collections-contrib` =
   crossProj("collections-contrib", file("collections-contrib"))
     .dependsOn(collections)
     .settings(
-      name := "collections-contrib"
+      name := "collections-contrib",
+      libraryDependencies ++= Seq(
+        ("org.scalacheck" %% "scalacheck" % "1.13.5" % Test).withDottyCompat()
+      )
     )
 
 val `collections-contrib-jvm` = `collections-contrib`.jvm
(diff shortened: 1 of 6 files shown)


markus1189:
Okay I updated the pr, ready for comments.  The things I don't like:

1) we have to duplicate the Generators in the test because the `scalacheck` subproject does not cross-compile

2) `splitWith` currently works only for `Seq` types, it would be nice if for example `String` would be supported as well

julienrf:
Thanks @markus1189! I’ll try to find some time to address the issues you raised.

markus1189:
@julienrf I'm happy to do more work as well, If you could give me some pointers

julienrf:
The decorator doesn’t work on `String` values for two reasons:
- `String` is not a collection, it needs to be implicitly converted to a `StringView` to become a collection, but since our decorators also require an implicit conversion to be applied, we have to apply two implicit conversions in a row, which is disallowed by the type system. We could circumvent this issue by taking the implicit conversion to a collection as an implicit parameter of the implicit conversion to the decorator.
- `StringView` doesn’t conform to the upper bound of the `SeqDecorator` implicit conversion parameter type: `CC[X] <: SeqOps[A, CC, _]`. `StringView` is an `IndexedView[Char]`, which is a `SeqOps[Char, View, _]` (whereas we need it to be a `SeqOps[Char, IndexedView, _]`).

PR #291, opened 2017-11-07
Scalafix rewrite for lazyZip

marcelocenerine:
Enhancing the scalafix rule to rewrite `(xs, ys).zipped` to `xs.lazyZip(ys)` (likewise for triples).

There is one caveat here: `LazyZipN.filter` [is not equivalent to](https://github.com/scala/collection-strawman/pull/223#issuecomment-336372816) `TupleNZipped.filter`. This will likely require manual intervention to be fully fixed. Choosing not to rewrite `(xs, ys).zipped.filter(p)` would leave it in a broken state anyway because a `TupleNZipped` cannot be created with iterables from strawman.

/cc @olafurpg

### DIFF
--- a/scalafix/input/src/main/scala/fix/Collectionstrawman_v0_TupleNZipped.scala
+++ b/scalafix/input/src/main/scala/fix/Collectionstrawman_v0_TupleNZipped.scala
@@ -1,0 +1,36 @@
+/*
+rule = "scala:fix.Collectionstrawman_v0"
+ */
+package fix
+
+import scala.language.postfixOps
+object Collectionstrawman_v0_Tuple2Zipped {
+  def zipped(xs: List[Int], ys: List[Int]): Unit = {
+    (xs, ys).zipped
+    (xs,ys).zipped
+    ((xs, ys) zipped)
+    (((xs)    ,    (ys)).zipped)
+    (xs, // foo
+      ys).zipped
+    /* a */(/* b */ xs /* c */, /* d */ ys /* e */)/* f */./* g */zipped/* h */
+    (coll(1), coll(2)).zipped
+    (List(1, 2, 3), Stream.from(1)).zipped
+  }
+  def coll(x: Int): List[Int] = ???
+}
+
+object Collectionstrawman_v0_Tuple3Zipped {
+  def zipped(xs: List[Int], ys: List[Int], zs: List[Int]): Unit = {
+    (xs, ys, zs).zipped
+    (xs,ys,zs).zipped
+    ((xs, ys, zs) zipped)
+    (((xs)    ,    (ys)    ,    (zs)).zipped)
+    (xs, // foo
+      ys, // bar
+      zs).zipped
+    /* a */(/* b */ xs /* c */, /* d */ ys /* e */, /* f */ zs /* g */)/* h */./* i */zipped/* j */
+    (coll(1), coll(2), coll(3)).zipped
+    (List(1, 2, 3), Set(1, 2, 3), Stream.from(1)).zipped
+  }
+  def coll(x: Int): List[Int] = ???
+}
--- a/scalafix/output/src/main/scala/fix/Collectionstrawman_v0_TupleNZipped.scala
+++ b/scalafix/output/src/main/scala/fix/Collectionstrawman_v0_TupleNZipped.scala
@@ -1,0 +1,34 @@
+package fix
+
+import scala.language.postfixOps
+import strawman.collection.immutable.{ LazyList, List, Set }
+object Collectionstrawman_v0_Tuple2Zipped {
+  def zipped(xs: List[Int], ys: List[Int]): Unit = {
+    xs.lazyZip(ys)
+    xs.lazyZip(ys)
+    (xs.lazyZip(ys) )
+    ((xs).lazyZip((ys)))
+    xs.lazyZip(// foo
+      ys)
+    /* a *//* b */ xs /* c */.lazyZip(/* d */ ys /* e */)/* f *//* g *//* h */
+    coll(1).lazyZip(coll(2))
+    List(1, 2, 3).lazyZip(LazyList.from(1))
+  }
+  def coll(x: Int): List[Int] = ???
+}
+
+object Collectionstrawman_v0_Tuple3Zipped {
+  def zipped(xs: List[Int], ys: List[Int], zs: List[Int]): Unit = {
+    xs.lazyZip(ys).lazyZip(zs)
+    xs.lazyZip(ys).lazyZip(zs)
+    (xs.lazyZip(ys).lazyZip(zs) )
+    ((xs).lazyZip((ys)).lazyZip((zs)))
+    xs.lazyZip(// foo
+      ys).lazyZip(// bar
+      zs)
+    /* a *//* b */ xs /* c */.lazyZip(/* d */ ys /* e */).lazyZip(/* f */ zs /* g */)/* h *//* i *//* j */
+    coll(1).lazyZip(coll(2)).lazyZip(coll(3))
+    List(1, 2, 3).lazyZip(Set(1, 2, 3)).lazyZip(LazyList.from(1))
+  }
+  def coll(x: Int): List[Int] = ???
+}
--- a/scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
+++ b/scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
@@ -116,6 +116,10 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
     Symbol("_root_.scala.collection.LinearSeqLike.iterator."),
     Symbol("_root_.scala.collection.TraversableLike.toIterator.")
   )
+  val tupleZipped = SymbolMatcher.normalized(
+    Symbol("_root_.scala.runtime.Tuple2Zipped.Ops.zipped."),
+    Symbol("_root_.scala.runtime.Tuple3Zipped.Ops.zipped.")
+  )
 
   def replaceToList(ctx: RuleCtx) =
     ctx.tree.collect {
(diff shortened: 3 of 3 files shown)


scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
```diff
@@ -135,10 +139,23 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
...
+      case tupleZipped(t @ Term.Select(Term.Tuple(args), _)) =>
+        val argTokens = args.flatMap(_.tokens).toSet
+        val removeTokensPatch = t.tokens.filter(!argTokens.contains(_)).map(ctx.removeToken).asPatch
+        val addTokensPatch = args.tail.map { arg =>
+          val tokens = arg.tokens
+          ctx.addLeft(tokens.head, ".lazyZip(") + ctx.addRight(tokens.last, ")")
+        }.asPatch
+        removeTokensPatch + addTokensPatch
+    }.asPatch
+
```
marcelocenerine: I had to do this at the token level as suggested by @olafurpg due to a limitation in scalafix that is explaned here: https://gitter.im/scalacenter/scalafix?at=59ff60b35a1758ed0f90981b

scalafix/output/src/main/scala/fix/Collectionstrawman_v0_TupleNZipped.scala
```diff
@@ -0,0 +1,20 @@
...
+  }
+  def coll(x: Int): List[Int] = ???
+}
+
+object Collectionstrawman_v0_Tuple3Zipped {
+  def zipped(xs: List[Int], ys: List[Int], zs: List[Int]): Unit = {
+    xs.lazyZip(ys).lazyZip(zs)
+    coll(1).lazyZip(coll(2)).lazyZip(coll(3))
+    List(1, 2, 3).lazyZip(Set(1, 2, 3)).lazyZip(LazyList.from(1))
+  }
```
marcelocenerine: Apparently there isn't a consensus on the syntax for `lazyZip` in the discussions in #223. I chose `xs.lazyZip(ys).xxx(f)` over (xs lazyZip ys).xxx(f)`. Please let me know if you prefer the other option.

olafurpg: I would go for `a.lazyZip(b)`, it's harder to get the parentheses right for infix operators. For example, `(a, (b, c)).zipped` becomes `a lazyZip ((b, c))` under -Yno-adapted-args. You might also have to add wrapping parentheses depending on the parent tree node `(a lazyZip b).c`

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
```diff
@@ -135,10 +139,23 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
             ctx.replaceToken(close, ")")).getOrElse(Patch.empty)
     }.asPatch
 
+  def replaceTupleZipped(ctx: RuleCtx) =
+    ctx.tree.collect {
+      case tupleZipped(t @ Term.Select(Term.Tuple(args), _)) =>
+        val argTokens = args.flatMap(_.tokens).toSet
```
olafurpg: `tupleElementTokens` is more descriptive than `argTokens`

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
```diff
@@ -135,10 +139,23 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
...
+        val removeTokensPatch = t.tokens.filter(!argTokens.contains(_)).map(ctx.removeToken).asPatch
```
olafurpg: `nonTupleElementTokens`

`t.tokens.filterNot(tupleElementTokens)`

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
```diff
@@ -135,10 +139,23 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
...
+        val addTokensPatch = args.tail.map { arg =>
+          val tokens = arg.tokens
+          ctx.addLeft(tokens.head, ".lazyZip(") + ctx.addRight(tokens.last, ")")
```
olafurpg: There are `addRight/addLeft` overloads for trees: `ctx.addLeft(arg, ".lazyZip(") + ctx.addRight(arg, ")")`

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
olafurpg: PS, you might be removing comments here, I would keep comments

scalafix/input/src/main/scala/fix/Collectionstrawman_v0_TupleNZipped.scala
```diff
@@ -0,0 +1,22 @@
+/*
+rule = "scala:fix.Collectionstrawman_v0"
+ */
+package fix
+
+object Collectionstrawman_v0_Tuple2Zipped {
+  def zipped(xs: List[Int], ys: List[Int]): Unit = {
+    (xs, ys).zipped
+    (coll(1), coll(2)).zipped
```
olafurpg: I recommend you add at least one test with comments inside of the tuple, the biggest reason scalafix exposes low-level token APIs is so that trivia like comments and formatting can be preserved.

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
olafurpg: The only tokens that need to be removed a leading `(`, trailing `)`, commas `,` and `.zipped`.

marcelocenerine: yeah, you are right. Everything that is not a tuple element gets removed.

Preserving comments + trivia makes the rewrite a lot more complicated :smile:. For instance:
`(((xs, ys) /* foo1 */ zipped) /* foo2 */)` 

would ideally be rewritten to... 

`((xs.lazyZip(ys) /* foo1 */ ) /* foo2 */)`

Note that in this case the leading `(` and trailing `)` are not the outmost ones. It is totally doable (in fact I got it working already) but requires some extra steps to figure out the correct parenthesis and deal with the potentially missing `dot`.

Also, I think it would be nice to get rid of empty spaces between the comma and the first non-empty space token that follows it. Otherwise the output will be something like: 
`xs.lazyZip( ys)`

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
marcelocenerine: cool, I missed that

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
```diff
@@ -132,13 +136,55 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
...
+              removeSurroundingWhiteSpaces(head) ++
+              tail.map { comma =>
+                ctx.replaceToken(comma, ").lazyZip(") +
+                  removeSurroundingWhiteSpaces(comma)
+              }
+          case _ => Patch.empty
+        }
+
+        removeTokensPatch + replaceCommasPatch
     }.asPatch
```
marcelocenerine: Handling comments make the rewrite a lot more complicated compared to the previous version :sweat_smile:.
Note that in lines 153-155 I'm preserving the original closing `)` in order to allow comments on the left side of the last tuple element to be enclosed by the correct parenthesis. Similar care needs to be taken in regards to comments on the left side of a middle triple element (that's the reason for the additional lines in line 174-175).

marcelocenerine:
@olafurpg thank you for your awesome suggestions! I updated the rule, so please take a look at it again

scalafix/input/src/main/scala/fix/Collectionstrawman_v0_TupleNZipped.scala
marcelocenerine: done

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
```diff
@@ -132,13 +136,55 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
...
+            closeTuple <- ctx.tokenList.leading(zipped).find(_.is[Token.RightParen])
+            openTuple <- ctx.matchingParens.open(closeTuple.asInstanceOf[Token.RightParen])
+            maybeDot = ctx.tokenList.slice(closeTuple, zipped).find(_.is[Token.Dot])
+          } yield {
+            ctx.removeToken(openTuple) +
+              maybeDot.map(ctx.removeToken).asPatch +
+              ctx.removeToken(zipped)
+          }).asPatch
+
+        def removeSurroundingWhiteSpaces(tk: Token) =
```
olafurpg: Utilities for trimming whitespace are severely missing in scalafix. I'd love to hear if you have idea how this can be generalized

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
```diff
@@ -132,13 +136,55 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
...
+
+  def replaceTupleZipped(ctx: RuleCtx) =
+    ctx.tree.collect {
+      case tupleZipped(Term.Select(Term.Tuple(args), name)) =>
+        val removeTokensPatch =
+          (for {
+            zipped <- name.tokens.headOption
+            closeTuple <- ctx.tokenList.leading(zipped).find(_.is[Token.RightParen])
+            openTuple <- ctx.matchingParens.open(closeTuple.asInstanceOf[Token.RightParen])
+            maybeDot = ctx.tokenList.slice(closeTuple, zipped).find(_.is[Token.Dot])
```
olafurpg: Utilities to find dots is missing in scalafix, I've written something similar to this many times 😅

marcelocenerine:
One thing I realized that this PR is missing is replacing the old types in case someone assigned the zipped result to a explicitly typed variable or parameter before invoking operations on it (not sure if this really happens in practice as the result would be rather verbose):
```Tuple3Zipped[A, Col1[A], B, Col2[B], C, Col3[C]] zipped = (xs, ys).zipped```

If you think this is a scenario worth covering, I can address it before the PR is merged.

julienrf:
I don’t think lots of people have explicitly used the `TupleNZipped` type. In my opinion, this rewrite is great in its current state.

olafurpg:
I agree with @julienrf I would not worry about it. If you want you can report a warning if `Tuple3Zipped` is inferred

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
marcelocenerine: it would definitely be handy to have such utility in the api. Maybe something like `ctx.trimLeft(token)` / `ctx.trimRight(token)` or `ctx.removeTokens(tokenList.spacesBefore(token))` / `ctx.removeTokens(tokenList.spacesAfter(token))`. I can put more thoughts on this and create a PR in scalafix with a proposal.

marcelocenerine:
ok, I'm going to keep the PR as is. @julienrf you can merge PR if you are happy with it.

julienrf:
Thanks a lot @marcelocenerine !

PR #295, opened 2017-11-13
Make foldSomeLeft and lazyFoldRight decorators

### LINKED ISSUE #107: FoldWhile

julienrf:
Fixes #107 and #186. Based on #228 and #225, rebased, squashed and moved to collections-contrib. I removed `foldSomeRight` (I only kept `foldSomeLeft` and `lazyFoldRight`).

### DIFF
--- a/collections-contrib/src/main/scala/strawman/collection/decorators/IterableDecorator.scala
+++ b/collections-contrib/src/main/scala/strawman/collection/decorators/IterableDecorator.scala
@@ -4,7 +4,30 @@ package decorators
 
 class IterableDecorator[A](val `this`: Iterable[A]) extends AnyVal {
 
+  /**
+    * Left to right fold that stops if the combination function `op`
+    * returns `None`
+    * @param z the start value
+    * @param op the binary operator
+    * @tparam B the result type of the binary operator
+    * @return the result of inserting `op` between consecutive elements of the collection,
+    *         going left to right with the start value `z` on the left, and stopping when
+    *         all the elements have been traversed or earlier if the operator returns `None`
+    */
   def foldSomeLeft[B](z: B)(op: (B, A) => Option[B]): B =
     `this`.iterator().foldSomeLeft(z)(op)
+
+  /**
+    * Right to left fold that can be interrupted before traversing the whole collection.
+    * @param z the start value
+    * @param op the operator
+    * @tparam B the result type
+    * @return the result of applying the operator between consecutive elements of the collection,
+    *         going right to left, with the start value `z` on the right. The result of the application
+    *         of the function `op` to each element drives the process: if it returns `Left(result)`,
+    *         then `result` is returned without iterating further; if it returns `Right(f)`, the function
+    *         `f` is applied to the previous result to produce the new result and the fold continues.
+    */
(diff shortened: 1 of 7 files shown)


SethTisue:
bikeshed: can we call it `foldLeftSome` rather than `foldSomeLeft`? @odd thought so on #228; @Ichoran disagreed but didn't have a strong feeling about it.

I do have a pretty strong feeling about this; I find `foldSomeLeft` is really strange and unnatural. it sounds like "some" modifies "left" which is a wrong impression to give, whereas `foldLeftSome` correctly conveys that this is a variant of `foldLeft`.

Ichoran:
@SethTisue - Er, but it _does_ fold only some of the left?

Not that it's an aesthetically appealing name, but I don't think it gives the wrong idea.  You need some way to indicate that drops the right tail rather than filtering out the offending elements.

SethTisue:
`foldLeft` until you hit a `Some`, or `foldLeft` only "some" of the way to the end — not seeing the problem.

Ichoran:
I don't see much of a problem either; I just meant that the reasoning you used against the more-awkward-sounding version doesn't make much sense to me.  However, "it sounds too awkward" can be a good enough reason to change the color of the bike shed, so to speak.

hayd:
bikeshed: Apologies if this is already mentioned elsewhere (ah, it's here: https://github.com/scala/collection-strawman/pull/228), but should this be an example of a more general `foldLeftUntil`? Where in this case the we're folding until `_.isEmpty`. I think that naming would be immediately clear vs this one.
But perhaps `foldLeftUntil(_.isEmpty)(...)` is too verbose.

Lasering:
Implementing the `foldLeftUntil` instead of the `foldSomeLeft` sounds like a very good balance between being a generic method, conveying well what it will do and still be very succint method to use.

Issue #298, opened 2017-11-16
Add ceiling and floor methods for TreeSet and TreeMap 


Psycho7:
`floor` method returns the max value less than or equal to a given value. Since TreeSet and TreeMap are based on red-black tree, `O(logn)` should be the complexity of this kind operations. However, currently there is no way to to this in `O(logn)`. For example, `TreeSet(1, 3, 5).to(4).max` will return the max value less than 4, which is 3. However, an intermediate TreeSet will be created (and rebalanced) because `to` method was called, thus the complexity is worse than `O(logn)`. I'm not sure if it's `O(nlogn)` or `O(n)`.

In Java, `floor` and `ceiling` are provided for TreeSet, and `floorKey`, `floorEntry`, `ceilingKey` and `ceilingEntry` for TreeMap. I think they should be provided Scala in a elegant way.

acruise:
+1, I've wanted these before and gone back to j.u equivalents

julienrf:
Also, [here](https://users.scala-lang.org/t/when-using-treeset-how-to-get-ceiling-or-floor-value-efficiently/1920) is a relevant discussion.

Psycho7:
What do those label mean? I'm glad to contribute if it is not done yet.

julienrf:
“low hanging fruit” means that it’s an affordable issue (it doesn’t require a lot of work). “ready” means that anyone can start working on it! If you want to work on it I can pass it in “in progress” state :)

Psycho7:
More than glad to help. :)

julienrf:
Awesome, thanks!

Psycho7:
I found that `maxBefore` and `minAfter` in `mutable.RedBlackTree` and they are used in `mutable.TreeMapVew` and `mutable.TreeSetView`. These methods can do exactly what I want and are elegant. Actually these methods are implicitly called when you call `head / last` on `TreeSetView / TreeMapView` with `from / to` enabled. 

However, there is no immutable `TreeMapView` or `TreeSetView`. Is there any plan to have them?

Another question is, should I implement these methods in `TreeMap` / `TreeSet` directly and make them public?

julienrf:
> However, there is no immutable TreeMapView or TreeSetView. Is there any plan to have them?

No. You can add them if they are necessary (but it doesn’t look like it’s the case).

> Another question is, should I implement these methods in TreeMap / TreeSet directly and make them public?

Yes!

Psycho7:
[...]

PR #302, opened 2017-11-23
Add `maxBefore` and `minAfter` methods to `TreeMap` and `TreeSet`

Psycho7:
Related to #298

### DIFF
--- a/collections/src/main/scala/strawman/collection/SortedMap.scala
+++ b/collections/src/main/scala/strawman/collection/SortedMap.scala
@@ -59,6 +59,18 @@ trait SortedMapOps[K, +V, +CC[X, Y] <: Map[X, Y] with SortedMapOps[X, Y, CC, _],
   def firstKey: K = head._1
   def lastKey: K = last._1
 
+  /** Find the element with smallest key larger than or equal to a given key.
+    * @param key The given key.
+    * @return `None` if there is no such node.
+    */
+  def minAfter(key: K): Option[(K, V)] = from(key).headOption
+
+  /** Find the element with largest key less than a given key.
+    * @param key The given key.
+    * @return `None` if there is no such node.
+    */
+  def maxBefore(key: K): Option[(K, V)] = until(key).lastOption
+
   def rangeTo(to: K): C = {
     val i = keySet.from(to).iterator()
     if (i.isEmpty) return coll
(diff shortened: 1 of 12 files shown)


Psycho7:
> Do we really need both versions (the one that returns an Option and the one that doesn’t)?

In fact we don't. At first I only wanted to implement ones without `Option`. For `Map`, it's fine to return `null` when no such element, but it's bad for `Set`. So I added Option versions. In some 3rd party libraries they keep both versions (w/o option), that's why I keep them. I'd like to delete ones without `Option` just as you said.

> I think it would be nice to have these methods defined on collection.SortedSet and collection.SortedMap rather than only in the leaf classes.

Yeah this is a good idea! I even thought about adding a new trait saying they are implemented in BST. It makes sense to put them in `collection.SortedSet` and `colletion.SortedMap`. 

Thank you for your excellent suggestion!

julienrf:
Hey @Psycho7, would you have time to address my feedback and squash your commits in the next couple of days? It would be nice to include your contribution in the next release.

Psycho7:
Updated. The error message from Travis CI is ...

> sbt.ResolveException: download failed: com.google.guava#guava;19.0!guava.jar(bundle)

It seems like there are some network issues in Travis CI's servers. How to restart the build process? This commit should be good.

collections/src/main/scala/strawman/collection/SortedMap.scala
```diff
@@ -59,6 +59,18 @@ trait SortedMapOps[K, +V, +CC[X, Y] <: Map[X, Y] with SortedMapOps[X, Y, CC, _],
   def firstKey: K = head._1
   def lastKey: K = last._1
 
+  /** Find the element with smallest key larger than or equal to a given key.
+    * @param key The given key.
+    * @return `None` if there is no such node.
+    */
+  def minAfter(key: K) = from(key).headOption
```
julienrf: Could you add the return type explicitly?

julienrf: Also, I’d be tempted to not have a default implementation if we always override it in leaf classes.

Psycho7:
@julienrf Done. Previously compiling failed because `Option` was not imported, so I tried removing return types but forgot to get them back. It's awkward...

Psycho7:
> Also, I’d be tempted to not have a default implementation if we always override it in leaf classes.

Actually that's not true. At first I also didn't want a default implementation. But there are some classes extending `SortedSet` or `SortedMap`. For example, `BitSet extends SortedSet[Int]`.

julienrf:
OK that’s perfect, thanks for the explanation! Could you squash your commits into a single one?

Psycho7:
@julienrf Like this?

julienrf:
Yes, looks OK!

PR #304, opened 2017-11-26
Rename TreeMapView and TreeSetView to avoid confusion with Views.

### LINKED ISSUE #301: Rename TreeMapView and TreeSetView to avoid confusion with Views
We should replace the “View” suffix with something different (e.g. `TreeMapSlice`, `TreeSetProjection`, …).

jiminhsieh:
Fix #301. 
There is one thing I am not so sure. I didn't rename values' name that declares in each test. For example, I still kept `mapView` instead of `mapProjection`. As far as I know, those two words could be interchange. If it's better to keep the same suffix for values' name in each test, please let me know. Thanks! :)


julienrf:
Perfect, thanks!

PR #307, opened 2017-11-30
Update Dotty to 0.5.0-RC1

allanrenucci:


allanrenucci:
BTW, this should unblock #278

PR #315, opened 2017-12-07
Fix inconsistencies between size and length implementations

### LINKED ISSUE #290: Improve performance compared to the current collections
In general the strawman is slower than the current collections (~1.5x, in general). We should identify the reasons of this slowdown and fix it.

julienrf:
Fixes #290.

The implementation of `size` was O(n) in `Vector` instead of O(1) (only `length` was O(1)), which was at the origin of the severe regression observed in #290.

I fixed that by making `size` a `final` forwarder to `length`. We could also remove `length` at all, as suggested in #147, however I found convenient to have `length` as an abstract member in `Indexed` collections so that I couldn’t forget to implement it.

I will post the benchmark results as soon as they have finished to run… (so, not before tomorrow…)

### DIFF
--- a/collections/src/main/scala/strawman/collection/ArrayOps.scala
+++ b/collections/src/main/scala/strawman/collection/ArrayOps.scala
@@ -2,7 +2,6 @@ package strawman
 package collection
 
 import scala.{AnyVal, Array, ArrayIndexOutOfBoundsException, Char, Int, throws}
-import scala.Predef.???
 import mutable.{ArrayBuffer, GrowableBuilder}
 
 import scala.reflect.ClassTag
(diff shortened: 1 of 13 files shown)


julienrf:
When running the benchmarks I’ve discovered some other cases where the strawman was slower than the current collections. I’ve added a couple of commits to fix these performance regressions. Here are the benchmarks now (on `List` and `Vector`):

access_find
access_init
access_last
access_random
access_slice
access_tail
create
expand_append
expand_appendall
expand_padto
expand_prepend
expand_prependall
transform_distinct
transform_groupby
transform_map
transform_patch
transform_reverse
transform_span
transform_unzip
transform_updatelast

As you can see the strawman has, in general, the same performance as the current collections, and sometimes better.

PR #318, opened 2017-12-10
Enhancements in the scalafix rule and tests

marcelocenerine:
- fix import of `jawn-ast` in the `benchmarks/memory` module which causes sbt errors when building project or importing it on IDE. `jawn-ast` is not available for Scala 2.13 (default)
- add rewrites for more collection types. Also, fix imports of types imported via full package name rather than via `scala.*` or `scala.Predef` (e.g.: `import scala.collection.immutable.Set`, `import scala.collection.Iterable`).
- remove global imports of `scala.collection.*` types to avoid clashes with counterparts in `strawman.collection.*`
- add more tests and move the input/output `object`s originally in `Collectionstrawman_v0` into separate files so that the rewrites for each collection type can be tested in isolation. This avoids that a symbol replacement for one collection type accidentally fixes the import for another collection type which is not correctly handled by the scalafix rule.


scalafix/input/src/main/scala/fix/Collectionstrawman_v0_BitSet.scala
```diff
@@ -0,0 +1,11 @@
+/*
+rule = "scala:fix.Collectionstrawman_v0"
+ */
+package fix
+
+import scala.collection.{BitSet, immutable, mutable}
+object Collectionstrawman_v0_BitSet {
+  val xs: BitSet = ???
```
marcelocenerine: `scala.collection.BitSet` has a companion object but `strawman.collection.BitSet` doesn't. Is it intentional or missing?

scalafix/input/src/main/scala/fix/Collectionstrawman_v0_IndexedSeq.scala
```diff
@@ -0,0 +1,12 @@
...
+rule = "scala:fix.Collectionstrawman_v0"
+ */
+package fix
+
+import scala.collection.immutable
+object Collectionstrawman_v0_IndexedSeq {
+  val xs: IndexedSeq[Int] = IndexedSeq(1, 2, 3)
+  val ys: collection.IndexedSeq[Int] = collection.IndexedSeq(1, 2, 3)
+  val ws: immutable.IndexedSeq[Int] = immutable.IndexedSeq(1, 2, 3)
+  //val zs: mutable.IndexedSeq[Int] = mutable.IndexedSeq(1, 2, 3)
```
marcelocenerine: a replacement for `scala.collection.mutable. IndexedSeq` does not exist in strawman. Is it intentional?

scalafix/input/src/main/scala/fix/Collectionstrawman_v0_LinearSeq.scala
```diff
@@ -0,0 +1,11 @@
+/*
+rule = "scala:fix.Collectionstrawman_v0"
+ */
+package fix
+
+import scala.collection.{LinearSeq, immutable}
+object Collectionstrawman_v0_LinearSeq {
+  val xs: LinearSeq[Int] = ???
+  val ys: immutable.LinearSeq[Int] = ???
+  //val zs: mutable.LinearSeq[Int] = ???
```
marcelocenerine: a replacement for `scala.collection.mutable.LinearSeq` does not exist in strawman. Is it intentional?

scalafix/input/src/main/scala/fix/Collectionstrawman_v0_Traversable.scala
```diff
@@ -3,11 +3,10 @@ rule = "scala:fix.Collectionstrawman_v0"
  */
 package fix
 
+import scala.collection.{immutable, mutable}
 object Collectionstrawman_v0_Traversable {
```
marcelocenerine: renamed to `Collectionstrawman_v0_Converters`

scalafix/input/src/main/scala/fix/ListTest.scala
```diff
@@ -1,10 +0,0 @@
-/*
-rule = "scala:fix.Collectionstrawman_v0"
-*/
-package fix
-
-import scala.concurrent.Future
-
-class ListTest {
-  def foo(x: Future[List[Int]]) = ???
-}
```
marcelocenerine: removed as similar test exist in `Collectionstrawman_v0_List`

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
```diff
@@ -58,12 +56,18 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
...
-  val rangeImport = Symbol("_root_.strawman.collection.immutable.Range.")
-  val inclusiveRange = Symbol(
-    "_root_.scala.runtime.RichInt#to(I)Lscala/collection/immutable/Range/Inclusive;.")
-  val rangeSymbol = Symbol(
-    "_root_.scala.runtime.RichInt#until(I)Lscala/collection/immutable/Range;.")
+  def removeGlobalImports(ctx: RuleCtx): Patch =
+    ctx.removeGlobalImport(Symbol("_root_.scala.collection.")) +
+      ctx.removeGlobalImport(Symbol("_root_.scala.collection.concurrent.")) +
+      ctx.removeGlobalImport(Symbol("_root_.scala.collection.immutable.")) +
+      ctx.removeGlobalImport(Symbol("_root_.scala.collection.mutable."))
```
marcelocenerine: even though the rule adds the imports for the collection types in strawman, the scala imports are still left in the source files. This basically remove all the global imports of scala's collections to avoid naming clashes. This is not perfect and needs improvement to handle imports such as:

`import collection._`
or 
```
import scala.collection
import collection.mutable
```

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
```diff
@@ -101,31 +141,40 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
       i("`+:`"),
       s("`:+`"),
       i("`:+`"),
-      i("Stream", Some("LazyList")),
       s("Stream", Some("LazyList")),
-      s("Traversable", Some("Iterable")),
```
marcelocenerine: I believe the one above was incorrect. It rewrites `scala.collection.Traversable` to `strawman.collection.immutable.Iterable`. I assume the correct import would be `strawman.collection.Iterable`

scalafix/input/src/main/scala/fix/Collectionstrawman_v0_BitSet.scala
julienrf: That’s an oversight, thanks for reporting :)

scalafix/input/src/main/scala/fix/Collectionstrawman_v0_IndexedSeq.scala
julienrf: That’s an oversight, thanks for reporting :)

scalafix/input/src/main/scala/fix/Collectionstrawman_v0_LinearSeq.scala
julienrf: We currently have no implementation of a `mutable.LinearSeq` in the strawman.

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
julienrf: Indeed, I thought `scala.Traversable[A] = scala.collection.immutable.Traversable[A]` but I was wrong.

julienrf:
Hey @marcelocenerine, thanks a lot for helping on this :) However I was thinking that we should wait for the new collections to be part of the `scala` namespace (instead of the current `strawman` namespace) before putting too much effort in the migration rules because it seems that 80% of the work is about re-arranging the imports to target `strawman.…` instead of `scala.…`.

What do you think?

scalafix/rules/src/main/scala/fix/Collectionstrawman_v0.scala
marcelocenerine: after thinking a bit more about this, maybe a better approach would be to rewrite the imports to the corresponding packages in strawman. Then in `replaceSymbols` instead of having all collection types in there we would keep only the special cases (like predef, LazyList, Traversable, ...). This would likely avoid that local imports become global imports. I can dig more on this before proceeding with this PR

marcelocenerine:
@julienrf, that's true :D. Most of the work is to re-arrange the imports. Let's wait then. I'm gonna close this PR

PR #319, opened 2017-12-11
Add dedupe(By) operations to collections-contrib

### LINKED ISSUE #310: Consider adding dedupe to Seq
Is there place for a `Seq.dedupe` method in the new collections library?

It's similar to `distinct` but removes only consecutive duplicates. I find myself needing this from time to time. For reference, Clojure has it: https://clojuredocs.org/clojure.core/dedupe.

joroKr21:
Resolves #310


Ichoran:
Sorry for bikeshedding, but `dedupe` is a confusing name (not to mention abbreviated where most method names are not).  What about `noRepeats` or `noSequentialRepeats` or `slidingDistinct` or `destutter` or really anything that won't be confused with `distinct` (i.e. a method that removes all duplicates)?

julienrf:
Hey @joroKr21, thanks for this work. I’m not sure this method is used by enough people to be here, though.

joroKr21:
@Ichoran good point. I went with `dedupe` because that's the name I have used this function under. From your suggestions I like `slidingDistinct` the most. The rest are kind of awkward. I wouldn't want to start a method name with `no` and `destutter` is a word I hear for the first time. But, if there is no place here for this operation anyway, we don't have to think of a good name :smile: 

@julienrf It's pretty much a copy-paste of `distinct` but using a single var instead of hash set. Hard for me to say how many people use something like it. It's mostly useful for sorted seqs or lazy seqs (in cases where keeping a whole hash set around is a no-go).

PR #321, opened 2017-12-13
Make `collection-contrib` and `junit` compile with Dotty

allanrenucci:


test/junit/src/test/scala/strawman/collection/LazyZipOpsTest.scala
```diff
@@ -64,7 +65,7 @@ class LazyZipOpsTest {
 
   @Test
   def lazyZip2_toIterable(): Unit = {
-    val iter: Iterable[(Int, Int)] = zipped2
+    val iter: collection.Iterable[(Int, Int)] = zipped2
```
julienrf: Why was this change needed?

test/junit/src/test/scala/scala/collection/mutable/MutableListTest.scala
```diff
@@ -1,37 +0,0 @@
...
-    val it = is.iterator
-    val x  = Some(3)
-    is += x
-    assertNotReachable(x, it) {
-      it.next()
-      it.next()
-    }
-    assertTrue(it.isEmpty)
-  }
-}
```
julienrf: Why did you have to remove this file?

allanrenucci: I removed:
```scala
def assertSameElements[A, B >: A](expected: IterableLike[A, _], actual: GenIterable[B], message: String = ""): Unit
def assertSameElements[A, B >: A](expected: IterableLike[A, _], actual: Iterator[B]): Unit =
```
from `AssertUtil` which were causing ambiguous overloading issues in Dotty. This test relied on them.

test/junit/src/test/scala/strawman/collection/LazyZipOpsTest.scala
allanrenucci: This is due to a difference in naming resolution. `strawman.collection.immutable._` is imported so here Dotty resolve `Iterable` to `strawman.collection.immutable.Iterable` but `zipped2` returns `strawman.collection.LazyZip2` which is implicitly converted to `View` which is a subtype of `strawman.collection.Iterable` and not `strawman.collection.immutable.Iterable`

test/junit/src/test/scala/scala/collection/mutable/MutableListTest.scala
allanrenucci: Sure I can do that. I removed this file since it was testing something that was not part of the strawman collection anyway.

julienrf: Yes, you are right but I prefer keeping these files so that we immediately see what has been migrated to the strawman and what hasn’t.

PR #329, opened 2017-12-23
Rename unordered to unsorted.

### LINKED ISSUE #324: Rename unordered to unsorted
The `unordered` operation upcasts a `SortedThing` into a `Thing` (forgetting that it is sorted). I think we should use the name “unsorted” rather than “unordered”, because we use the “Sorted” prefix.

jiminhsieh:
Fix #324.


julienrf:
Thanks @jiminhsieh! We should apply the change also on `SortedSet` and `SortedMap`, in the core module.

jiminhsieh:
@julienrf Thanks for pointing out! But I didn't find the method named `unordered` in `SortedSet` and `SortedMap`. However, I did find this method at [this line](https://github.com/scala/collection-strawman/blob/ecd97a8a7dac3b218b32cbbf2911413662e33287/collections/src/main/scala/strawman/collection/immutable/BitSet.scala#L56) in BitSet and [this line](https://github.com/scala/collection-strawman/blob/ecd97a8a7dac3b218b32cbbf2911413662e33287/collections/src/main/scala/strawman/collection/immutable/TreeSet.scala#L113) in `TreeSet`. About those 2 methods, I haven't had any good idea that should I change or not. If you actually mean there are methods need to be changed in the core collection of SortedSet and SortedMap, but not the **exact** naming, I will try to find out.

julienrf:
We should pull up the method definition to SortedSet and SortedMap, if possible, and rename it to "unsorted". Basically, every sorted collection (so, anything that inherits from SortedSet or SortedMap) should have an "unsorted" operation.

collections/src/main/scala/strawman/collection/SortedMap.scala
```diff
@@ -10,7 +10,9 @@ import scala.{Boolean, Int, Option, Ordering, PartialFunction, Serializable, `in
 /** Base type of sorted sets */
 trait SortedMap[K, +V]
   extends Map[K, V]
-    with SortedMapOps[K, V, SortedMap, SortedMap[K, V]]
+    with SortedMapOps[K, V, SortedMap, SortedMap[K, V]] {
+  def unsorted: Map[K, V]
```
julienrf: Why don’t you implement the operation directly here?

collections/src/main/scala/strawman/collection/SortedSet.scala
```diff
@@ -4,7 +4,9 @@ import scala.{Boolean, Ordering, `inline`, None, Option, Some}
 import scala.annotation.unchecked.uncheckedVariance
 
 /** Base type of sorted sets */
-trait SortedSet[A] extends Set[A] with SortedSetOps[A, SortedSet, SortedSet[A]]
+trait SortedSet[A] extends Set[A] with SortedSetOps[A, SortedSet, SortedSet[A]] {
+  def unsorted: Set[A]
```
julienrf: Same here: why not implementing the operation here?

collections/src/main/scala/strawman/collection/SortedMap.scala
jiminhsieh: I got it. Thank you for pointing out! :) I will push another commit.

julienrf:
Thanks!

PR #336, opened 2018-01-10
A port of mutable.Queue from the old collections

szeiger:
This includes mutable.LinkedList and mutable.MutableList. They were
already deprecated in 2.12 and are now package-private. We should
consider simplifying the implementation of Queue so that they can be
removed entirely in the future.


julienrf:
I’m not sure it’s worth adding a package private collection and remove it soon.

I would rather get `mutable.Queue` from #49. What do you think?

szeiger:
Switching to ArrayDeque is probably a good idea for a Queue replacement. I made the helper classes package-private so we could easily remove them when changing the implementation without concern for compatibility.

szeiger:
I'm merging this because I have further dependencies on mutable.Queue in my integration branch. We should still replace it with #49 before 2.13.0-M4.

PR #351, opened 2018-01-18
Bring back WrappedString

szeiger:
The motivation is the same as for Arrays which were originally
wrapped in ArrayView but we later changed it back to WrappedArray
(and ImmutableArray): It is very convenient to have an implicit
conversion of a String to a Seq[Char] but a View is no longer a Seq
in the new collection library.

### DIFF
--- a/collections/src/main/scala/strawman/collection/immutable/WrappedString.scala
+++ b/collections/src/main/scala/strawman/collection/immutable/WrappedString.scala
@@ -1,0 +1,59 @@
+package strawman.collection
+package immutable
+
+import scala.{Char, Int}
+import scala.Predef.String
+import mutable.{Builder, StringBuilder}
+
+/**
+  *  This class serves as a wrapper augmenting `String`s with all the operations
+  *  found in indexed sequences.
+  *
+  *  The difference between this class and `StringOps` is that calling transformer
+  *  methods such as `filter` and `map` will yield an object of type `WrappedString`
+  *  rather than a `String`.
+  *
+  *  @param self    a string contained within this wrapped string
+  *
+  *  @since 2.8
+  *  @define Coll `WrappedString`
+  *  @define coll wrapped string
+  */
+final class WrappedString(val self: String) extends AbstractSeq[Char] with IndexedSeq[Char] {
+
+  def apply(i: Int): Char = self.charAt(i)
+
+  protected[this] def fromSpecificIterable(coll: strawman.collection.Iterable[Char]): IndexedSeq[Char] =
+    WrappedString.fromSpecific(coll)
+  protected[this] def newSpecificBuilder(): Builder[Char, IndexedSeq[Char]] = WrappedString.newBuilder()
+  def iterableFactory: SeqFactory[IndexedSeq] = ImmutableArray
+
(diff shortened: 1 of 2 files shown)


collections/src/main/scala/strawman/collection/immutable/WrappedString.scala
```diff
@@ -0,0 +1,63 @@
...
+  *  methods such as `filter` and `map` will yield an object of type `WrappedString`
+  *  rather than a `String`.
+  *
+  *  @param self    a string contained within this wrapped string
+  *
+  *  @since 2.8
+  *  @define Coll `WrappedString`
+  *  @define coll wrapped string
+  */
+final class WrappedString(val self: String) extends AbstractSeq[Char] with IndexedSeq[Char] {
```
julienrf: Maybe it’s time to have `AbstractIndexedSeq`?

collections/src/main/scala/strawman/collection/immutable/WrappedString.scala
```diff
@@ -0,0 +1,63 @@
...
+  */
+object WrappedString extends SpecificIterableFactory[Char, WrappedString] {
+  def fromSpecific(it: IterableOnce[Char]): WrappedString = {
+    val b = newBuilder()
+    it match {
+      case it: Iterable[_] =>
+        val s = it.knownSize
+        if(s >= 0) b.sizeHint(s)
+      case _ =>
+    }
```
julienrf: Note that `IterableOnce` *has* `knownSize` so you don’t have to type test for `Iterable`

collections/src/main/scala/strawman/collection/immutable/WrappedString.scala
szeiger: This is the same as in the old collections. There's `collection.AbstractSeq` but no `collection.AbstractIndexedSeq` or `immutable.Abstract*`. I'll do some more experiments and possibly open another PR. First tests show that `immutable.AbstractSeq` could be worth it, but `immutable.AbstractIndexedSeq` is probably not.

collections/src/main/scala/strawman/collection/immutable/WrappedString.scala
```diff
@@ -0,0 +1,59 @@
...
+  *  @define coll wrapped string
+  */
+final class WrappedString(val self: String) extends AbstractSeq[Char] with IndexedSeq[Char] {
+
+  def apply(i: Int): Char = self.charAt(i)
+
+  protected[this] def fromSpecificIterable(coll: strawman.collection.Iterable[Char]): IndexedSeq[Char] =
+    WrappedString.fromSpecific(coll)
+  protected[this] def newSpecificBuilder(): Builder[Char, IndexedSeq[Char]] = WrappedString.newBuilder()
+  def iterableFactory: SeqFactory[IndexedSeq] = ImmutableArray
```
julienrf: Why not use the default `IndexedSeq` implementation?

szeiger: Being a fixed-size array it's the most similar to `String` in spirit

PR #352, opened 2018-01-19
Create two distinct scalafix rules targeting 2.12 and 2.13

julienrf:
I created two subdirectories, `scalafix/2.12` and `scalafix/2.13` to support two kinds of rules: (1) migration from a 2.12 code base to a 2.12 code base using the strawman as a library, and (2) migration of a 2.12 code base to a 2.13 code base (where the standard collection *is* the strawman).

A consequence is that I can not anymore use the nice `github` scheme in scalafix URLs (see [here](https://github.com/scala/collection-strawman/pull/352/files#diff-04c6e90faac2675aa89e2176d2eec7d8L64)) but the 2.12 and 2.13.

cc @olafurpg.

### DIFF
--- a/scalafix/2.13/rules/src/main/scala/fix/Collectionstrawman_v0.scala
+++ b/scalafix/2.13/rules/src/main/scala/fix/Collectionstrawman_v0.scala
@@ -86,9 +86,20 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
         removeTokensPatch + replaceCommasPatch
     }.asPatch
 
+  val copyToBuffer = SymbolMatcher.normalized(
+    Symbol("_root_.scala.collection.TraversableOnce.copyToBuffer.")
+  )
+
+  def replaceCopyToBuffer(ctx: RuleCtx): Patch =
+    ctx.tree.collect {
+      case t @ q"${copyToBuffer(Term.Select(collection, _))}($buffer)" =>
+        ctx.replaceTree(t, q"$buffer ++= $collection".syntax)
+    }.asPatch
+
   override def fix(ctx: RuleCtx): Patch = {
     replaceToList(ctx) +
       replaceSymbols(ctx) +
-      replaceTupleZipped(ctx)
+      replaceTupleZipped(ctx) +
+      replaceCopyToBuffer(ctx)
   }
 }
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -129,6 +136,7 @@ val scalacheck = project.in(file("test") / "scalacheck")
    // Dotty 0.3.0-RC1 crashes when trying to compile this project
   .settings(disableDotty)
   .settings(
+    collectionsScalaVersionSettings,
     fork in Test := true,
     javaOptions in Test += "-Xss1M",
     libraryDependencies ++= Seq(
```
julienrf: Also, a side-effect of the changes in this file is that the project can be imported without errors in IntelliJ.

olafurpg:
What do you think would be the most convenient to support different directories for rules without missing `github:` scheme support? For example, we could read a URL parameter `github:scala/collection-strawman?id=2.13` and that would use `2.13` instead of the default `rules` directory.

julienrf:
I wouldn’t put in scalafix something that specific to the collection-strawman project. Also, I’m not sure I will keep the 2.12 rewrite because it is super hard to get right (because of we have to unimport several default imports)

scalafix/2.13/rules/src/main/scala/fix/Collectionstrawman_v0.scala
```diff
@@ -86,9 +86,20 @@ case class Collectionstrawman_v0(index: SemanticdbIndex)
         removeTokensPatch + replaceCommasPatch
     }.asPatch
 
+  val copyToBuffer = SymbolMatcher.normalized(
+    Symbol("_root_.scala.collection.TraversableOnce.copyToBuffer.")
+  )
+
+  def replaceCopyToBuffer(ctx: RuleCtx): Patch =
+    ctx.tree.collect {
+      case t @ q"${copyToBuffer(Term.Select(collection, _))}($buffer)" =>
```
olafurpg: Pro tip, you can put the `collection` qualifier like this to avoid Term.Select
```scala
@ val q"$qual.${q"foo"}(..$arg)" = q"qual.foo(1, 2)"
qual: Term = Term.Name("qual")
arg: List[Term] = List(Lit.Int(1), Lit.Int(2))

```

People have different tastes however, I personally prefer Term.Select :)

Issue #338, opened 2018-01-11
Improve usability of lengthCompare


joroKr21:
In most codebases I've seen people just use `seq.length/size <> n` even though `seq.lengthCompare(n) <> 0` is recommended for sequential collections. In some cases it's probably for lack of knowledge, but in others the latter is just deemed too unreadable. 

My suggestion is to add some shortcuts that make it more pleasant to use.
Something along those lines:
```scala
def lengthEq(n: Int):  Boolean = lengthCompare(n) == 0
def lengthMin(n: Int): Boolean = lengthCompare(n) >= 0
def lengthMax(n: Int): Boolean = lengthCompare(n) <= 0
```

Other suggestions also welcome, of course.

This particular issue came up on Gitter.

joroKr21:
Link to chat: https://gitter.im/scala/collection-strawman?at=5a55f0fcd0514c785b120ece

SethTisue:
maybe `lengthIs`, `lengthAtLeast`, `lengthAtMost`

sjrd:
`length_==`, `length_>=`, `length_<=`

jtjeferreira:
what about adding `seq.lengthCompare(anotherSeq): Int`

isaias-b:
What about this kind of interface? 
```scala
  implicit class SeqLikeOps[+A, +Repr](s: SeqLike[A, Repr]) {
    object lazyLengh {
      def <(value:  Int) = s.lengthCompare(value) < 0
      def >(value:  Int) = s.lengthCompare(value) > 0
      def ==(value:  Int) = s.lengthCompare(value) == 0
      def <=(value:  Int) = s.lengthCompare(value) <= 0
      def >=(value:  Int) = s.lengthCompare(value) >= 0
    }
  }
```

That allows to use:
```scala
val s: Seq[String] = ???
s.lazyLength < 10
s.lazyLength == 10
s.lazyLength > 10
```

However, i am not a DSL expert maybe this can be implemented without an extra inner `object`?

joroKr21:
No, but the usability is almost the same (modulo `_`)
> Oh is this suggestion doing the exact kind of thing? sweat_smile #338 (comment)

aleks-sidorenko:
`lengthEquals, lengthMoreThan, lengthLessThan`

NthPortal:
Are most people happy (or at least okay) with Sébastien's suggestion?

SethTisue:
I like Sébastien's suggestion best.

[...]

PR #356, opened 2018-01-22
Standardize on .size (but still offer .length)

### LINKED ISSUE #147: length vs size
Do we need both `length` and `size`? I’d rather just keep `size`.

julienrf:
Fixes #147.

The idea is to offer to users only one way to get the size of a collection: the `size` operation. Therefore, `length` is deprecated and `final` (it just calls `size`).

That being said, I found it convenient for indexed collections to have an abstract method giving the size so that implementors *have to* implement it instead of having to *remember* of overriding the `size` operation (whose default implementation is O(n)). Previously, indexed collections were using the `length` method in that way. I replaced it with a protected `finiteSize` method.

### DIFF
--- a/collections/src/main/scala/strawman/collection/ArrayOps.scala
+++ b/collections/src/main/scala/strawman/collection/ArrayOps.scala
@@ -33,7 +33,7 @@ class ArrayOps[A](val xs: Array[A]) extends AnyVal
   protected[this] def coll: Array[A] = xs
   override def toSeq: immutable.Seq[A] = fromIterable(toIterable)
 
-  def length = xs.length
+  protected def finiteSize = xs.length
   @throws[ArrayIndexOutOfBoundsException]
   def apply(i: Int) = xs.apply(i)
 
--- a/collections/src/main/scala/strawman/collection/IndexedSeq.scala
+++ b/collections/src/main/scala/strawman/collection/IndexedSeq.scala
@@ -13,12 +13,16 @@ trait IndexedSeq[+A] extends Seq[A] with IndexedSeqOps[A, IndexedSeq, IndexedSeq
 object IndexedSeq extends SeqFactory.Delegate[IndexedSeq](immutable.IndexedSeq)
 
 /** Base trait for indexed Seq operations */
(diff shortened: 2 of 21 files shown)


julienrf:
Good point! I didn’t change `lengthCompare` but I agree that it would make more sense to rename it.

lrytz:
I support this change.

julienrf:
About `lengthCompare`: I’m fine with leaving as it is and introducing `size_==`, `size_<=` and `size_>=` operations as suggested in #338.

lrytz:
If `length` is deprecated then `lengthCompare` should be as well, IMO.

Ichoran:
I agree--you have to deprecate `lengthCompare` if you deprecate `length`.  Also, I dislike partially symbolic method names.  Something like `sizeIs`, `sizeLEq`, `sizeGEq` would be better.  (Or the wordier `sizeAtLeast`, `sizeNoMoreThan`.)

julienrf:
OK to deprecate `lengthCompare` in favor of a `sizeCompare`!

SethTisue:
I support this change as well. Julien, were you able to convince Martin?

odersky:
No, I still think we should leave `length` alone. It's one of the strengths of Scala that arrays and strings are just normal collections. But if we remove `length` they stop being that! Compared to this having two common and well-understood names for the same operation seems like a small price to pay. So by extension, I would also leave `lengthCompare` alone, and not try to add new variants.

julienrf:
I updated the PR to not deprecate `length`. It is defined as an alias (`inline` and `final`) for `size`, in `SeqOps`.

szeiger:
This should be done the other way around. Every single collection implementation needs to be changed to override `size` instead of `length`, causing compatibility problems. Previously `size` was an alias for `length`.

szeiger:
After doing the revert locally and looking for possible changes that should be done to ensure `size` is always in sync with `length`, I realized that everything was already done correctly: `Seq` defines `size` as a final alias for `length`. It may be useful for linear Seqs not to do this so that you get a decent default behavior based on `iterator()` but given that `size` and `length` are such essential operations and there should always be a faster way to implement them (even if just to take the `Iterator` out of the loop) we can probably live with that. `IndexedSeq` further defines `knownSize` as a final alias for `length`.

PR #357, opened 2018-01-22
Add deprecated CanBuildFrom alias to BuildFrom

julienrf:
That would allow a small subset of existing code to be source compatible with the new collections.

Typically, the following will cross compile:

~~~ scala
def foo[F[_], A, T](from: F[A])(implicit cbf: CanBuildFrom[F[A], A, T]) = {
  val builder = cbf(from)
  …
  builder.result
}
~~~

(Unfortunately, my guess is that most usages of `CanBuildFrom` will require adaptation, but at least that one would be compatible…)


lrytz:
Could you explain the use case / link to discussions? Is it about extension methods?

julienrf:
Sure, I’ve just added a description.

PR #360, opened 2018-01-24
Make sure the code is compiled by Dotty

julienrf:
Currently the CI doesn’t compile the code with Dotty. This PR fixes this issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: scala
 jdk: oraclejdk8
 script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,7 +1,8 @@
 language: scala
 jdk: oraclejdk8
 script:
-  - sbt ++2.12.4 memoryBenchmark/compile +test ++2.13.0-M2 timeBenchmark/compile collections-contribJVM/test junit/test scalacheck/test
```
julienrf: `+test` doesn’t do what is supposed to do because `crossScalaVersions` is scoped to a project, not to the build. As a workaround, I explicitly set with which scala version we should compile the project.

PR #345, opened 2018-01-17
Fix laziness test for LazyList

julienrf:
Fixes #308.

I’ve also renamed `lazyAppendAll` to `lazyAppendedAll` but the important part of the PR is in the test file.

The new test passes only on 2.13, though, not on 2.12 nor dotty.

PR #361, opened 2018-01-24
Generalize LazyList to allow implementation of Stream

szeiger:
Since the semantics are subtly different we should add a deprecated `Stream` with the expected semantics rather than aliasing to `LazyList`

### DIFF
--- a/collections/src/main/scala/strawman/collection/immutable/LazyList.scala
+++ b/collections/src/main/scala/strawman/collection/immutable/LazyList.scala
@@ -7,6 +7,7 @@ import strawman.collection.mutable.{ArrayBuffer, Builder}
 import scala.{Any, AnyRef, Boolean, Int, None, NoSuchElementException, noinline, Nothing, Option, PartialFunction, Some, StringContext, Unit, UnsupportedOperationException, deprecated}
 import scala.Predef.String
 import scala.annotation.tailrec
+import scala.annotation.unchecked.uncheckedVariance
 
 /**  The class `LazyList` implements lazy lists where elements
   *  are only evaluated when they are needed. Here is an example:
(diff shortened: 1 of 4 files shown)


collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -261,10 +225,52 @@ sealed abstract class LazyList[+A]
     if (this.isEmpty) z
     else tail.foldLeft(op(z, head))(op)
   }
+}
+
+trait LazyListOps[+A, +CC[+X] <: LinearSeq[X] with LazyListOps[X, CC, CC[X]], +C <: CC[A]] extends LinearSeqOps[A, CC, C] {
```
julienrf: What do you think of making this trait package private so that we won’t have to support binary compatibility for it?

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -434,84 +523,136 @@ object LazyList extends SeqFactory[LazyList] {
...
-      // of our constructed lazy list is not evaluated (e.g. when one calls `drop`).
-      lazy val evaluatedElem = it.next()
-      new LazyList.Cons(evaluatedElem, { evaluatedElem; fromIterator(it) })
-    } else LazyList.Empty
+  if (it.hasNext) {
+    // Be sure that `it.next()` is called even when the `head`
+    // of our constructed lazy list is not evaluated (e.g. when one calls `drop`).
+    lazy val evaluatedElem = it.next()
+    new LazyList.Cons(evaluatedElem, { evaluatedElem; fromIterator(it) })
+  } else LazyList.Empty
```
julienrf: The reindentation looks wrong.

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -434,84 +523,136 @@ object LazyList extends SeqFactory[LazyList] {
...
+    * @param z The initial value seeded into the function `op`.
+    * @param op The operation to perform on successive elements of the `LazyList`.
+    * @return The accumulated value from successive applications of `op`.
+    */
+  @tailrec
+  override final def foldLeft[B](z: B)(op: (B, A) => B): B = {
+    if (this.isEmpty) z
+    else tail.foldLeft(op(z, head))(op)
+  }
+}
```
julienrf: Why isn’t it possible to put `foreach` and `foldLeft` in `LazyListOps`?

julienrf:
And we should also remove #345 if we merge this one. (I’d like to have `lazyAppendAll` renamed to `lazyAppendedAll`, though, for consistency with `appendedAll` in `Seq`, do you mind applying this change here too?)

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
szeiger: They can't be tail-recursive because the types of `this` and `tail` can be different.

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
szeiger: Still needs to be binary-compatible. Types that implement methods leak out.

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
szeiger: And implementing `foreach` with a `while` loop doesn't work, either, because you cannot drop the `this` reference manually.

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
julienrf: You’re right it’s not really about binary compatibility, but more about not exposing a type that we don’t want people to extend so that we can remove it if it happens to be unnecessary (after we remove `Stream`).

szeiger:
I'm changing the tests to use method call syntax. We want to test the behavior of the implementation after all, not the state of lazy operator desugaring. I'm also swapping out the old `Stream` for the new one.

collections/src/test/scala/strawman/collection/test/Test.scala
```diff
@@ -455,7 +455,7 @@ class StrawmanTest {
 
     // laziness may differ in dotty, so test only that we are as lazy as Stream
     import scala.collection.{immutable => old}
-    lazy val fibsStream: old.Stream[Int] = 0 #:: 1 #:: fibsStream.zip(fibsStream.tail).map { n => n._1 + n._2 }
+    lazy val fibsStream: Stream[Int] = 0 #:: 1 #:: fibsStream.zip(fibsStream.tail).map { n => n._1 + n._2 }
     if(old.List(0,1,1,2)==fibsStream.take(4).toList) {
```
julienrf: This will always return `false`

collections/src/test/scala/strawman/collection/test/Test.scala
```diff
@@ -465,7 +465,7 @@ class StrawmanTest {
     var lazeCountL = 0
     def lazeL(i: Int) = {lazeCountL += 1; i}
     def lazeS(i: Int) = {lazeCountS += 1; i}
-    val xs20 = lazeS(1) #:: lazeS(2) #:: lazeS(3) #:: old.Stream.empty
+    val xs20 = lazeS(1) #:: lazeS(2) #:: lazeS(3) #:: Stream.empty
     val xs21 = lazeL(1) #:: lazeL(2) #:: lazeL(3) #:: LazyList.empty
     assert(lazeCountS==lazeCountL)
```
julienrf: Since `Stream` and `LazyList` don’t have the same laziness level this assertion will never be fulfilled.

szeiger:
@julienrf What did I overlook here? Why did the tests succeed for the old Stream implementation? The new one should be the same (i.e. different from LazyList)

julienrf:
Because that test used to be affected by the evaluation rules of right-associative by-name operators (this was giving `LazyList` the same level of laziness as `Stream`).

szeiger:
Oh, those are not the reason for the test failure. This PR doesn't change Scala versions, so the new Stream should behave exactly the same as the old one. I suggest doing the remaining changes in #345 where you switch to a newer 2.13 build.

julienrf:
I agree.

(also it seems that there is a variance related issues with Dotty)

szeiger:
The variance error seems wrong. We need an `@uncheckedVariance` annotation on the method because it has to be package-private but the class is `private[this]` which should be fine. After fixing that I'm still running into another Dotty compiler bug. Why doesn't the push job run on Dotty though? The pr job was also green in the beginning which makes me wonder if one of my trivial changes in the 2nd push made it fail on Dotty or if it was never tested in the first place.

szeiger:
OK, so the push build doesn't use Dotty because the PR that enables Dotty in PR validation was not in my branch yet. It also wasn't merged when I pushed the original version of this PR which explains why the first version passed validation. I'll open a bug report against Dotty for the new issue.

szeiger:
Updated to disable smoke tests for LazyList so we can compile on Dotty again

PR #364, opened 2018-01-25
Some more things to improve backward compatibility

julienrf:
When cross-compiling better-files I noticed a few things that were missing:

- an overload of `collect` returning a `Map` in `MapOps`,
- deprecated `toIterator` operation in `IterableOnce`.

### DIFF
--- a/collections/src/main/scala/strawman/collection/IterableOnce.scala
+++ b/collections/src/main/scala/strawman/collection/IterableOnce.scala
@@ -55,23 +55,23 @@ final class IterableOnceExtensionMethods[A](private val it: IterableOnce[A]) ext
   def toList: immutable.List[A] = immutable.List.from(it)
 
   @deprecated("Use Set.from(it) instead of it.toSet", "2.13.0")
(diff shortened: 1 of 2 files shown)


collections/src/main/scala/strawman/collection/IterableOnce.scala
```diff
@@ -55,24 +55,24 @@ final class IterableOnceExtensionMethods[A](private val it: IterableOnce[A]) ext
   def toList: immutable.List[A] = immutable.List.from(it)
 
   @deprecated("Use Set.from(it) instead of it.toSet", "2.13.0")
-  @`inline` final def toSet[B >: A]: immutable.Set[B] = immutable.Set.from(it)
+  @`inline` def toSet[B >: A]: immutable.Set[B] = immutable.Set.from(it)
```
lrytz: Why did you un-finalize these methods? Note that `@inline` only works on final methods.

julienrf: Because the outer class is already final.

PR #366, opened 2018-01-25
Introduce a deprecated MultiMap implementation

julienrf:
This change can be controversial.

In 2.12 we have a `mutable.MultiMap` trait (see API [here](http://scala-lang.org/api/current/scala/collection/mutable/MultiMap.html)).

In 2.13 I introduced a `mutable.MultiMap` collection that has the _same name_ but a _different usage_ (see API [here](https://static.javadoc.io/ch.epfl.scala/collections-contrib_2.12/0.8.0/strawman/collection/mutable/MultiMap.html)). Thus, these two `MultiMap` are incompatible. One solution could have been to name the new `MultiMap` differently but … it really is a `MultiMap` so I’m not sure we should use a different name… Also, I think I should remind why in the first place I decided to go with different approach for `MultiMap`: I think the existing `MultiMap` is confusing and doesn’t provide the level of support that users would expect. Indeed when you use it it is very easy to mistakenly call the wrong `update` operation (instead of `addBinding`) and it doesn’t work with sorted maps and lacks transformation operations (it only provides three operations).

So, this PR brings solutions to help migration and to support cross-compatibility between Scala versions.

---

For users upgrading from 2.12 to 2.13, the scalafix migration rule will replace `collection.mutable.MultiMap` (in 2.12) with `collection.mutable.DeprecatedMultiMap` (in 2.13), which is a copy of the existing `MultiMap`, but deprecated.
Their code will still compile, with a warning inviting them to migrate to the new `collection.mutable.MultiMap` type.

---

For users who want their project to compile for _both_ 2.12 and 2.13, they will have to add the following import:

~~~ scala
import collection.compat.mutable.MultiMap
~~~

This import is an alias to `collection.mutable.MultiMap` in 2.12 and to `collection.mutable.DeprecatedMultiMap` in 2.13. Note that people will have to add a dependency on an extra `collections-compat` module (that we might have to introduce anyway, to support the `to` operation in cross-compatible way).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,15 +5,19 @@ import org.scalajs.sbtplugin.cross.CrossProject
 val dotty = settingKey[String]("dotty version")
 dotty in ThisBuild := "0.6.0-RC1"
 
+val scala211 = "2.11.12"
+val scala212 = "2.12.4"
+val scala213 = "2.13.0-M2"
+
 val collectionsScalaVersionSettings = Seq(
-  scalaVersion := "2.13.0-M2",
-  crossScalaVersions := scalaVersion.value :: "2.12.4" :: dotty.value :: Nil
+  scalaVersion := scala213,
+  crossScalaVersions := scala213 :: scala212 :: dotty.value :: Nil
 )
 
 val commonSettings = Seq(
   organization := "ch.epfl.scala",
   version := "0.9.0-SNAPSHOT",
-  scalaVersion := "2.12.4",
+  scalaVersion := scala212,
   scalacOptions ++= Seq("-deprecation", "-feature", "-unchecked", "-language:higherKinds"/*, "-opt:l:classpath"*/),
   scalacOptions ++= {
     if (!isDotty.value)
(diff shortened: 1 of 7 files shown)


lrytz:
I think we'd better go the conservative route: have `MultiMap` behave the same as in 2.12, but deprecated. Introduce the new version as `NewMultiMap` (?), possibly with an alias just named `MultiMap` somewhere. In the future, once the deprecated version is removed in version N, we can move the new implementation to `mutable.MultiMap` in N+1.

julienrf:
I understand the argument of being conservative.

Now, which name could be use instead of `MultiMap`? The problem is that we have several `MultiXxx` collections in the contrib module: `MultiMap`, `MultiSet`, `SortedMultiMap` and `SortedMultiSet`, in generic, mutable and immutable versions… The only one that clashes with the old collections is `mutable.MultiMap`, but if I rename only this one, this is not going to be consistent with the other `MultiXxx` collections of the family. What do you think of `MultiDict` instead of `MultiMap`? According to [wikipedia](https://en.wikipedia.org/wiki/Multimap) this is an alternative name…

PR #368, opened 2018-01-25
Address warnings

julienrf:
Our own code base were raising a few warnings. This PR addresses them.

### DIFF
--- a/collections/src/main/scala/strawman/collection/convert/Wrappers.scala
+++ b/collections/src/main/scala/strawman/collection/convert/Wrappers.scala
@@ -10,13 +10,14 @@ package strawman
 package collection
 package convert
 
-import scala.{Some, None, Serializable, AnyRef, Boolean, Option, Int, ClassCastException, SerialVersionUID, Any, Unit}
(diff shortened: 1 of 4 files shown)


collections/src/main/scala/strawman/collection/convert/Wrappers.scala
```diff
@@ -426,7 +427,7 @@ private[collection] trait Wrappers {
 
     def iterator() = enumerationAsScalaIterator(underlying.keys) map (k => (k, underlying get k))
 
-    override def clear() = underlying.clear()
+    def clear() = iterator().foreach(entry => underlying.remove(entry._1))
```
julienrf: Interestingly this made me find this bug: we were previously calling `underlying.clear()`, whereas the `underlying` value (of type `java.util.Dictionary`) had no such method. It was implicitly converted to the same Scala wrapper class, infinitely calling `clear()` again…

This bug is also present in 2.12.

SethTisue:
is enabling `-Xfatal-warnings` within reach?

julienrf:
@SethTisue We have at least one “expected” warning because `Traversable` is deprecated but the non-deprecated `Iterable` extends it.

PR #369, opened 2018-01-25
Introduce Cloneable trait

### LINKED ISSUE #365: Mutable collections should extend Cloneable like in 2.12
Found by https://github.com/scala/scala/blob/2.13.x/test/files/pos/array-interfaces.scala

julienrf:
Fixes #365

### DIFF
--- a/collections/src/main/scala/strawman/collection/mutable/Cloneable.scala
+++ b/collections/src/main/scala/strawman/collection/mutable/Cloneable.scala
@@ -1,0 +1,13 @@
+package strawman.collection.mutable
+
+import scala.AnyRef
+
+/** A trait for cloneable collections.
+  *
+  *  @since 2.8
+  *
+  *  @tparam A    Type of the collection, covariant and with reference types as upperbound.
+  */
+trait Cloneable[+A <: AnyRef] extends scala.Cloneable {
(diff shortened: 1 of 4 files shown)


lrytz:
There's also `mutable.Queue`, which currently extends plain `Cloneable` - this should probably just be removed, as `Queue` inherits from `mutable.SeqOps`.

collections/src/main/scala/strawman/collection/mutable/Cloneable.scala
```diff
@@ -0,0 +1,13 @@
+package strawman.collection.mutable
+
+import scala.AnyRef
+
+/** A trait for cloneable collections.
+  *
+  *  @since 2.8
+  *
+  *  @tparam A    Type of the elements contained in the collection, covariant and with reference types as upperbound.
```
hrhino: shouldn't this be `Type of the collection, ...`?

julienrf:
Good catch @hrhino, I copy-pasted the existing scaladoc without even checking that it was correct…

@lrytz `mutable.Queue` is not currently in master but, yes, we will have to remember to simplify its inheritance chain when we put it to master.

lrytz:
@julienrf I must have confused my intellij windows again :-)

Issue #226, opened 2017-09-05
Fix withDefault operation on Map


julienrf:
`TreeMap(1 -> "one").withDefault("NaN") + (2 -> "two")` doesn’t return a `SortedMap`

dhanesharole:
Is anyone working on this? Can I have a look?

dhanesharole:
@julienrf I have raised [PR](https://github.com/scala/collection-strawman/pull/381) for this issue. Can you please review it and give some feedback ?

PR #381, opened 2018-01-28
[226] Fix withDefault operations on SortedMap

dhanesharole:
This PR fixes [#226](https://github.com/scala/collection-strawman/issues/226).

Current implementation implements `withDefault` on `Map`. As a part of this fix, `SortedMap` now overrides `withDefault` method from base `Map` trait and provides the custom implementation of default value in `SortedMap.WithDefault` wrapper.


collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
```diff
@@ -3,52 +3,126 @@ package collection
...
+
+    override protected[this] def sortedMapFromIterable[K2, V2](it: collection.Iterable[(K2, V2)])(implicit ordering: Ordering[K2]): SortedMap[K2, V2] =
+      SortedMap.from(it)
+
+    override def iteratorFrom(start: K): Iterator[(K, V)] = underlying.iteratorFrom(start)
+
+    override def keysIteratorFrom(start: K): Iterator[K] = underlying.keysIteratorFrom(start)
+
+    override protected[this] def fromSpecificIterable(coll: collection.Iterable[(K, V)]): SortedMap[K, V] =
+      new WithDefault[K, V](SortedMap.from(coll), d)
```
julienrf: Instead of using the “default” `SortedMap` factory you should use a factory method from `underlying`:

~~~ scala
new WithDefault[K, V](underlying.fromSpecificIterable(coll), d)
~~~

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
```diff
@@ -3,52 +3,126 @@ package collection
...
+
+    override def iterableFactory: IterableFactoryLike[Iterable] = underlying.iterableFactory
+
+    override protected[this] def newSpecificBuilder(): Builder[(K, V), SortedMap[K, V]] = {
+      SortedMap.newBuilder[K, V]().mapResult(new WithDefault[K, V](_, d))
```
julienrf: Same here: `underlying.sortedMapFactory.newBuilder().mapResult(…)`

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
```diff
@@ -3,52 +3,126 @@ package collection
...
+    }
+  }
```
julienrf: I think you will also have to override some operations inherited from `SortedOps` (e.g. `rangeTo`).

test/junit/src/test/scala/strawman/collection/immutable/SortedMapTest.scala
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  @Test
+  def testDefaultValueIsPersistedWhenNewSortedMapIsCreatedFromSpecificIterableInOperationsLikeFilter(): Unit = {
+    val evenNumbers: SortedMap[Int, String] = SortedMap.from(Map(1 -> "One", 2 -> "Two", 4 -> "Four", 5 -> "Five"))
+      .withDefault(defaultValueFunction)
+      .filter((kv: (Int, String)) => kv._1 % 2 == 0)
+
+    assertEquals("5 is not present in this map", evenNumbers(5))
```
julienrf: Is that consistent with 2.11/2.12 behavior?

test/junit/src/test/scala/strawman/collection/immutable/SortedMapTest.scala
```diff
@@ -0,0 +1,94 @@
...
+      .withDefault(defaultValueFunction)
+
+    val newTreeMap = originalTreeMap - 3
+
+    assertEquals("3 is not present in this map", newTreeMap(3))
+    assertEquals(2, newTreeMap.size)
+  }
+
+  @Test
+  def testDefaultValueIsLostWhenNewSortedMapIsCreatedFromIterablesInOperationsLikeFlatMap(): Unit = {
```
julienrf: I’m wondering whether we should take advantage of the collections rewrite to fix that limitation (and to fix it in `Map.WithDefault` as well).

Do you think people rely on the fact that transformation operations such as `map` and `flatMap` do not preserve the default value? (cc @lrytz and @szeiger)

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
```diff
@@ -3,52 +3,126 @@ package collection
...
+  /** The same map with a given default value.
+    * Note: The default is only used for `apply`. Other methods like `get`, `contains`, `iterator`, `keys`, etc.
+    * are not affected by `withDefaultValue`.
+    *
+    * Invoking transformer methods (e.g. `map`) will not preserve the default value.
+    *
+    * @param d default value used for non-present keys
+    * @return a wrapper of the map with a default value
+    */
+  override def withDefaultValue[V1 >: V](d: V1): Map[K, V1] = new SortedMap.WithDefault[K, V1](this, _ => d)
```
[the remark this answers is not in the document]
dhanesharole: Fixed. It was a miss. Added a unit test with explicit type annotation on `val` assignment. [here] (https://github.com/dhanesharole/collection-strawman/blob/3ec1db989f9e9439945fa31d41b9dc935895a300/test/junit/src/test/scala/strawman/collection/immutable/SortedMapTest.scala#L93)

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
```diff
@@ -3,52 +3,126 @@ package collection
...
-      def rangeImpl(from: Option[K], until: Option[K]): SortedSet[K] = {
-        val map = self.rangeImpl(from, until)
-        new map.ImmutableKeySortedSet
-      }
-      def empty: SortedSet[K] = sortedIterableFactory.empty
-      def incl(elem: K): SortedSet[K] = fromSpecificIterable(this).incl(elem)
-      def excl(elem: K): SortedSet[K] = fromSpecificIterable(this).excl(elem)
-    }
+    with collection.SortedMapOps[K, V, CC, C] {
+  self =>
```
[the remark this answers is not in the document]
dhanesharole: Fixed. Removed unintentional code style changes and indentation changes in a separate commit. Will be careful of this from here on. Mistakenly, clicked `cmd+l` out of habit while refactoring on intellij. :)

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
```diff
@@ -3,52 +3,126 @@ package collection
...
+    result
+  }
+
 }
 
-object SortedMap extends SortedMapFactory.Delegate[SortedMap](TreeMap)
\ No newline at end of file
+object SortedMap extends SortedMapFactory.Delegate[SortedMap](TreeMap) {
+
+  final class WithDefault[K, +V](underlying: SortedMap[K, V], d: K => V) extends SortedMap[K, V] {
```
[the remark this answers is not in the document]
dhanesharole: Thinking on how to go about this. Current fixes don't include it. Will push changes related to it soon. I have been spiking how to do that. Will push my preliminary design changes in working code by EOD today.

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
```diff
@@ -3,52 +3,126 @@ package collection
...
+    override implicit def ordering: Ordering[K] = underlying.ordering
```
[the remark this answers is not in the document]
dhanesharole: I have removed it. But any specific reason for this? I follow [this](https://github.com/databricks/scala-style-guide#override_modifier)  Happy to adapt to current project's guideline.

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
```diff
@@ -3,52 +3,126 @@ package collection
...
+
+    override def rangeImpl(from: Option[K], until: Option[K]): SortedMap[K, V] =
+      new WithDefault[K, V](underlying.rangeImpl(from, until), d)
+
+    override def remove(key: K): SortedMap[K, V] = new WithDefault[K, V](underlying - key, d)
+
+    override def sortedMapFactory: SortedMapFactory[SortedMap] = underlying.sortedMapFactory
+
+    override protected[this] def sortedMapFromIterable[K2, V2](it: collection.Iterable[(K2, V2)])(implicit ordering: Ordering[K2]): SortedMap[K2, V2] =
+      SortedMap.from(it)
```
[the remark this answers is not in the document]
dhanesharole: We can't do this with current implementation of `Collection.SortedMapOps` as `sortedMapFromIterable` has access modifier of `protected[this]`. Is it kept like that for some specific reason? I see that TreeMap is providing its own implementation of this. Hence, I used default SortedMapFactory via Delegate to build new `SortableMap[K2, V2]`. Also, it can't be used with to create `SotedMap[K,V].WithDefault` as types `K`, `K2` and `V`, `V2` are invariant in nature. The same method is used by `flatMap` as well. Happy to discuss more on this :)

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
dhanesharole: Same problem as mentioned above `fromSpecificIterable` has access modifier of `protected[this]`. Maybe as a part of this fix, we can make it just `protected`. This will also make it easy to extract a common `Map.WithDefault[_]` type.

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
dhanesharole: Same problem of access modifiers. It is also `protected[this]`

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
dhanesharole: I have overriden [rangeImpl](https://github.com/dhanesharole/collection-strawman/blob/3ec1db989f9e9439945fa31d41b9dc935895a300/collections/src/main/scala/strawman/collection/immutable/SortedMap.scala#L94) and have written [ unit test ](https://github.com/dhanesharole/collection-strawman/blob/3ec1db989f9e9439945fa31d41b9dc935895a300/test/junit/src/test/scala/strawman/collection/immutable/SortedMapTest.scala#L36) to verify the behaviour.

test/junit/src/test/scala/strawman/collection/immutable/SortedMapTest.scala
dhanesharole: That's a good idea. But if we have to persist this value then we will have to put constraints on type variance of `K, V` parameters when someone uses `withDefault` operations. This will be more constraining than the compromise of losing a default value when underlying types of Iterables are changed due to operations such as `map` and `flatmap`

test/junit/src/test/scala/strawman/collection/immutable/SortedMapTest.scala
dhanesharole: Yeah, it is. Checked on REPL with both versions. I would suggest we should keep the same behaviour going ahead as well.

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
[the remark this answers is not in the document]
julienrf: Unfortunately, always putting `override` also has a drawback that is not mentioned in this style guide: you can accidentally override a parent implementation.

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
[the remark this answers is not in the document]
julienrf: Indeed :-/

I don’t think it would be sound to have it just `protected`, though. Can you give it a try?

test/junit/src/test/scala/strawman/collection/immutable/SortedMapTest.scala
julienrf: Oh, good point, I missed that…

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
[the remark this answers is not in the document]
dhanesharole: +1. Makes sense.

dhanesharole:
@julienrf I have pushed few more changes to accommodate feedback that you had given me in the morning. I have tried extracting out `WithDefaultOps`. Some common logic required by the `withDefault operations` is put into it. Although, because of type preciseness for operations like `updated`, `remove` which call the same methods on underlying object can't be moved into `WithDefaultOps` trait. Because their result types are different viz. `Map[K, V]` and `SortedMap[K, V]`. There is one **costly** way by which we can actually move all operations to `WithDefaultOps`. I was thinking of making `WithDefaultOps` a typeclass with an implicit from that can convert type `Map[K, V]` to `CC[K, V]`, then we can do all operations on type `Map` inside `WithDefaultOps` at the typeclass and finally wrap it with corresponding `CC[K, V]`. But doing so will involve the cost of rebuilding the whole map everytime an operation like `update` or `remove` is called. Let me know your thoughts regarding this.

dhanesharole:
Also, implementation of Mutable types is pending. Will push changes for it asap. Certainly, I will use the same `WithDefaultOps` trait there.

julienrf:
Thanks for sharing your thoughts with us. Don’t over generalize things :)

If it’s easy and safer to have a common super class between `Map.WithDefault` and `SortedMap.WithDefault`, let’s do it. If not, it’s fine to duplicate a few lines of code…

dhanesharole:
@julienrf , I think it won't be safer to have a common base type such `BaseWithDefaultMap` for `WithDefaultMap` and `WithDefaultSortableMap`. Because methods that can go into that base class will be methods like `update`, `remove` et al. So if a user creates default map for underlying type `SortedMap` and calls `remove` then she will get back type `BaseWithDefaultMap` not `SortedMap`. I guess that's not desirable as it will break all the assumptions about the ordering of keys from the result type. Hence, instead of extracting operations into Common base class of type `BaseWithDefaultMap`, I have created a trait `WithDefaultOps` which provides implementations for few common functions where types preciseness is not required. ( For instance methods like `mapFactory` from `Map` which always return type `Map` even when called on `SortableMap`). And then each kind of  `WithDefaultMap` extends the corresponding  `Map` type at the call site. For instance, `SortedMapWithDefault` extends `SortedMap` and `MapWithDefault` just extends `Map`. This each default implementation have to just implement specific functions to be in their corresponding categories like `SortedMap` or `Map`.  I hope that makes sense. Let me know, If I 'm missing something. Can you please re-review this PR? And let me know if something needs to be changed to be able to be merged :) (excited!)

collections/src/main/scala/strawman/collection/immutable/Map.scala
```diff
@@ -128,27 +128,19 @@ trait MapOps[K, +V, +CC[X, +Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]
...
-    def get(key: K) = underlying.get(key)
-    def iterator() = underlying.iterator()
-    override def default(key: K): V = d(key)
-    override def empty = new WithDefault(underlying.empty, d)
-    override def updated[V1 >: V](key: K, value: V1): WithDefault[K, V1] = new WithDefault[K, V1](underlying.updated[V1](key, value), d)
-    override def remove (key: K): WithDefault[K, V] = new WithDefault(underlying - key, d)
-    override def withDefault[V1 >: V](d: K => V1): immutable.Map[K, V1] = new WithDefault[K, V1](underlying, d)
-    override def withDefaultValue[V1 >: V](d: V1): immutable.Map[K, V1] = new WithDefault[K, V1](underlying, x => d)
+  final class WithDefaultMap[K, +V](val underlying: Map[K, V], val defaultValue: K => V) extends Map[K, V] with WithDefaultImmutableMapOps[K, V, Map[K, V]] {
+    override def remove(key: K): Map[K, V] = new WithDefaultMap[K, V](underlying - key, defaultValue)
```
julienrf: Could you please remove the `override` keyword`?

collections/src/main/scala/strawman/collection/immutable/WithDefaultImmutableMapOps.scala
```diff
@@ -0,0 +1,24 @@
+package strawman.collection.immutable
+
+import strawman.collection.{IterableFactoryLike, MapFactory}
+import scala.{Option}
+
+trait WithDefaultImmutableMapOps[K, +V, +C <: Map[K, V]] { self: C =>
```
julienrf: Could you make it `private[collection]`?

collections/src/main/scala/strawman/collection/immutable/WithDefaultImmutableMapOps.scala
```diff
@@ -0,0 +1,24 @@
...
+  
+  val underlying: C
```
julienrf: Looks like this is not necessary since we already have the `self` reference.

collections/src/main/scala/strawman/collection/immutable/WithDefaultImmutableMapOps.scala
```diff
@@ -0,0 +1,24 @@
...
+  
+  val defaultValue: K => V
```
julienrf: You should remove this and keep only the `default(key: K): V` method.

collections/src/main/scala/strawman/collection/immutable/WithDefaultImmutableMapOps.scala
```diff
@@ -0,0 +1,24 @@
...
+  
+  def mapFactory: MapFactory[Map] = underlying.mapFactory
+  
+  def get(key: K): Option[V] = underlying.get(key)
+  
+  def iterableFactory: IterableFactoryLike[Iterable] = underlying.iterableFactory
+  
+  def iterator(): strawman.collection.Iterator[(K, V)] = underlying.iterator()
+  
+  override protected[this] def mapFromIterable[K2, V2](it: strawman.collection.Iterable[(K2, V2)]): Map[K2, V2] =
```
julienrf: Please remove the `override` keyword.

collections/src/main/scala/strawman/collection/immutable/Map.scala
```diff
@@ -128,27 +128,19 @@ trait MapOps[K, +V, +CC[X, +Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]
...
-    override def size = underlying.size
-    def get(key: K) = underlying.get(key)
-    def iterator() = underlying.iterator()
-    override def default(key: K): V = d(key)
-    override def empty = new WithDefault(underlying.empty, d)
-    override def updated[V1 >: V](key: K, value: V1): WithDefault[K, V1] = new WithDefault[K, V1](underlying.updated[V1](key, value), d)
-    override def remove (key: K): WithDefault[K, V] = new WithDefault(underlying - key, d)
-    override def withDefault[V1 >: V](d: K => V1): immutable.Map[K, V1] = new WithDefault[K, V1](underlying, d)
-    override def withDefaultValue[V1 >: V](d: V1): immutable.Map[K, V1] = new WithDefault[K, V1](underlying, x => d)
+  final class WithDefaultMap[K, +V](val underlying: Map[K, V], val defaultValue: K => V) extends Map[K, V] with WithDefaultImmutableMapOps[K, V, Map[K, V]] {
```
julienrf: Why did you change the name? I would keep using `Map.WithDefault` and `SortedMap.WithDefault`. The common trait can be just `MapWithDefault`.

collections/src/main/scala/strawman/collection/immutable/WithDefaultImmutableMapOps.scala
julienrf: Hmm, no actually they are different so we need both, indeed!

collections/src/main/scala/strawman/collection/immutable/Map.scala
dhanesharole: I changed it, as it only provides an implementation for `MapOps` from immutable hierarchy. But making it `private[collection]` will not need it this. Reverting rename commit.

collections/src/main/scala/strawman/collection/immutable/Map.scala
dhanesharole: Sorry again. Will do that. I have got Intellij template which adds them, will disable it for future commits to strawman-collection. :)

collections/src/main/scala/strawman/collection/immutable/WithDefaultImmutableMapOps.scala
dhanesharole: Yes.

collections/src/main/scala/strawman/collection/immutable/WithDefaultImmutableMapOps.scala
dhanesharole: I added so that I can provide the common implementation of `default(key: V): V` method from in `WithDefaultOps` trait. If I remove this, then implementors of this trait will have to implement `deafult` method in their own class with code like this `default(k) = defaultFunction(k)`. where `defaultFunction: K => V` is constructor parameter. Hence, did this change to enforce every implementor of `WithDefaultMapOps` to provide default function implementation so as to get the common implementation of `default(k: V): V` for free. This just removes small duplication from implementor's class.

collections/src/main/scala/strawman/collection/immutable/WithDefaultImmutableMapOps.scala
dhanesharole: Yes.

collections/src/main/scala/strawman/collection/immutable/WithDefaultImmutableMapOps.scala
julienrf: Ok then.

collections/src/main/scala/strawman/collection/immutable/WithDefaultImmutableMapOps.scala
dhanesharole: Made it `private[collection]` and moved it into collection package instead of `immutable`.

dhanesharole: Instead of `private[collection]`, making `private[immutable]`. As it can be only used for `immutable` collections. For mutable ones will have to define seperate trait, as return types are different. E.g. return type of mapFactory for immutable is `mapFactory[Immutable.Map]` whereas for mutable it is `mapFactory[mutable.Map]`.

dhanesharole:
Pushed, latest changes by fixing all the above-mentioned issues.

collections/src/main/scala/strawman/collection/immutable/WithDefaultImmutableMapOps.scala
dhanesharole: You are right. Removed that `defaultValue` from `WithDefaultOps`

dhanesharole:
@julienrf , can you checkout this latest changes. I have made `SortedMap.WithDefault` extend `SortedMapOps' and  `Map.WithDefault` extend `MapOps`. This makes sure that, whenever default  value is lost in combinators like `map`, `flatMap`  etc, type inference will automatically infer the type to be `SortedMap` or `Map`. But for other operations where default value is persisted type will always be `SortedMap.WithDefault` or `Map.WithDefault`. This gives more compile time checking and knowledge about when default value is lost or when not ? If you find it be useful, then I will keep it. Otherwise, I will revert these few commits. Apart from that, implementation for `mutable` is also added, as asked by you.

dhanesharole:
@julienrf, any word on this ?

collections/src/main/scala/strawman/collection/immutable/Map.scala
```diff
@@ -128,27 +128,37 @@ trait MapOps[K, +V, +CC[X, +Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]
...
+      new WithDefault[K, V](mapFactory.from(coll), defaultValue)
+
+    protected[this] def newSpecificBuilder(): Builder[(K, V), Map.WithDefault[K, V]] =
+      mapFactory.newBuilder[K, V]().mapResult(new WithDefault[K, V](_, defaultValue))
+
+    override def mapFactory: MapFactory[Map] = WithDefault
+
+    override def get(key: K): Option[V] = underlying.get(key)
+
+    override def iterableFactory: IterableFactoryLike[Iterable] = underlying.iterableFactory
```
julienrf: The return type should be `IterableFactory[Iterable]`

collections/src/main/scala/strawman/collection/immutable/Map.scala
```diff
@@ -128,27 +128,37 @@ trait MapOps[K, +V, +CC[X, +Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]
...
+
+    override def iterator(): strawman.collection.Iterator[(K, V)] = underlying.iterator()
+
+    override protected[this] def mapFromIterable[K2, V2](it: strawman.collection.Iterable[(K2, V2)]): Map[K2, V2] =
+      Map.from[K2, V2](it)
```
julienrf: We could use `underlying.mapFactory.from`, to preserve the same map implementation.

collections/src/main/scala/strawman/collection/immutable/Map.scala
```diff
@@ -128,27 +128,37 @@ trait MapOps[K, +V, +CC[X, +Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]
...
-    override def default(key: K): V = d(key)
-    override def empty = new WithDefault(underlying.empty, d)
-    override def updated[V1 >: V](key: K, value: V1): WithDefault[K, V1] = new WithDefault[K, V1](underlying.updated[V1](key, value), d)
-    override def remove (key: K): WithDefault[K, V] = new WithDefault(underlying - key, d)
-    override def withDefault[V1 >: V](d: K => V1): immutable.Map[K, V1] = new WithDefault[K, V1](underlying, d)
-    override def withDefaultValue[V1 >: V](d: V1): immutable.Map[K, V1] = new WithDefault[K, V1](underlying, x => d)
+  object WithDefault extends MapFactory.Delegate[Map](HashMap)
+
+  final class WithDefault[K, +V](val underlying: Map[K, V], val defaultValue: K => V)
+    extends Map[K, V]
```
julienrf: Shouldn’t you also extend `MapOps[K, V, Map, WithDefault[K, V]]`?

collections/src/main/scala/strawman/collection/immutable/WithDefaultOps.scala
```diff
@@ -0,0 +1,13 @@
...
+import scala.{Option}
+
+private[immutable] trait WithDefaultOps[K, +V, +C <: Map[K, V]] { self: C =>
+
+  val underlying: C
+
+  val defaultValue: K => V
+
+  override def default(key: K): V = defaultValue(key)
+}
```
julienrf: This trait is so small, it seems that this abstract is not worth (I know it’s my fault that you tried to abstract over `Map.WithDefault` and `SortedMap.WithDefault` but now I see that this doesn’t seem to be useful).

collections/src/main/scala/strawman/collection/immutable/Map.scala
```diff
@@ -128,27 +128,37 @@ trait MapOps[K, +V, +CC[X, +Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]
...
-    def iterator() = underlying.iterator()
-    override def default(key: K): V = d(key)
-    override def empty = new WithDefault(underlying.empty, d)
-    override def updated[V1 >: V](key: K, value: V1): WithDefault[K, V1] = new WithDefault[K, V1](underlying.updated[V1](key, value), d)
-    override def remove (key: K): WithDefault[K, V] = new WithDefault(underlying - key, d)
-    override def withDefault[V1 >: V](d: K => V1): immutable.Map[K, V1] = new WithDefault[K, V1](underlying, d)
-    override def withDefaultValue[V1 >: V](d: V1): immutable.Map[K, V1] = new WithDefault[K, V1](underlying, x => d)
+  object WithDefault extends MapFactory.Delegate[Map](HashMap)
+
+  final class WithDefault[K, +V](val underlying: Map[K, V], val defaultValue: K => V)
```
julienrf: Also, could you make `Map#withDefault` return a `Map.WithDefault`? (a consequence of that is that `SortedMap.WithDefault` will have to extend `Map.WithDefault`, I don’t know if that’s possible)

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
```diff
@@ -51,4 +72,50 @@ trait SortedMapOps[K, +V, +CC[X, +Y] <: Map[X, Y] with SortedMapOps[X, Y, CC, _]
...
+    def rangeImpl(from: Option[K], until: Option[K]): WithDefault[K, V] =
+      new WithDefault[K, V](underlying.rangeImpl(from, until), defaultValue)
+
+    protected[this] def fromSpecificIterable(coll: collection.Iterable[(K, V)]): WithDefault[K, V] =
+      new WithDefault[K, V](SortedMap.from(coll), defaultValue)
+
+    protected[this] def newSpecificBuilder(): Builder[(K, V), WithDefault[K, V]] =
+      SortedMap.newBuilder[K, V]().mapResult(new WithDefault[K, V](_, defaultValue))
+
+    def sortedMapFactory = SortedMap
```
julienrf: Why not `underlying.sortedMapFactory`?

collections/src/main/scala/strawman/collection/immutable/Map.scala
```diff
@@ -128,27 +128,37 @@ trait MapOps[K, +V, +CC[X, +Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]
...
+
+    def empty: Map.WithDefault[K, V] = new WithDefault[K, V](underlying.empty, defaultValue)
+
+    protected[this] def fromSpecificIterable(coll: collection.Iterable[(K, V)]): Map.WithDefault[K, V] =
+      new WithDefault[K, V](mapFactory.from(coll), defaultValue)
+
+    protected[this] def newSpecificBuilder(): Builder[(K, V), Map.WithDefault[K, V]] =
+      mapFactory.newBuilder[K, V]().mapResult(new WithDefault[K, V](_, defaultValue))
+
+    override def mapFactory: MapFactory[Map] = WithDefault
```
julienrf: Why not `underlying.mapFactory`?

collections/src/main/scala/strawman/collection/mutable/Map.scala
```diff
@@ -170,33 +170,17 @@ trait MapOps[K, V, +CC[X, Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]]
...
-    def addOne(elem: (K, V)): this.type = { underlying.addOne(elem); this }
-    def subtractOne(elem: K): this.type = { underlying.subtractOne(elem); this }
-    override def put(key: K, value: V): Option[V] = underlying.put(key, value)
-    override def update(key: K, value: V): Unit = underlying.update(key, value)
-    override def remove(key: K): Option[V] = underlying.remove(key)
-
-    override def withDefault(d: K => V): Map[K, V] = new WithDefault[K, V](underlying, d)
-    override def withDefaultValue(d: V): Map[K, V] = new WithDefault[K, V](underlying, x => d)
+
+  final class WithDefault[K, V](val underlying: Map[K, V], defaultValue: K => V) extends Map[K, V] with WithDefaultOps[K, V, Map[K, V]] {
```
julienrf: Shouldn’t we extend `MapOps[K, V, Map, WithDefault[K, V]]`?

collections/src/main/scala/strawman/collection/mutable/Map.scala
```diff
@@ -170,33 +170,17 @@ trait MapOps[K, V, +CC[X, Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]]
...
+
+    def empty: Map[K, V] = new WithDefault[K, V](underlying.empty, defaultValue)
+
+    protected[this] def fromSpecificIterable(coll: collection.Iterable[(K, V)]): Map[K, V] = mapFactory.from(coll)
```
julienrf: We should return a `WithDefault` here.

collections/src/main/scala/strawman/collection/mutable/SortedMap.scala
```diff
@@ -19,4 +43,38 @@ trait SortedMapOps[K, V, +CC[X, Y] <: Map[X, Y] with SortedMapOps[X, Y, CC, _],
...
+  final class WithDefault[K, V](val underlying: SortedMap[K, V], defaultValue: K => V) extends SortedMap[K, V] with WithDefaultOps[K, V, SortedMap[K, V]] {
+
+    implicit def ordering: Ordering[K] = underlying.ordering
+
+    def empty: SortedMap[K, V] = new WithDefault[K, V](underlying.empty, defaultValue)
+
+    def sortedMapFactory: SortedMapFactory[SortedMap] = underlying.sortedMapFactory
+
+    override protected[this] def sortedMapFromIterable[K2, V2](it: collection.Iterable[(K2, V2)])(implicit ordering: Ordering[K2]): SortedMap[K2, V2] =
+      SortedMap.from(it)
```
julienrf: `underlying.sortedMapFactory.from(it)`?

dhanesharole:
@julienrf , I have pushed changes with this new Design approach. It's indeed looking good to me. Very clean to read and typesafe as return types of operations now indicate the behaviour 'default value'. After this new change, `WithDefault` is just another implementation of standard maps which was very straightforward to implement as u said. Please have a look when you get time.

dhanesharole:
I agree, the only reason I pulled it out into a separate file so that PR remain clean from original MapOps and added implementation in separate files, rather than putting it into Map object and SorteeMap object.

dhanesharole:
If u think this makes it more consistent with original design and code structuring, I will more it there.

dhanesharole:
If u think this makes it more consistent with original design and code structuring, I will move it there.

dhanesharole:
@julienrf , I have pushed changes by migrating those types form separate file to a Map and SortedMap objects. Can you have a look ?  I hope, I 've got it right this time. Really sorry for so much of to and fro that's going on for this small feature. :)

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
```diff
@@ -40,15 +61,55 @@ trait SortedMapOps[K, +V, +CC[X, +Y] <: Map[X, Y] with SortedMapOps[X, Y, CC, _]
...
+    def keysIteratorFrom(start: K): strawman.collection.Iterator[K] = underlying.keysIteratorFrom(start)
+
+    def rangeImpl(from: Option[K], until: Option[K]): WithDefault[K, V] =
+      new WithDefault[K, V](underlying.rangeImpl(from, until), defaultValue)
+
+    // Need to override following methods to match type signatures of `SortedMap.WithDefault`
+    // for operations preserving default value
+
+    override protected[this] def mapFromIterable[K2, V2](it: collection.Iterable[(K2, V2)]): Map[K2, V2] =
+      mapFactory.from(it)
```
julienrf: It looks like this is the same implementation as in the super type.

collections/src/main/scala/strawman/collection/mutable/SortedMap.scala
```diff
@@ -19,4 +43,41 @@ trait SortedMapOps[K, V, +CC[X, Y] <: Map[X, Y] with SortedMapOps[X, Y, CC, _],
...
+
+    protected[this] def sortedMapFromIterable[K2, V2](it: strawman.collection.Iterable[(K2, V2)])(implicit ordering: Ordering[K2]): SortedMap[K2, V2] =
+      sortedMapFactory.from(it)
+
+    def rangeImpl(from: Option[K], until: Option[K]): WithDefault[K, V] =
+      new WithDefault[K, V](underlying.rangeImpl(from, until), defaultValue)
+
+    // Need to override following methods to match type signatures of `SortedMap.WithDefault`
+    // for operations preserving default value
+    override def mapFromIterable[K2, V2](it: collection.Iterable[(K2, V2)]): Map[K2, V2] = mapFactory.from(it)
```
julienrf: Same question here.

collections/src/main/scala/strawman/collection/mutable/WithDefaultOps.scala
```diff
@@ -0,0 +1,31 @@
...
+    *  Note: The default is only used for `apply`. Other methods like `get`, `contains`, `iterator`, `keys`, etc.
+    *  are not affected by `withDefaultValue`.
+    *
+    *  Invoking transformer methods (e.g. `map`) will not preserve the default value.
+    *
+    *  @param d     default value used for non-present keys
+    *  @return      a wrapper of the map with a default value
+    */
+  def withDefaultValue(d: V): WithDefault[K, V]
+}
```
julienrf: I guess this file can be removed now?

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
dhanesharole: Yeah, actually I had to add it purposefully with `override`. `SortedMap.WithDefault` is extending `SortedMap' and `Map.WithDefault`. But since `SortedMap` already implements `mapFromIterable`, there is collision happening in the inherited members of `SortedMap.WithDefault`.  See [this in implementation of `mapFromIterable` in `SortedMap`](https://github.com/scala/collection-strawman/blob/1660bb51be0175bd6b64758727e95b5d8c096d30/collections/src/main/scala/strawman/collection/immutable/SortedMap.scala#L38)

collections/src/main/scala/strawman/collection/mutable/SortedMap.scala
dhanesharole: Same as above.

collections/src/main/scala/strawman/collection/mutable/WithDefaultOps.scala
dhanesharole: Yeah, that was miss. Removed and pushed.

collections/src/main/scala/strawman/collection/immutable/SortedMap.scala
dhanesharole: also, super type always returns `Map.from(it)`. Whereas, default implementations both in mutable and immutable return map created from `underlying.mapFactory.From(it)`. This makes sure that, whenever `mapFromIterable` is called on withDefault type, it will return map from underlying map's factory than plain `Map.from(it)`.  Based on [this feedback](https://github.com/scala/collection-strawman/pull/381/files/83fa419316f90b1d7f11003cb0000a0ae146da57#diff-d2ae80cbea21c68573a69ce27e2d0323R152) I thought, it's better to `override` it in WithDefault implementation than making super type's implementation final.

dhanesharole:
Thanks a lot @julienrf , looking forward to contributing more to collection-strawaman.

PR #383, opened 2018-01-29
The UnrolledBuffer type from the old collections

szeiger:


collections/src/main/scala/strawman/collection/mutable/UnrolledBuffer.scala
```diff
@@ -0,0 +1,435 @@
...
+
+  override def className = "UnrolledBuffer"
+}
+
+
+object UnrolledBuffer extends StrictOptimizedSeqFactory[UnrolledBuffer] {
+
+  def empty[A]: UnrolledBuffer[A] = empty(ClassTag.Any.asInstanceOf[ClassTag[A]])
+
+  def empty[A](implicit t: ClassTag[A]): UnrolledBuffer[A] = new UnrolledBuffer[A]
```
julienrf: What’s the advantage of having both overloads? (the one that uses `ClassTag.Any` and the one that uses the implicit `ClassTag[A]`)

szeiger: You want the methods with the `ClassTag` wherever possible but `IterableFactory` requires the ones without. This is not the first such collection type and the old collections have dedicated support for this. We should consider doing the same and adding a new factory type.

julienrf: if we write `empty[SomeType]` which overload is resolved? It depends on whether you have an implicit `ClassTag[SomeType]` in scope? Did you do some tests? (my intuition is that what’s is in the implicit scope doesn’t influence overloading resolution)

szeiger: Right, they're ambiguous in this case. We need to move the more general version into a superclass to disambiguate (like the overloads with `Ordering` in some collection types)

szeiger: Better ClassTag-based factories that could be used for `UnrolledBuffer`, too, in https://github.com/scala/collection-strawman/pull/391

szeiger:
Updated to use the new `ClassTagSeqFactory`

julienrf:
I noticed that this PR re-introduced some warnings. Could you take care of cleaning them @szeiger?

julienrf:
I also just noticed that you didn’t port the junit tests as well. (in test/junit/src/test/scala/scala/collection/mutable/UnrolledBufferTest.scala)

PR #384, opened 2018-01-30
Add search methods from scala.collection.Searching

szeiger:
It is not clear from https://github.com/scala/scala/pull/903 why these
methods were not added directly to the collection classes, so that is
what I’m doing here.

`Searching` still exists and contains `SearchResult` and its subtypes.
We should consider moving `SearchResult` directly into the package
and `Found` and `InsertionPoint` into its companion object.


szeiger:
OK, `git clean -d -f` fixed the Dotty problem. `sbt clean` was not enough.

PR #386, opened 2018-01-31
Overload unzip in ArrayOps

### LINKED ISSUE #374: Array[(T, U)] unzip builds Vectors not Arrays
```
scala> def f[T, U](a: Array[(T, U)]) = a.unzip
f: [T, U](a: Array[(T, U)])(Seq[T], Seq[U])

scala> f(Array((1,2)))
res1: (Seq[Int], Seq[Int]) = (Vector(1),Vector(2))
```

In 2.12, the definition of `f` doesn't compile, it requires `ClassTag`s on the type parameters. Then

```
scala> def f[T: reflect.ClassTag, U: reflect.ClassTag](a: Array[(T, U)]) = a.unzip
f: [T, U](a: Array[(T, U)])(implicit evidence$1: scala.reflect.ClassTag[T], implicit evidence$2: scala.reflect.ClassTag[U])(Array[T], Array[U])

scala> f(Array((1,2)))
res0: (Array[Int], Array[Int]) = (Array(1),Array(2))
```

marcelocenerine:
Fixes #374

### DIFF
--- a/collections/src/main/scala/strawman/collection/ArrayOps.scala
+++ b/collections/src/main/scala/strawman/collection/ArrayOps.scala
@@ -76,4 +76,30 @@ class ArrayOps[A](val xs: Array[A]) extends AnyVal
 
   def patch[B >: A : ClassTag](from: Int, other: Iterable[B], replaced: Int): Array[B] =
     fromTaggedIterable(new View.Patched(toIterable, from, other, replaced)) //TODO optimize
+
+  /** Converts an array of pairs into an array of first elements and an array of second elements.
+    *
+    *  @tparam A1    the type of the first half of the element pairs
+    *  @tparam A2    the type of the second half of the element pairs
+    *  @param asPair an implicit conversion which asserts that the element type
+    *                of this Array is a pair.
+    *  @param ct1    a class tag for `A1` type parameter that is required to create an instance
+    *                of `Array[A1]`
+    *  @param ct2    a class tag for `A2` type parameter that is required to create an instance
+    *                of `Array[A2]`
+    *  @return       a pair of Arrays, containing, respectively, the first and second half
+    *                of each element pair of this Array.
+    */
+  def unzip[A1, A2](implicit asPair: A => (A1, A2), ct1: ClassTag[A1], ct2: ClassTag[A2]): (Array[A1], Array[A2]) = {
+    val a1 = new Array[A1](length)
+    val a2 = new Array[A2](length)
+    var i = 0
+    while (i < length) {
+      val e = asPair(apply(i))
+      a1(i) = e._1
+      a2(i) = e._2
+      i += 1
+    }
+    (a1, a2)
+  }
 }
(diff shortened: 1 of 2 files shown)


collections/src/main/scala/strawman/collection/ArrayOps.scala
```diff
@@ -76,4 +76,30 @@ class ArrayOps[A](val xs: Array[A]) extends AnyVal
...
+    val a2 = new Array[A2](length)
+    var i = 0
+    while (i < length) {
+      val e = apply(i)
+      a1(i) = e._1
+      a2(i) = e._2
+      i += 1
+    }
+    (a1, a2)
+  }
```
marcelocenerine: copy/paste from `scala.collection.mutable.ArrayOps`

marcelocenerine: explicitly calling `asPair` to make Dotty compiler happy:

```
[error] -- [E008] Member Not Found Error: /home/travis/build/scala/collection-strawman/collections/src/main/scala/strawman/collection/ArrayOps.scala:99:16 
[error] 99 |      a1(i) = e._1
[error]    |              ^^^^
[error]    |              value `_1` is not a member of A - did you mean `e.==`?
```

test/junit/src/test/scala/strawman/collection/ArrayOpsTest.scala
```diff
@@ -0,0 +1,20 @@
...
+import org.junit.runners.JUnit4
+
+@RunWith(classOf[JUnit4])
+class ArrayOpsTest {
+
+  @Test
+  def unzip(): Unit = {
+    val zipped = new ArrayOps(Array((1, 'a'), (2, 'b'), (3, 'c')))
+
+    val (a1: Array[Int], a2: Array[Char]) = zipped.unzip
```
julienrf: Does that also work if you don’t specify the expected types? Also, why not let the implicit conversion to `ArrayOps` do its job instead of explicitly converting to `ArrayOps`?

marcelocenerine: >Does that also work if you don’t specify the expected types?

It does. I removed the types as `assertArrayEquals` is enough to make sure that what comes out of `unzip` is a tuple of arrays. 

>why not let the implicit conversion to ArrayOps do its job instead of explicitly converting to ArrayOps?

For some reason the implicit conversion from `Predef` is used when selecting `unzip`, even though they are unimported. Any idea?

julienrf: In the `junit` project `Predef` is not unimported by default so you have to unimport it manually.

marcelocenerine: sorry, it turns out Intellij was tricking me. It showed the implicit conversions coming from `Predef` even though I had unimported them manually.

Issue #331, opened 2018-01-03
proper serialization support for LazyList and Stream


eyalfa:
my original intention was to clone https://github.com/scala/bug/issues/10679 to this project.
after checking out the project and cloning the test I found out LazyList is not serializable at all.

is this intentional? if not will you accept a pr implementing proper serialization for LazyList avoiding the issue Stream serialization encounters (described in the linked issue).

Once fixed, re-enable the tests in `test/files/run/serialize-stream.scala`, `test/files/jvm/serialization-new.scala`, `test/files/jvm/serialization.scala` and run`--update-check` (marked with `TODO-newColl`).

julienrf:
We could make it `Serializable`, indeed!

eyalfa:
Great! I'm on it

eyalfa:
@julienrf , serialization support seems to be broken all over the place,
after modifying LazyList's serialization, my test started failing on: java.io.NotSerializableException: strawman.collection.immutable.NumericRange$$anon$2 (instead of the expected stack overflow).
is there a global task of adding serialization support to the strawman's collections? dropping this support is a regression affecting many applications (i.e. spark jobs and alike).

eyalfa:
issue seems global to the entire library, changed from `LazyList.range(..)` to `LazyList.from(List.range(...))`, now my test fails on java.io.NotSerializableException: strawman.collection.LinearSeqOps$$anon$1.
a quick check with javap proved this is the anonymous iterator class returned by `LinearSeqOps.iterator`.

julienrf:
We should fix serialization support, indeed. That’s out of the scope of the current issue, however, but would you interested in fixing what needs to be fixed to get `LazyList` serializable?

eyalfa:
I'd be happy to, problem is it seems a bit eclactic as I'd be touching a
very specific code path. I.e. if I stick with LazyList.range I'd have to
touch NumericRange an probably at least one of the Integral type class
implementation, if stick with List.range I'd have to make the iterator
serializable...
I sense I'd face many micro decisions where a macro one is in order... Good
thing is this code will be reviewed back and forth...

eyalfa:
@julienrf , feels like I've hit a wall:

strawman.collection.immutable.LazyList#fromIterator:
```
def fromIterator[A](it: Iterator[A]): LazyList[A] =
    if (it.hasNext) {
      // Be sure that `it.next()` is called even when the `head`
[...]

PR #389, opened 2018-01-31
Fix serialization issues in HashMap, LinkedHashMap, HashSet and LinkedHashSet

### LINKED ISSUE #377: mutable HashSet/HashMap not serializable
```
scala>   def serializeDeserialize[T <: AnyRef](obj: T): T = {
     |     import java.io._
     |     val buffer = new ByteArrayOutputStream
     |     val out = new ObjectOutputStream(buffer)
     |     out.writeObject(obj)
     |     val in = new ObjectInputStream(new ByteArrayInputStream(buffer.toByteArray))
     |     in.readObject.asInstanceOf[T]
     |   }

scala> serializeDeserialize(mutable.HashSet())
java.io.NotSerializableException: scala.collection.mutable.FlatHashTable

scala> serializeDeserialize(mutable.HashMap())
java.io.NotSerializableException: scala.collection.mutable.HashMap$$anon$1
```

marcelocenerine:
This PR fixes #377. 

The reason for the `NotSerializableException`s is that the `HashTable`/`FlatHashTable` classes are not serializable nor the `table` fields in `HashMap`/`LinkedHashMap`/`HashSet`/`LinkedHashSet` classes are marked as transient. 

The easiest solution would be to make `HashTable`/`FlatHashTable` serializable. However, there are a few more issues here:

1. `out.defaultWriteObject` and `in.defaultReadObject` should be called from the `writeObject`/`readObject` methods of the class being serialized/deserialized (usually in the 1st line). Currently this is done from methods in `HashTable`/`FlatHashTable`, which is confusing and causes the issue below;
2. `table.init` is being called before `defaultReadObject`, which causes a NPE during deserialization (need to get past the `NotSerializableException` to see this one);
3. `HashMap`/`LinkedHashMap`/`LinkedHashSet` creates anonymous instances of `HashTable`. In Java (I assume the same applies to Scala) it is not recommended to serialize inner classes as these use compiler-generated synthetic fields to store references to enclosing instances and to store values of local variables from enclosing scopes (it is not a big deal now as the attributes in the enclosed classes are marked transient). Also, it's unspecified how these fields correspond to the class definition, as well as the names of anonymous and local classes.
4. `HashTable` and `FlatHashTable` are private classes in the `strawman.collection.mutable` package. If these classes are included in the serialized form of dependent classes, the former becomes part of the exported API of the latter. This decreases the flexibility to change them once they are released.

Issues 3 and 4 (name of anonymous and private classes leaking) are exemplified in the output below:

```
val obj = mutable.HashMap(1 -> "one", 2 -> "two", 3 -> "three")`:
val buffer = new ByteArrayOutputStream
val out = new ObjectOutputStream(buffer)
out.writeObject(obj)
println(new String(buffer.toByteArray)) 
```

```
��sr#strawman.collection.mutable.HashMapLtablet'Lstrawman/collection/mutable/HashTable;xpsr+strawman.collection.mutable.HashMap$$anon$1˜-���xr%strawman.collection.mutable.HashTab�srjava.lang.Integer⠤���8Ivaluexrjava.lang.Number���
                                                    ���xpttwosq~tonesq~tthreex
```

Below is the output for the same object after the changes in this PR were implemented:

```
�srjava.lang.Integer⠤���8Ivaluexrjava.lang.Number���
                                                    ���xpttwosq~tonesq~tthreex
```

### DIFF
--- a/collections/src/main/scala/strawman/collection/immutable/TreeSet.scala
+++ b/collections/src/main/scala/strawman/collection/immutable/TreeSet.scala
@@ -5,7 +5,7 @@ package immutable
 import mutable.{Builder, ImmutableBuilder}
 import immutable.{RedBlackTree => RB}
 
-import scala.{Boolean, Int, math, NullPointerException, Option, Ordering, Some, Unit}
+import scala.{Boolean, Int, math, NullPointerException, Option, Ordering, Serializable, SerialVersionUID, Some, Unit}
 
 /** This class implements immutable sorted sets using a tree.
   *
@@ -25,11 +25,13 @@ import scala.{Boolean, Int, math, NullPointerException, Option, Ordering, Some, 
   *  @define mayNotTerminateInf
   *  @define willNotTerminateInf
   */
+@SerialVersionUID(-5685982407650748405L)
 final class TreeSet[A] private (tree: RB.Tree[A, Unit])(implicit val ordering: Ordering[A])
   extends SortedSet[A]
     with SortedSetOps[A, TreeSet, TreeSet[A]]
     with StrictOptimizedIterableOps[A, Set, TreeSet[A]]
-    with StrictOptimizedSortedSetOps[A, TreeSet, TreeSet[A]] {
+    with StrictOptimizedSortedSetOps[A, TreeSet, TreeSet[A]]
+    with Serializable {
 
   if (ordering eq null) throw new NullPointerException("ordering must not be null")
 
--- a/test/junit/src/test/scala/strawman/collection/immutable/SerializationTest.scala
+++ b/test/junit/src/test/scala/strawman/collection/immutable/SerializationTest.scala
@@ -58,11 +58,11 @@ class SerializationTest {
     assertEqualsAfterDeserialization(BitSet(1, 2, 3))
   }
 
-//  @Test
-//  def treeSet(): Unit = {
-//    assertEqualsAfterDeserialization(TreeSet.empty)
-//    assertEqualsAfterDeserialization(TreeSet(1, 2, 3))
-//  }
+  @Test
+  def treeSet(): Unit = {
+    assertEqualsAfterDeserialization(TreeSet.empty[Int])
+    assertEqualsAfterDeserialization(TreeSet(1, 2, 3))
+  }
 
 //  @Test
 //  def lazyList(): Unit = {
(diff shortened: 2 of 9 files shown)


test/junit/src/test/scala/strawman/collection/immutable/SerializationTest.scala
```diff
@@ -0,0 +1,114 @@
...
+  def bitSet(): Unit = {
+    assertEqualsAfterDeserialization(BitSet.empty)
+    assertEqualsAfterDeserialization(BitSet(1, 2, 3))
+  }
+
+//  @Test
+//  def treeSet(): Unit = {
+//    assertEqualsAfterDeserialization(TreeSet.empty)
+//    assertEqualsAfterDeserialization(TreeSet(1, 2, 3))
+//  }
```
marcelocenerine: bug?

test/junit/src/test/scala/strawman/collection/immutable/SerializationTest.scala
```diff
@@ -0,0 +1,114 @@
...
+
+//  @Test
+//  def lazyList(): Unit = {
+//    assertEqualsAfterDeserialization(LazyList.empty)
+//    assertEqualsAfterDeserialization(LazyList.from(1))
+//  }
```
marcelocenerine: `Stream` is serializable whereas `LazyList` is not.

marcelocenerine: #331

marcelocenerine:
I realized that `HashTable` and `FlatHashTable` both exist in `scala.collection`. However, they are public traits mixed in with maps/sets rather than private classes used in private fields. What I mentioned above is not relevant in that case.

julienrf:
The motivation to use aggregation rather than inheritance was to simplify the public facing type hierarchy.

marcelocenerine:
@szeiger / @lrytz, any inputs here?

collections/src/main/scala/strawman/collection/immutable/TreeSet.scala
```diff
@@ -25,11 +25,13 @@ import scala.{Boolean, Int, math, NullPointerException, Option, Ordering, Some,
   *  @define mayNotTerminateInf
   *  @define willNotTerminateInf
   */
+@SerialVersionUID(-5685982407650748405L)
```
lrytz: I started writing a comment here, but actually let's leave it for now and come back to it later :-) https://github.com/scala/collection-strawman/issues/418

marcelocenerine: yeah, when I saw the issue you created it immediately came up to my mind.

Issue #308, opened 2017-11-30
StrawmanTest.mainTest failing in Scala 2.13.0-M3


SethTisue:
```
[collection-strawman] [error] Test strawman.collection.test.StrawmanTest.mainTest failed: java.lang.AssertionError: assertion failed, took 0.472 sec
[collection-strawman] [error]     at scala.Predef$.assert(Predef.scala:204)
[collection-strawman] [error]     at strawman.collection.test.StrawmanTest.lazyListOps(Test.scala:466)
[collection-strawman] [error]     at strawman.collection.test.StrawmanTest.mainTest(Test.scala:576)
[collection-strawman] [error]     ...
```

details at https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-community-build/856/consoleFull

doesn't seem to be intermittent (failed in both runs last night)

SethTisue:
or, hmm, maybe it is intermittent, there is a run currently in progress where collection-strawman already went green: https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-community-build/857/consoleFull

we'll see how the next few runs go

julienrf:
Did you change something about the evaluation of by-name right-associative operators?

julienrf:
Yeah now I remember that one: https://github.com/scala/scala/pull/5969

But the test should _always_ fail with the new evaluation model.

SethTisue:
I failed to notice that run 857 was manually triggered by Miles against a different 2.13 build.

So yes, always failing.

julienrf:
Hey @SethTisue! What should we do about that? Is this blocking you?

SethTisue:
it's not blocking me.

I don't understand the issue, I don't know what you should do. but in general, this repo ought to compile, and tests pass, on whatever the latest 2.13 nightly is, since before long some descendant of that nightly will become 2.13.0-M3. (@szeiger?)

julienrf:
The issue is that we have a test that checks the semantics of an operation (`#::` on ~`Stream`~ `LazyList`) and that this semantics has changed because the language has changed. I would be in favor of changing the test so that it takes the new language semantics into account. However, the fact that the semantics of `#::` will have changed between 2.12 and 2.13 might be a problem for existing users, though I doubt something relies on the fact that this operator is not really lazy.

SethTisue:
> I would be in favor of changing the test so that it takes the new language semantics into account

👍

[...]

PR #390, opened 2018-02-01
upgrade Scala 2.13 version to 2.13.0-M3

SethTisue:


julienrf:
Looks like we should merge #361 or #345 before :)

SethTisue:
ah right, #308

julienrf:
Now that #361 has been merged I’ve run again the CI on this PR and we have a different error:

~~~
[error] /home/travis/build/scala/collection-strawman/collections-contrib/src/test/scala/strawman/collection/decorators/MapDecoratorTest.scala:12: Cannot construct a collection of type C with elements of type (Int, String) based on a collection of type stabilizer$1.this.type.
[error]     val zipped = map1.zipByKeyWith(map2)(_ ++ _)
[error]                                         ^
~~~

For some reason the compiler is unable to pick the right implicit `BuildFrom`. This was working in 2.13.0-M2.

julienrf:
With `-Xlog-implicits` I get the following message:

~~~
[info] /home/julien/workspace/dev/scala/collection-strawman/collections-contrib/src/test/scala/strawman/collection/decorators/MapDecoratorTest.scala:12: collection.this.BuildFrom.buildFromMapOps is not a valid implicit value for strawman.collection.BuildFrom[stabilizer$1.this.type,(Int, String),C] because:
[info] hasMatchingSymbol reported error: polymorphic expression cannot be instantiated to expected type;
[info]  found   : [CC[X, Y] <: strawman.collection.Map[X,Y] with strawman.collection.MapOps[X, Y, CC, _], K0, V0, K, V]strawman.collection.BuildFrom[CC[K0,V0],(K, V),CC[K,V]]
[info]  required: strawman.collection.BuildFrom[stabilizer$1.this.type,(Int, String),C]
[info]     val zipped = map1.zipByKeyWith(map2)(_ ++ _)
[info]                                         ^
~~~

However, this candidate should match. And if I explicitly supply it like the following it compiles:

~~~ scala
val zipped = map1.zipByKeyWith(map2)(_ ++ _)(BuildFrom.buildFromMapOps)
~~~

SethTisue:
rebased

julienrf:
Is it possible for you to investigate the regression in the implicit search or do you want me to minimize the test case?

julienrf:
Maybe scala/scala#6308 would be the culprit?

julienrf:
I’ve created https://github.com/scala/bug/issues/10736 to keep track of the regression.

julienrf:
I’m closing this PR. The changes made here have been replicated in #345 and adapted a bit to not use M3 but a prior scala version that doesn’t produce compilation errors.

PR #391, opened 2018-02-02
First-class treatment of ClassTag-based factories

szeiger:
This reinstates the factory part of the abstraction for arbitrary
evidence type constructors which we previously abandoned in favor of
SortedIterableFactory. Now we have a good reason for an
EvidenceIterableFactory with several Ordering-based and several
ClassTag-based factory instances.

ImmutableArray and WrappedArray now have a ClassTag-based factory in
their companion object so that building unboxed arrays with a proper
element type is always the default (and used automatically in calls like
`xs.to(ImmutableArray)` which would previously go through the regular
IterableFactory and build a boxed representation). IterableFactories are
still available as `{ImmutableArray/WrappedArray}.untagged`.

There are no new abstractions at the collection level. We could
consider reintroducing a ConstrainedIterable but at lower levels of
the hierarchy it doesn’t make sense because the Ordering- and
ClassTag-based collections are of very different types.

There are also some fixes for bad arraycopy calls (between boxed and
unboxed representations) and unnecessary boxing in combinators in
ImmutableArray.

### DIFF
--- a/collections/src/main/scala/strawman/collection/Factory.scala
+++ b/collections/src/main/scala/strawman/collection/Factory.scala
@@ -6,7 +6,7 @@ import strawman.collection.immutable.NumericRange
 import scala.language.implicitConversions
 import strawman.collection.mutable.Builder
 
-import scala.{Any, Array, Char, Int, Integral, Nothing, Ordering, Some}
+import scala.{Any, Array, Char, Int, Integral, Nothing, Ordering, Some, `inline`}
 import scala.Predef.{implicitly, String}
 import scala.annotation.unchecked.uncheckedVariance
 import scala.reflect.ClassTag
@@ -227,7 +227,6 @@ trait IterableFactoryLike[+CC[_]] {
     */
   def tabulate[A](n1: Int, n2: Int, n3: Int, n4: Int, n5: Int)(f: (Int, Int, Int, Int, Int) => A): CC[CC[CC[CC[CC[A]]]] @uncheckedVariance] =
     tabulate(n1)(i1 => tabulate(n2, n3, n4, n5)(f(i1, _, _, _, _)))
-
 }
 
 /** Base trait for companion objects of unconstrained collection types that can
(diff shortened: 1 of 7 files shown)


collections/src/main/scala/strawman/collection/Factory.scala
```diff
@@ -398,51 +397,249 @@ object MapFactory {
   * @define coll collection
   * @define Coll `Iterable`
   */
-trait SortedIterableFactory[+CC[_]] {
+trait EvidenceIterableFactory[+CC[_], Ev[_]] {
```
julienrf: Could you document the `Ev` type parameter in the scaladoc?

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -124,36 +129,34 @@ sealed abstract class ImmutableArray[+A]
...
   * @define Coll `ImmutableArray`
   */
-object ImmutableArray extends StrictOptimizedSeqFactory[ImmutableArray] {
-
-  private[this] lazy val emptyImpl = new ImmutableArray.ofRef[Nothing](new scala.Array[Nothing](0))
+object ImmutableArray extends StrictOptimizedClassTagSeqFactory[ImmutableArray] { self =>
+  val untagged: SeqFactory[ImmutableArray] = new ClassTagSeqFactory.AnySeqDelegate(self)
 
-  def empty[A]: ImmutableArray[A] = emptyImpl
+  private[this] lazy val emptyImpl = new ImmutableArray.ofRef[Nothing](new Array[Nothing](0))
```
julienrf: Why does this need to be `lazy`?

julienrf:
Can you also check that IDE auto-completion doesn’t show an `Ev` but an `Ordering` or `ClassTag` when writing `ImmutableArray.appl` or `TreeSet.appl`?

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
szeiger: It doesn't need to be but there is usually no reason to ever allocate these objects.

szeiger:
Improved scaladocs plus some small fixes

szeiger:
Auto-complete in IntelliJ doesn't work correctly either way. In master it shows `empty[A]` in `mutable.TreeSet` and in this branch it additionally shows `empty[A : Any]`, both of which are obviously wrong.

julienrf:
@szeiger This is what I currently have with master:

treeset-apply

julienrf:
According to a small test that still work fine with your design:

classtag-factory

Issue #372, opened 2018-01-26
Bring back collection.convert.ImplicitConversions


lrytz:
...and its two halves, `ImplicitConversionsToScala` and `ImplicitConversionsToJava`

julienrf:
Fixed by #394.

PR #394, opened 2018-02-03
Bring back collection.convert.ImplicitConversions and fix StackOverflowError in Java wrappers' iterator()

marcelocenerine:
This PR addresses #372 and fixes `StackOverflowError`s found in Java wrappers' iterator() while testing.

### DIFF
--- a/collections/src/main/scala/strawman/collection/convert/ImplicitConversions.scala
+++ b/collections/src/main/scala/strawman/collection/convert/ImplicitConversions.scala
@@ -1,0 +1,172 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___     Scala API                            **
+**    / __/ __// _ | / /  / _ |    (c) 2006-2016, LAMP/EPFL             **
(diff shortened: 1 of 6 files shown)


collections/src/main/scala/strawman/collection/convert/ImplicitConversions.scala
```diff
@@ -0,0 +1,172 @@
...
+** /____/\___/_/ |_/____/_/ | |                                         **
+**                          |/                                          **
+\*                                                                      */
+
+package strawman
+package collection
+package convert
+
+import java.{ lang => jl, util => ju }, java.util.{ concurrent => juc }
+import scala.Predef.String
```
marcelocenerine: with the exception of this import, this file has exact the same code as https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/convert/ImplicitConversions.scala

julienrf:
Perfect, as usual! Thanks @marcelocenerine

lrytz:
Thanks!

PR #414, opened 2018-02-05
Fix Vector.prependedAll

### LINKED ISSUE #399: Vector flips order on ++:
The order of elements should be unchanged when using `++` or `++:`; the difference should only be in which collection is "added to" (which also determines the result type).  This is the case for `List` and `ArrayBuffer` among others, but `Vector` gets the "left hand side" backwards.

```
Vector(1, 2) ++: Vector(3, 4)
```

should be `Vector(1, 2, 3, 4)`.  Presently it is `Vector(2, 1, 3, 4)`.  (It doesn't matter what collection is on the "left" side. it is the `Vector.++:` method that pulls the other side in backwards.  I haven't traced it to Vector itself; it could be a superclass.)

marcelocenerine:
This PR aims to fix #399

`Vector` provides an optimized version of `prependedAll` (overrides the one inherited from `StrictOptimizedSeqOps`). In one of the cases, the `prefix` elements are incorrectly stacked in front of `this`, which messes up with the order of the resulting vector.

My solution to this was to relax the visibility of `Iterable.reversed` as it seemed to be the fastest alternative to have the elements prepended in the correct order.

I ran some [benchmarks](https://gist.github.com/marcelocenerine/0bac34b3c9def261cd5029a874b4d4ae) to compare the performance of:
- Scala's `Vector.++:`
- `prependedAll` inherited from `StrictOptimizedSeqOps`
- the buggy `prependedAll`
- the fixed `prependedAll`

It turns out that the optimization for `Vector` is ~5x faster than the one from `StrictOptimizedSeqOps`. Also, Scala's `Vector` does not override `++:` which makes its performance be similar to `StrictOptimizedSeqOps`'s.

`prependedAll` became slightly slower after my fix (<0.2x). Here are the [JMH logs](https://github.com/scala/collection-strawman/files/1693389/jmh-result.txt).

**Note:** the size parameter in the table/charts below correspond to the prefix size. The Vector size is calculated based on the prefix.size so that the lines of code we are interested in always execute. 

image

image

I also tested a similar approach where the`prefix` iterable is transformed into an `IndexedSeq` and then the traversal is done via `reverseIterator()`. The advantage of this approach is that it would avoid the additional traversal to reverse the elements if `prefix` is already an `IndexedSeq`. However, for non-`IndexedSeq` collections the performance was slightly slower.

image

@julienrf / @Ichoran, please let me know if you have a better strategy.

### DIFF
--- a/collections/src/main/scala/strawman/collection/immutable/Vector.scala
+++ b/collections/src/main/scala/strawman/collection/immutable/Vector.scala
@@ -206,7 +206,8 @@ final class Vector[+A] private[immutable] (private[collection] val startIndex: I
       prefix.size match {
         case n if n <= TinyAppendFaster || n < (this.size >>> Log2ConcatFaster) =>
           var v: Vector[B] = this
-          for (b <- prefix) v = b +: v
+          val it = prefix.toIndexedSeq.reverseIterator()
+          while (it.hasNext) v = it.next() +: v
           v
         case n if this.size < (n >>> Log2ConcatFaster) && prefix.isInstanceOf[Vector[_]] =>
           var v = prefix.asInstanceOf[Vector[B]]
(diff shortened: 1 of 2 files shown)


Ichoran:
I'd optimize for the IndexedSeq case (by using `reverseIterator` I guess?).  From the benchmarks it looks to me like the win is significant, and I think the most common use case is for collections of the same type to be stuck together.

marcelocenerine:
Ok. I updated the PR already

PR #415, opened 2018-02-05
Make LazyList.drop safe

### LINKED ISSUE #395: Unsafe drop in LazyList
This throws an exception:

```
LazyList.from(Array(1)).drop(2)
```

`drop` should be a safe operation.

marcelocenerine:
Fixes #395

- Add extra condition in `LinearSeq.drop` to stop recursion when `tail` is empty (seemed pointless).
- Remove override in `List` as it basically duplicates what `LinearSeq.drop` does now


Ichoran:
Are you sure the extra instance stuff doesn't slow anything down?

marcelocenerine:
That occurs only once per call rather than in ever iteration. Let me run a benchmark to be sure

marcelocenerine:
@Ichoran, here are the results. There is no slow down after this change (even the slightest is still within the margin of error).

image

Ichoran:
@marcelocenerine - Great, thanks for checking!

PR #416, opened 2018-02-05
Make Buffer implement SeqOps to retain original type after invoking operations on it

### LINKED ISSUE #396: Buffers think they're just Seq
Buffers revert to Seq on pretty much any operation.  For instance,

```
Buffer.from(Array(1))  // Is a buffer
Buffer.from(Array(1)).map(identity)  // Now just Seq
```

Buffer should be updated, if possible, to remember it is a buffer.  (The underlying type in both cases is `ArrayBuffer`; this is entirely the compile-time side that is losing track of identity.)

marcelocenerine:
Fixes #396


Ichoran:
LGTM.  Thanks!

Issue #400, opened 2018-02-03
+=: on ArrayBuffer can be a no-op


Ichoran:
ArrayBuffer doesn't prepend an element when the `+=:` method is called.  It instead slides everything down but doesn't increase the length (as far as I can tell)

```
val b = ArrayBuffer.empty[Int]
1 +=: b
b
```

demonstrates the problem.

```
val b = ArrayBuffer(2)
1 +=: b
b
```

gives `ArrayBuffer(1)`.  Probably the length isn't being updated?

Ichoran:
Issue #411 might be the deeper problem that is causing this one.

julienrf:
Fixed by #417.

PR #417, opened 2018-02-05
Increment end in Buffer.insert

### LINKED ISSUE #411: `insert` on mutable collections doesn't insert in empty `ArrayBuffer`
```
val xs = ArrayBuffer.empty[String]
xs.insert(0, "fish")
xs
```

shows `xs` to still be empty.  On a non-empty `ArrayBuffer`, it shifts the last element off the end instead of increasing the size.

marcelocenerine:
Fixes #411 & #400


Ichoran:
LGTM.  Thanks!

Issue #410, opened 2018-02-04
dropInPlace and dropRightInPlace are unsafe


Ichoran:
They should mimic the behavior of `drop` and `dropRight`, where elements beyond the end of the collection (or negative numbers of elements) are ignored.  Presently, an IndexOutOfBoundsException is thrown instead.

Sample code:

```
ArrayBuffer(1, 2).dropRightInPlace(4)
ArrayBuffer(1, 2).dropInPlace(4)
```

These should both clear out the `ArrayBuffer`.  (`ArrayBuffer` is not the only class affected.)

julienrf:
Fixed by #423.

Issue #409, opened 2018-02-04
takeInPlace overshoots bounds


Ichoran:
```
sm.ArrayBuffer(1, 2).takeInPlace(1)
```

fails with an exception.

julienrf:
Fixed by #423.

Issue #406, opened 2018-02-04
`takeWhileInPlace` overshoots bounds


Ichoran:
```
ArrayBuffer(true, false).takeWhileInPlace(_ == true)
```

throws an out of bounds exception (as do all the mutable sequences).

(`ListBuffer` also throws an index out of bounds exception--should check to make sure this is working efficiently.)

Ichoran:
Fix #409 first--this one is probably a consequence of that?

marcelocenerine:
Yes indeed.

julienrf:
Fixed by #423.

PR #423, opened 2018-02-05
Bug fixes for Buffer/ArrayBuffer/ListBuffer

### LINKED ISSUE #413: filterInPlace throws exceptions and/or fails on many inputs
```
ArrayBuffer(1, 2).filterInPlace(_ => false)
```

is one such example of throwing an exception.  (`ArrayBuffer` always throws.)

`ListBuffer` returns incorrect results.

```
ListBuffer(1, 2).filterInPlace(_ => false)
```

returns `ListBuffer(2)`.

If anything else has `filterInPlace`, it is okay.

marcelocenerine:
This PR fixes #413, #409, #410, #406

My original goal was to fix #413 only. It turns out that I had to fix other bugs along the way to allow the original one to be fixed or was somehow related. Some of these bugs have issues and others don't (e.g. `ListBuffer.remove(idx)`, `Buffer.trimStart/trimEnd`, ...). Each fix is in a separate commit to ease the code review

There are still some bugs in those classes which are covered by other issues that @Ichoran has already filed

### DIFF
--- a/collections/src/main/scala/strawman/collection/mutable/Buffer.scala
+++ b/collections/src/main/scala/strawman/collection/mutable/Buffer.scala
@@ -1,7 +1,7 @@
 package strawman.collection
 package mutable
 
-import scala.{Int, `inline`, throws, IndexOutOfBoundsException, IllegalArgumentException, Unit, Boolean, Array, deprecated}
+import scala.{Int, `inline`, throws, IndexOutOfBoundsException, IllegalArgumentException, Unit, Boolean, Array, deprecated, math}
 import scala.Predef.intWrapper
 
 /** A `Buffer` is a growable and shrinkable `Seq`. */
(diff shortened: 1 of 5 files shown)


collections/src/main/scala/strawman/collection/mutable/Buffer.scala
```diff
@@ -117,17 +129,18 @@ trait IndexedOptimizedBuffer[A] extends IndexedOptimizedSeq[A] with Buffer[A] {
...
+          this(j) = this(i)
+        }
         j += 1
       }
       i += 1
     }
-    takeInPlace(j)
+
+    if (i == j) this else takeInPlace(j)
   }
```
marcelocenerine: #413

julienrf:
Awesome, thanks!

PR #425, opened 2018-02-05
specific C for StringBuilder

lrytz:


collections/src/main/scala/strawman/collection/mutable/Builder.scala
```diff
@@ -87,10 +88,10 @@ class StringBuilder(private val sb: java.lang.StringBuilder) extends Builder[Cha
...
   def apply(i: Int): Char = sb.charAt(i)
   def iterableFactory: strawman.collection.SeqFactory[IndexedSeq] = IndexedSeq
-  protected[this] def fromSpecificIterable(coll: strawman.collection.Iterable[Char]): IndexedSeq[Char] =
-    iterableFactory.from(coll)
-  protected[this] def newSpecificBuilder(): strawman.collection.mutable.Builder[Char, IndexedSeq[Char]] =
-    iterableFactory.newBuilder()
+  protected[this] def fromSpecificIterable(coll: strawman.collection.Iterable[Char]): StringBuilder =
+    new StringBuilder() ++= coll
+  protected[this] def newSpecificBuilder(): strawman.collection.mutable.Builder[Char, StringBuilder] =
+    IndexedSeq.newBuilder[Char]().mapResult(r => new StringBuilder() ++= r)
```
lrytz: should this be more efficient?

test/junit/src/test/scala/strawman/collection/mutable/StringBuilderTest.scala
```diff
@@ -0,0 +1,24 @@
...
+  @Test
+  def specificBuilder: Unit = {
+    val b = new StringBuilder() ++= "abcd"
+    val b1 = b.filter(_ % 2 == 0)
+    val b1t: StringBuilder = b1
+    assertEquals(b1t.toString, "bd")
+
+    val b2 = b.map(c => (c + 1).toChar)
+    val b2t: IndexedSeq[Char] = b2
+    assertEquals(b2t.toString, "ArrayBuffer(b, c, d, e)")
```
lrytz: to improve this i'd have to add an overload for `map` i guess? what other operations have to be added manually in for such collections? `flatMap`, ...? (how would a user / collection implementer find out?)

lrytz: for the record, the behavior as tested here is the same as in 2.12

lrytz:
I pushed a second commit that brings back more methods on StringBuilder that existed in the 2.12 version. Let me what you think.

collections/src/main/scala/strawman/collection/mutable/Builder.scala
julienrf: Maybe `IndexedSeq.newBuilder[Char]().mapResult(r => new StringBuilder(r))`.

test/junit/src/test/scala/strawman/collection/mutable/StringBuilderTest.scala
julienrf: > what other operations have to be added manually in for such collections? flatMap, ...? (how would a user / collection implementer find out?)

All the operations that return a `CC[B]`. I’ve just had a look and several of them are missing overloads (e.g. `scanLeft`)…

I’ve added the following entry in the FAQ: https://github.com/scala/collection-strawman/wiki/FAQ#which-methods-should-i-overload-to-support-the-same-result-type-principle

collections/src/main/scala/strawman/collection/mutable/Builder.scala
szeiger: I'd use a `GrowableBuilder` (like in 2.12) so that `StringBuilder` can build itself without copying on `result()`

collections/src/main/scala/strawman/collection/mutable/Builder.scala
```diff
@@ -309,4 +311,119 @@ class StringBuilder(private val sb: java.lang.StringBuilder) extends Builder[Cha
   def setLength(len: Int): Unit = sb.setLength(len)
 
   def update(idx: Int, elem: Char): Unit = sb.setCharAt(idx, elem)
+
+
+  /** Like reverse, but destructively updates the target StringBuilder.
+   *
+   *  @return   the reversed StringBuilder (same as the target StringBuilder)
+   */
+  def reverseContents(): StringBuilder = {
```
szeiger: reverseInPlace would be a more consistent name in the new design, and these methods now return `this.type`

collections/src/main/scala/strawman/collection/mutable/Builder.scala
```diff
@@ -309,4 +311,119 @@ class StringBuilder(private val sb: java.lang.StringBuilder) extends Builder[Cha
...
+  def charAt(index: Int): Char = sb.charAt(index)
+
+  /** Removes the Char at the specified index.  The sequence is
+   *  shortened by one.
+   *
+   *  @param  index  The index to remove.
+   *  @return        This StringBuilder.
+   *  @throws IndexOutOfBoundsException  if the index is out of bounds.
+   */
+  def deleteCharAt(index: Int): StringBuilder = {
```
szeiger: also `this.type`

collections/src/main/scala/strawman/collection/mutable/Builder.scala
```diff
@@ -309,4 +311,119 @@ class StringBuilder(private val sb: java.lang.StringBuilder) extends Builder[Cha
...
+    this
+  }
+
+  /** Update the sequence at the given index to hold the specified Char.
+   *
+   *  @param  index   the index to modify.
+   *  @param  ch      the new Char.
+   *  @throws IndexOutOfBoundsException  if the index is out of bounds.
+   */
+  def setCharAt(index: Int, ch: Char): Unit = sb.setCharAt(index, ch)
```
szeiger: Why does this method return `Unit`?

lrytz:
@szeiger updated

julienrf:
@lrytz would you mind rebasing and fixing conflicts?

lrytz:
Done!

PR #429, opened 2018-02-06
Fix problems caused by merging #336

szeiger:
The branch was outdated without causing proper merge conflicts.


szeiger:
push build fails repeatedly, looks like some network problem

PR #430, opened 2018-02-06
mutable.PriorityQueue from the old collections

szeiger:
We don’t have ResizableArray anymore so we use ArrayBuffer instead.
It wasn’t clear to me why PriorityQueue didn’t use ArrayBuffer in the
first place rather than a custom implementation of ResizableQueue.


julienrf:
Note that a `SortedMultiSet` can also be used as a priority queue. (Although I guess this implementation is more performant)

PR #434, opened 2018-02-08
Bunch of bug fixes for `ArrayBuffer` & `ListBuffer`

### LINKED ISSUE #385: ArrayBuffer#clear still references “removed” elements
As mentioned in this comment: https://github.com/scala/collection-strawman/pull/49#issuecomment-360196113
This behavior prevents the removed elements to be GC-ed.

### LINKED ISSUE #398: Patch throws an exception on out of bounds indices
Patch should be completely safe, resolving bad parameter values by clipping them to safe values.

However, the current implementation throws exceptions instead.  For instance,

```
Buffer.from(Array(1)).patch(-1, LazyList.empty[Int], 2)
```

throws an exception.

### LINKED ISSUE #407: patchInPlace throws a class cast exception
```
java.lang.ClassCastException: strawman.collection.immutable.Nil$ cannot be cast to strawman.collection.immutable.$colon$colon
strawman.collection.mutable.ListBuffer.locate(ListBuffer.scala:153)
strawman.collection.mutable.ListBuffer.patchInPlace(ListBuffer.scala:272)
```

The call was to `ArrayBuffer.empty[Int].patch(-1, ArrayBuffer.empty[Int], -1)`, but this shouldn't happen regardless.

marcelocenerine:
- fix #398 (`Buffer.patch` throws out of bound index exception) by dropping validation in `View.Patched`
- fix #407 (`ListBuffer.patchInPlace` throws class cast exception) by clipping index/count to safe values
- fix various issues in `ArrayBuffer.patchInPlace` causing out of bound index exceptions
- fix #385 (`ArrayBuffer.clear` not dereferencing cleared elements)
- Ensure that index passed into `ArrayBuffer`'s `apply` and `update` is within bounds. This fix uncovered other issues in `addOne` and `insert`

Note: each fix is in a separate commit

### DIFF
--- a/collections/src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
+++ b/collections/src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
@@ -4,7 +4,7 @@ package mutable
 
 import java.lang.{IndexOutOfBoundsException, IllegalArgumentException}
 
-import scala.{AnyRef, Array, ArrayIndexOutOfBoundsException, Boolean, Exception, Int, Long, StringContext, Unit, math, Any, throws, Serializable, SerialVersionUID}
+import scala.{AnyRef, Array, ArrayIndexOutOfBoundsException, Boolean, Exception, Int, `inline`, Long, StringContext, Unit, math, Any, throws, Serializable, SerialVersionUID}
 import scala.Predef.intWrapper
 
 /** An implementation of the `Buffer` class using an array to
@@ -30,7 +30,7 @@ import scala.Predef.intWrapper
   *  @define willNotTerminateInf
   */
 @SerialVersionUID(1529165946227428979L)
-class ArrayBuffer[A] private (initElems: Array[AnyRef], initLength: Int)
+class ArrayBuffer[A] private (initElems: Array[AnyRef], initSize: Int)
   extends AbstractBuffer[A]
     with IndexedSeq[A]
     with IndexedSeqOps[A, ArrayBuffer, ArrayBuffer[A]]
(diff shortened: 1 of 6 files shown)


collections/src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
```diff
@@ -40,48 +40,55 @@ class ArrayBuffer[A] private (initElems: Array[AnyRef], initLength: Int)
 
   def this() = this(new Array[AnyRef](16), 0)
 
-  def this(initLength: Int) = this(new Array[AnyRef](initLength), 0)
+  def this(initSize: Int) = this(new Array[AnyRef](initSize), 0)
 
   private var array: Array[AnyRef] = initElems
-  private var end = initLength
+  private var size0 = initSize
```
marcelocenerine: renamed `end` to `size0` as the former is misleading

collections/src/main/scala/strawman/collection/mutable/ArrayBuffer.scala
```diff
@@ -40,48 +40,55 @@ class ArrayBuffer[A] private (initElems: Array[AnyRef], initLength: Int)
...
     if (lo < 0) throw new IndexOutOfBoundsException(lo.toString)
-    if (hi > end) throw new IndexOutOfBoundsException(hi.toString)
+    if (hi > size0) throw new IndexOutOfBoundsException(hi.toString)
   }
 
   @throws[ArrayIndexOutOfBoundsException]
-  def apply(n: Int) = array(n).asInstanceOf[A]
+  def apply(n: Int) = {
+    checkWithinBounds(n, n + 1)
+    array(n).asInstanceOf[A]
```
julienrf: Aren’t bounds already checked by the underlying array access?

marcelocenerine: the underlying array can be larger than the `ArrayBuffer` in many situations:

```
import strawman.collection.mutable.ArrayBuffer
val a = ArrayBuffer(1)
a: strawman.collection.mutable.ArrayBuffer[Int] = ArrayBuffer(1)

scala> a.remove(0)
res2: Int = 1

scala> a(0)
res3: Int = 0

scala> val b = ArrayBuffer.empty[Int]
b: strawman.collection.mutable.ArrayBuffer[Int] = ArrayBuffer()

scala> b(0)
res4: Int = 0

scala> val c = ArrayBuffer(1, 2, 3)
c: strawman.collection.mutable.ArrayBuffer[Int] = ArrayBuffer(1, 2, 3)

scala> c.clear()

scala> c.size
res6: Int = 0

scala> c(0)
res7: Int = 1
```

marcelocenerine: in summary, when an `ArrayBuffer` is created empty, its underlying array's size will be 16. When an `ArrayBuffer` is shrunken (remove/clear), the backed array remains with the same size

julienrf:
Thanks Marcelo :)

PR #439, opened 2018-02-09
Add back optimizations of HashSet

### LINKED ISSUE #380: immutable.HashSet.subsetOf should not invoke hashCode
This test fails: https://github.com/scala/scala/blob/2.13.x/test/files/run/t7326.scala#L45-L61
Added in https://github.com/scala/scala/pull/3315

julienrf:
Fixes #380 (I checked it locally but didn’t duplicate the test in our repository).

For some reason when I ported the code of `HashSet` to the new design I only took the minimum amount of code. This PR also takes optimizations from the old collections and put them in the new collections.

### DIFF
--- a/benchmarks/time/src/main/scala/benchmarks/immutableHashSet/package.scala
+++ b/benchmarks/time/src/main/scala/benchmarks/immutableHashSet/package.scala
@@ -1,0 +1,127 @@
+package benchmarks.immutableHashSet
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations._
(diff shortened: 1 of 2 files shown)


szeiger:
What's with all the `private[immutable]` methods? Are they supposed to be called from outside of `HashSet`?

julienrf:
I copy-pasted from the old collections without making them `protected` but indeed they should be `protected`, unless maybe there are some tests somewhere in the old collections that rely on these implementation details?

julienrf:
I’ve modified the PR to use `protected` when possible.

PR #443, opened 2018-02-09
Revert length / size changes

szeiger:
This reverts https://github.com/scala/collection-strawman/pull/356 which causes compatibility issues requiring unnecessary changes in every Seq implementation for no obvious benefit.

### DIFF
--- a/collections/src/main/scala/strawman/collection/Seq.scala
+++ b/collections/src/main/scala/strawman/collection/Seq.scala
@@ -82,33 +82,13 @@ object Seq extends SeqFactory.Delegate[Seq](immutable.Seq)
   * @define Coll `Seq`
   */
 trait SeqOps[+A, +CC[_], +C] extends Any
-  with IterableOps[A, CC, C] {
+  with IterableOps[A, CC, C]
+  with ArrayLike[A] {
 
   /**
     * @return This collection as a `Seq[A]`. This is equivalent to `to(Seq)` but might be faster.
     */
   def toSeq: Seq[A]
-
-  /** Selects an element by its index in the $coll.
-    *
-    * Example:
-    *
-    * {{{
-    *    scala> val x = List(1, 2, 3, 4, 5)
-    *    x: List[Int] = List(1, 2, 3, 4, 5)
-    *
-    *    scala> x(3)
-    *    res1: Int = 4
-    * }}}
-    *
-    *  @param  idx  The index to select.
-    *  @return the element of this $coll at index `idx`, where `0` indicates the first element.
-    *  @throws IndexOutOfBoundsException if `idx` does not satisfy `0 <= idx < length`.
-    */
-  def apply(idx: Int): A
-
-  /** Alias for `size` */
-  @`inline` final def length: Int = size
 
   /** A copy of the $coll with an element prepended.
     *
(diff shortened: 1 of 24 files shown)


collections/src/main/scala/strawman/collection/Seq.scala
```diff
@@ -82,34 +82,14 @@ object Seq extends SeqFactory.Delegate[Seq](immutable.Seq)
   * @define Coll `Seq`
   */
 trait SeqOps[+A, +CC[_], +C] extends Any
-  with IterableOps[A, CC, C] {
+  with IterableOps[A, CC, C]
+  with ArrayLike[A] {
```
julienrf: I’m not sure it’s a good idea to have `Seq` extend `ArrayLike`.

`ArrayLike` suggests that the size can be known in O(1), but that’s not the case for all `Seq`s (e.g. `LinearSeq`).

collections/src/main/scala/strawman/collection/View.scala
```diff
@@ -313,16 +312,16 @@ trait IndexedView[+A] extends View[A] with ArrayLike[A] with SeqOps[A, View, Vie
...
+    def hasNext = current < self.length
     def next(): A = {
       val r = self.apply(current)
       current += 1
       r
     }
   }
 
-  final override def knownSize: Int = finiteSize
+  override def knownSize: Int = length
```
julienrf: Why not `final`?

collections/src/main/scala/strawman/collection/mutable/Builder.scala
```diff
@@ -92,7 +92,8 @@ class StringBuilder(private val sb: java.lang.StringBuilder) extends Builder[Cha
   protected[this] def newSpecificBuilder(): strawman.collection.mutable.Builder[Char, IndexedSeq[Char]] =
     iterableFactory.newBuilder()
 
-  protected def finiteSize: Int = sb.length()
+  //TODO In the old collections, StringBuilder extends Seq -- should it do the same here to get this method?
```
julienrf: cf #425.

julienrf:
Also, I find it a bit non-consistent that to override the “number of elements” of a `Set` you have to override `size`, but for a `Seq` you have to override `length`. I prefer to consistently have to override the same method in all branches.

szeiger:
`ArrayLike` in 2.12:

```
A common supertrait of `ArrayOps` and `WrappedArray` that factors out the 
 * `deep` method for arrays and wrapped arrays and serves as a marker trait
 * for array wrappers.
```

And we currently have:

```
A trait representing indexable collections with finite length
```

There's nothing about O(1) access in this but the old "array wrapper" definition could be read to imply it. With the `deep` method gone we should simply remove `ArrayLike` instead of repurposing it. It was not meant as an API trait in the first place and there are 0 references to it in the compiler/library sources and tests.

szeiger:
> Also, I find it a bit non-consistent that to override the “number of elements” of a Set you have to override size, but for a Seq you have to override length. I prefer to consistently have to override the same method in all branches.

This is consistent with the way we handle similar cases where a more specific concept (`length` in this case) is used to implement an inherited less specific concept (`size`). For example, see `Iterable.concat` vs `Seq.appendedAll`. You are expected to implement the specific concept for the collection and the inherited one as a final alias.

Issue #418, opened 2018-02-05
Harmonize use of `@SerialVersionUID`


lrytz:
There won't be any serialization compatibility between 2.12 and 2.13 collections, so we should make sure to use `@SerialVersionUID` numbers different than in 2.12. Other than that, we can pick any number - there's no advantage in picking the number computed by `serialver`.

So I propose we just use `@SerialVersionUID(3L)` across the entire 2.13 collections (there are some uses of `1L` and `2L` in 2.12 collections).

See also https://github.com/scala/scala-dev/issues/237

julienrf:
Fixed by #444.

PR #444, opened 2018-02-10
Harmonize use of @SerialVersionUID

marcelocenerine:
This PR addresses #418:

- set all `@SerialVersionUID` to 3
- reviewed all classes implementing `Serializable` and added missing `@SerialVersionUID`
- made the following classes serializable: `immutable.LongMap`, `immutable.IntMap`, `immutable.Map.WithDefault`, `immutable.SortedMap.WithDefault`, `mutable.Map.WithDefault`, `mutable.SortedMap.WithDefault`

There are a few things left/questions:

- are the Java <-> Scala wrappers meant to be serializable? Almost all of them are case classes which auto-implement `Serializable`. However, most do not have a serial version uid: 
  - DictionaryWrapper
  - IterableWrapper
  - IteratorWrapper
  - JCollectionWrapper
  - JConcurrentMapWrapper
  - JDictionaryWrapper
  - JEnumerationWrapper
  - JIterableWrapper
  - JIteratorWrapper
  - JListWrapper
  - JMapWrapper
  - JPropertiesWrapper
  - JSetWrapper
  - WeakHashMap
  - MutableMapWrapper
  - ConcurrentMapWrapper
  - MutableBufferWrapper
  - MutableSeqWrapper
  - MutableSetWrapper
  - SeqWrapper
  - MutableSetWrapper

- The following classes are the type of sets returned by `Map.keySet`. These classes are serializable but do not have a serial version number either. An interesting detail is that they hold references to the concrete class where `MapOps` is mixed in, which would end up being part of the serialized form of the former:

  - KeySet
  - LinkedKeySet
  - KeySortedSet
  - ImmutableKeySortedSet
  - KeySortedSet
  - ImmutableKeySet
  - ImmutableKeySortedSet

Example:
```
scala> printOutputBytes(immutable.Map(1 -> 1).keySet)
��sr4strawman.collection.immutable.MapOps$ImmutableKeySet��kT�rL$outert&Lstrawman/collection/immutable/MapOps;xpsr*strawman.collection.immutable.ChampHashMapIcachedJavaKeySetHashCodeI
cachedSizerootNodet'Lstrawman/collection/immutable/MapNode;xp����sr2strawman.collection.immutable.BitmapIndexedMapNodeIdataMapInodeMap[contentt[Ljava/lang/Object;xr%strawman.collection.immutable.MapNodexp�ur[Ljava.lang.Object;��X�s)lxpsrjava.lang.Integer⠤���8Ivaluexrjava.lang.Number���
                                                                                                            ���xpq~


scala> serializeDeserialize(immutable.TreeMap(1 -> 1).keySet)
��sr@strawman.collection.immutable.SortedMapOps$ImmutableKeySortedSet�3,ؙ�L$outert,Lstrawman/collection/immutable/SortedMapOps;xpsr%strawman.collection.immutable.TreeMaporderingtLscala/math/Ordering;Ltreet1Lstrawman/collection/immutable/RedBlackTree$Tree;xpsrscala.math.Ordering$Int$��܂����xpsr4strawman.collection.immutable.RedBlackTree$BlackTreexr/strawman.collection.immutable.RedBlackTree$TreeIcountLkeytLjava/lang/Object;Lleftq~Lrightq~Lvalueq~
                                                                                            xpsrjava.lang.Integer⠤���8Ivaluexrjava.lang.Number���
                                                                                                                                                 ���xpppq~
```

Java's `Map` classes use similar strategy by returning inner classes which implement `Set`. However, those are **not** serializable:

```
scala> val j = new java.util.HashMap[Int, Int]()
j: java.util.HashMap[Int,Int] = {}

scala> serializeDeserialize(j.keySet)
java.io.NotSerializableException: java.util.HashMap$KeySet
  at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
  at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:348)
  at .serializeDeserialize(<console>:18)
  ... 36 elided
```

Current Scala collections implement `keySet` similarly to Strawman in the sense that they return instances of anonymous classes holding references to the enclosing `Map`, are serializable but don't define a fixed serial version uid.

I was wondering what would be the best approach here: 1) keep things as is and just add the serial version uid annotation to the `KeySet` classes; 2) scrap those classes and replace implementation of `keySet` with something like `def keySet: Set[K] = Set.from(View.Map(toIterable, { case (k: K, _) => k }))` (it would make it strict though); 3) make `KeySet` not serializable

&nbsp;&nbsp;
- The final one: `NumericRange.mapRange` returns an anonymous class which has access to enclosing fields. That could be turned into a class in the companion object, but would require the mapping function to be serialized alongside with it, which in turn would potentially hold references to fields/variables in the enclosing scope of code using it. Any suggestion?

### DIFF
--- a/test/junit/src/test/scala/strawman/collection/SeqTest.scala
+++ b/test/junit/src/test/scala/strawman/collection/SeqTest.scala
@@ -63,4 +63,20 @@ class SeqTest {
     assertEquals(1, Vector(0, 1, 2, 0, 1, 2).lastIndexOfSlice(Vector(1, 2), end = 3))
     assertEquals(-1, List(0, 1).lastIndexOfSlice(List(1, 2)))
   }
+
+  @Test
+  def hasCorrectDiff(): Unit = {
+    val s1 = Seq(1, 2, 3, 4, 5)
+    val s2 = Seq(1, 3, 5, 7, 9)
+
+    assertEquals(Seq(2, 4), s1.diff(s2))
+  }
+
+  @Test
+  def hasCorrectIntersect(): Unit = {
+    val s1 = Seq(1, 2, 3, 4, 5)
+    val s2 = Seq(1, 3, 5, 7, 9)
+
+    assertEquals(Seq(1, 3, 5), s1.intersect(s2))
+  }
 }
--- a/collections/src/main/scala/strawman/collection/Map.scala
+++ b/collections/src/main/scala/strawman/collection/Map.scala
@@ -3,7 +3,7 @@ package collection
 
 import collection.mutable.Builder
 
-import scala.{Any, Boolean, ClassCastException, Equals, Int, NoSuchElementException, None, Nothing, Option, Ordering, PartialFunction, Serializable, Some, StringContext, `inline`, throws}
+import scala.{Any, Boolean, ClassCastException, Equals, Int, NoSuchElementException, None, Nothing, Option, Ordering, PartialFunction, Serializable, SerialVersionUID, Some, StringContext, `inline`, throws}
 import scala.Predef.String
 import scala.annotation.unchecked.uncheckedVariance
 import scala.language.implicitConversions
@@ -105,6 +105,7 @@ trait MapOps[K, +V, +CC[X, Y] <: MapOps[X, Y, CC, _], +C]
 
   /** The implementation class of the set returned by `keySet`.
     */
+  @SerialVersionUID(3L)
   protected class KeySet extends Set[K] with GenKeySet {
     def iterableFactory: IterableFactory[Set] = Set
     protected[this] def fromSpecificIterable(coll: Iterable[K]): Set[K] = fromIterable(coll)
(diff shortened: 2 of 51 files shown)


julienrf:
1. Java wrappers: let’s add a serial version uid, that doesn’t cost anything!
2. keysets: the inner classes avoid the creation of *one* intermediate `Set`. I guess this is useful in some situation (if you want to check something or search something on that set). I would leave them as they are (just add the version uid).
3. `mapRange` is only used internally and the passed function has no free identifiers. We could just add a comment about that somewhere.

marcelocenerine:
@julienrf, cool! I added the serial version uid to the remaining classes.

test/junit/src/test/scala/strawman/collection/SeqTest.scala
```diff
@@ -63,4 +63,20 @@ class SeqTest {
...
+    assertEquals(Seq(2, 4), s1.diff(s2))
+  }
+
+  @Test
+  def hasCorrectIntersect(): Unit = {
+    val s1 = Seq(1, 2, 3, 4, 5)
+    val s2 = Seq(1, 3, 5, 7, 9)
+
+    assertEquals(Seq(1, 3, 5), s1.intersect(s2))
+  }
```
julienrf: Are these tests related with the changes of this PR?

marcelocenerine: I added these tests because of this change: https://github.com/scala/collection-strawman/pull/444/files#diff-d2679eddf2a5dee23263f82189f93c81 . It turns out that it was not necessary at all as the anonymous class is only used internally. The behavior is the same though.

julienrf:
Thanks @marcelocenerine!

Issue #420, opened 2018-02-05
LazyList.toString SOE


lrytz:
```
scala> val l = LazyList.from(1 to 10000)
l: scala.collection.immutable.LazyList[Int] = LazyList(?)

scala> l.foreach(x => ())

scala> l
java.lang.StackOverflowError
  at scala.collection.immutable.LazyList$Cons.toString(LazyList.scala:480)
  at scala.collection.immutable.LazyList$Cons.toString(LazyList.scala:480)
  at scala.collection.immutable.LazyList$Cons.toString(LazyList.scala:480)
```

lrytz:
Very funny thing:

I can reproduce this behavior when starting a new repl (with `JAVA_OPS` empty). But after running it for a few times, the JVM will inline / optimize something to make it actually work :-)

```
Welcome to Scala 2.13.0-pre-5ba8d97 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_162).
Type in expressions for evaluation. Or try :help.

scala> val l = LazyList.from(1 to 10000)
l: scala.collection.immutable.LazyList[Int] = LazyList(?)

scala> l.foreach(x => ())

scala> try { l.toString; println(s"worked") } catch { case _: StackOverflowError => println(s"failed") }
failed

scala> val l = LazyList.from(1 to 10000)
l: scala.collection.immutable.LazyList[Int] = LazyList(?)

scala> l.foreach(x => ())

scala> try { l.toString; println(s"worked") } catch { case _: StackOverflowError => println(s"failed") }
failed

scala> val l = LazyList.from(1 to 10000)
l: scala.collection.immutable.LazyList[Int] = LazyList(?)

scala> l.foreach(x => ())

scala> try { l.toString; println(s"worked") } catch { case _: StackOverflowError => println(s"failed") }
failed

scala> val l = LazyList.from(1 to 10000)
l: scala.collection.immutable.LazyList[Int] = LazyList(?)

scala> l.foreach(x => ())

scala> try { l.toString; println(s"worked") } catch { case _: StackOverflowError => println(s"failed") }
failed

scala> val l = LazyList.from(1 to 10000)
l: scala.collection.immutable.LazyList[Int] = LazyList(?)

scala> l.foreach(x => ())

scala> try { l.toString; println(s"worked") } catch { case _: StackOverflowError => println(s"failed") }
failed

scala> val l = LazyList.from(1 to 10000)
[...]

PR #445, opened 2018-02-11
Lazylist toString

dhanesharole:


marcelocenerine:
Hey Dhanesh! You probably need to rebase your master and reset your branch as commits from previous PR are showing up here.

My suggestion is to favor a more stack safe approach (tail recursive/loop) so that we can kill 2 birds with one stone. Otherwise it's very likely that there will be some rework in order to fix #420. I agree they are separate issues but quite related. I think `Iterator.addString`/`Iterable.mkString` can potentially give us some inspiration. What do you think, @julienrf?

dhanesharole:
@marcelocenerine , yeah makes sense. I will think about how to do tailrec solution for this. Will have a look at `Iterables`. 

Looks like tracking repo's git history is bit screwed. I shall create a new PR by rebasing these commits. Extremely sorry for that.

PR #447, opened 2018-02-12
More minor bug and compatibility fixes from my integration branch

szeiger:


collections/src/main/scala/strawman/collection/immutable/LongMap.scala
```diff
@@ -49,6 +49,18 @@ object LongMap {
   def apply[T](elems: (Long, T)*): LongMap[T] =
     elems.foldLeft(empty[T])((x, y) => x.updated(y._1, y._2))
 
+  def from[V](coll: IterableOnce[(Long, V)]): LongMap[V] = {
+    val b = newBuilder[V]()
+    b.sizeHint(coll)
```
julienrf: This is not really useful because in `ImmutableBuilder` this is a no-op.

collections/src/main/scala/strawman/collection/immutable/Map.scala
```diff
@@ -89,7 +89,7 @@ trait MapOps[K, +V, +CC[X, +Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]
     * @tparam V1 the type of the value in the key/value pair.
     * @return A new map with the new binding added to this map.
     */
-  /*@`inline` final*/ def + [V1 >: V](kv: (K, V1)): CC[K, V1] = updated(kv._1, kv._2)
+  override def + [V1 >: V](kv: (K, V1)): CC[K, V1] = updated(kv._1, kv._2)
```
julienrf: I’m a bit lost: what’s the base class that defines the method overridden here?

collections/src/main/scala/strawman/collection/immutable/Set.scala
```diff
@@ -9,7 +9,7 @@ import scala.{Any, Boolean, Int, deprecatedName, `inline`, None, Option, Seriali
 /** Base trait for immutable set collections */
 trait Set[A] extends Iterable[A] with collection.Set[A] with SetOps[A, Set, Set[A]] {
 
-  override final def toSet[B >: A]: Set[B] = this.asInstanceOf[Set[B]]
+  override final def toSet[B >: A]: Set[B] = Set.from(this)
```
julienrf: Why do we need to make a copy?

julienrf: Maybe we could make it a no-op in final subclasses of `Set` at least?

collections/src/main/scala/strawman/collection/mutable/UnrolledBuffer.scala
```diff
@@ -377,7 +377,7 @@ object UnrolledBuffer extends StrictOptimizedClassTagSeqFactory[UnrolledBuffer]
         // insert everything from iterable to this
         var curr = this
         var appended = 0
-        for (elem <- t) {
+        for (elem <- t.iterator()) {
```
julienrf: 👍

collections/src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -288,7 +288,7 @@ trait IterableOps[+A, +CC[_], +C] extends Any with IterableOnce[A] {
     *
     *  @return    `true` if the $coll contains at least one element, `false` otherwise.
     */
-  def nonEmpty: Boolean = iterator().hasNext
+  def nonEmpty: Boolean = !isEmpty
```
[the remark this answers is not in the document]
szeiger: > Implementing operations in terms of others can be confusing because then subclasses might override the wrong one

Yes, that's precisely the motivation. It was defined like this in 2.12 but without this change you would have to override *both* `nonEmpty` and `isEmpty` (which no existing collections do) which is never what you want.

collections/src/main/scala/strawman/collection/immutable/Set.scala
szeiger: `Set.from` doesn't copy when it's not necessary but we need to take "really invariant" subtypes of `Set` into account, which is what `Set.from` already does.

collections/src/main/scala/strawman/collection/immutable/Map.scala
szeiger: `collection.MapOps`

collections/src/main/scala/strawman/collection/Iterable.scala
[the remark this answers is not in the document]
szeiger: We can mark it `@deprecatedOverriding` for now and maybe make it `@inline final` later (but I doubt it makes much of a difference because the JIT can inline it anyway)

collections/src/main/scala/strawman/collection/immutable/Set.scala
szeiger: Actually we can just remove this override because the base implementation in `Iterable` already does the same.

collections/src/main/scala/strawman/collection/Iterable.scala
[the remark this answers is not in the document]
lrytz: I'm in favor of marking it final now, it's simple enough to fix custom subclasses, and this is hardly going to be the only thing that is source-incompatible.

collections/src/main/scala/strawman/collection/immutable/LongMap.scala
```diff
@@ -439,4 +447,7 @@ sealed abstract class LongMap[+T] extends Map[Long, T]
     case LongMap.Nil => throw new IllegalStateException("Empty set")
   }
 
+  def map[V2](f: ((Long, T)) => (Long, V2)): LongMap[V2] = LongMap.from(View.Map(coll, f))
+
+  def flatMap[V2](f: ((Long, T)) => IterableOnce[(Long, V2)]): LongMap[V2] = LongMap.from(View.FlatMap(coll, f))
```
lrytz: IntMap also overrides `concat`, it would probably make sense to do the same here.

PR #450, opened 2018-02-14
Added type parameter to Iterator.empty and test cases

kelebra:
Transfer of fix in [PR](https://github.com/scala/scala/pull/6321) for issue [#10711](https://github.com/scala/bug/issues/10711) in scala repository as per @szeiger suggestion

Also extended `IterableFactory` in `Iterator`'s companion object as was suggested in PR above.


test/junit/src/test/scala/strawman/collection/IteratorTest.scala
```diff
@@ -315,4 +315,40 @@ class IteratorTest {
...
+  @Test
+  def emptyIteratorInHigherOrderFunctions: Unit = {
+    val seqOfIterators = Seq(Seq(1, 2, 3).iterator(), Seq(3, 2, 1).iterator(), Seq(1, 3, 2).iterator())
+    val unified = seqOfIterators.foldLeft(Iterator.empty[Int])((a, b) => a ++ b)
+    assertEquals(List(1, 2, 3, 3, 2, 1, 1, 3, 2), List.from(unified))
+  }
+
+  @Test
+  def emptyIteratorBuilder: Unit = {
+    assertEquals(Iterator.empty[Int], Iterator.newBuilder[Int]().result())
```
mkeskells: use assertSame as results should be eq

test/junit/src/test/scala/strawman/collection/IteratorTest.scala
```diff
@@ -315,4 +315,40 @@ class IteratorTest {
 
     assertEquals("a,null,b,null,c,null", it.mkString(","))
   }
+
+  @Test
+  def emptyTypedIteratorsShouldBeEqual: Unit = {
+    val emptyDoubleIterator = Iterator.empty[Double]
+    val emptyIntIterator = Iterator.empty[Int]
+    assertEquals(emptyDoubleIterator, emptyIntIterator)
```
mkeskells: use assertSame when the results should be eq

collections/src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -1133,31 +1133,51 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+  /** Creates a target $coll from an existing source collection
+    *
+    * @param source Source collection
+    * @tparam A the type of the collection’s elements
+    * @return a new $coll with the elements of `source`
+    */
+  override def from[A](source: IterableOnce[A]): Iterator[A] = source.iterator()
+
+  /** The iterator which produces no values. */
+  @`inline` def empty[T]: Iterator[T] = _empty
```
mkeskells: inlining this method has little effect
If the method is inlinable it requires the caller to be monomorphically dispatched, which would imply that you have you pretty redundant code in this case

what you need to do in inline the target access. See https://github.com/rorygraves/scalac_perf/issues/46
maybe `@`inline` def empty[T]: Iterator[T] = _empty: @inline `
but you will need to check the bytecode :-(
@szeiger can you consider where else this should be applied, or discuss with @retronym if the optimiser should do this automatically

kelebra: All implementations below:

```scala
@inline final def empty[T]: Iterator[T] = _empty: @inline
@inline final def empty[T]: Iterator[T] = _empty
def empty[T]: Iterator[T] = _empty
```

result in the following bytecode:

```java
  public static <T> strawman.collection.Iterator<T> empty();
    Code:
       0: getstatic     #66                 // Field strawman/collection/Iterator$.MODULE$:Lstrawman/collection/Iterator$;
       3: invokevirtual #146                // Method strawman/collection/Iterator$.empty:()Lstrawman/collection/Iterator;
       6: areturn
```

As per client code is concerned I tried definition

```scala
@inline final def empty[T]: Iterator[T] = _empty: @inline
```

and the following client code:

```scala
object IteratorInlineTest {

  def inlineTest1: immutable.List[scala.Int] = immutable.List.from(Iterator.empty)

  def inlineTest2(iter: Iterator[scala.Int] = Iterator.empty): Iterator[scala.Int] =
    Iterator(iter, Iterator(3, 21), Iterator(1, 2, 34))
      .foldLeft(Iterator.empty[scala.Int])(_ ++ _)
  
  def inlineTest3: Iterator[scala.Int] = {
    var itr = Iterator.empty[Int]
    itr = itr ++ Iterator(123)
    itr = itr ++ Iterator(321)
    itr
  }
}
```

which resulted with the following bytecode:

```java
public final class strawman.collection.IteratorInlineTest {
  public static strawman.collection.Iterator<java.lang.Object> inlineTest2$default$1();
    Code:
       0: getstatic     #16                 // Field strawman/collection/IteratorInlineTest$.MODULE$:Lstrawman/collection/IteratorInlineTest$;
       3: invokevirtual #18                 // Method strawman/collection/IteratorInlineTest$.inlineTest2$default$1:()Lstrawman/collection/Iterator;
       6: areturn

  public static strawman.collection.Iterator<java.lang.Object> inlineTest3();
    Code:
       0: getstatic     #16                 // Field strawman/collection/IteratorInlineTest$.MODULE$:Lstrawman/collection/IteratorInlineTest$;
       3: invokevirtual #21                 // Method strawman/collection/IteratorInlineTest$.inlineTest3:()Lstrawman/collection/Iterator;
       6: areturn

  public static strawman.collection.Iterator<java.lang.Object> inlineTest2(strawman.collection.Iterator<java.lang.Object>);
    Code:
       0: getstatic     #16                 // Field strawman/collection/IteratorInlineTest$.MODULE$:Lstrawman/collection/IteratorInlineTest$;
       3: aload_0
       4: invokevirtual #25                 // Method strawman/collection/IteratorInlineTest$.inlineTest2:(Lstrawman/collection/Iterator;)Lstrawman/collection/Iterator;
       7: areturn

  public static strawman.collection.immutable.List<java.lang.Object> inlineTest1();
    Code:
       0: getstatic     #16                 // Field strawman/collection/IteratorInlineTest$.MODULE$:Lstrawman/collection/IteratorInlineTest$;
       3: invokevirtual #29                 // Method strawman/collection/IteratorInlineTest$.inlineTest1:()Lstrawman/collection/immutable/List;
       6: areturn
}
```

and actually I get the same bytecode in the case of current definition. Please advice if any other case has to be taken into consideration apart from these ones above.

test/junit/src/test/scala/strawman/collection/IteratorTest.scala
kelebra: Fixed, thanks for pointing that out.

kelebra:
Do not shoot the messenger: idea was initially proposed by @szeiger in scala PR. 

As per my understanding in terms of public API these two were different only in terms of `from(Iterable[A])` and `newBuilder[A]` methods which were trivial to implement. Given that and the fact that ~21 other companion objects extend it, it might make sense at least from uniformity standpoint.

test/junit/src/test/scala/strawman/collection/IteratorTest.scala
```diff
@@ -315,4 +315,40 @@ class IteratorTest {
...
+    iteratorBuilder += 3
+    assertEquals(List(5, 4, 3), List.from(iteratorBuilder.result()))
+  }
+
+  @Test
+  def nonEmptyIteratorAndClearBuilder: Unit = {
+    var iteratorBuilder = Iterator.newBuilder[Int]()
+    iteratorBuilder += 1
+    iteratorBuilder.clear()
+    assertEquals(Iterator.empty, iteratorBuilder.result())
```
julienrf: Why not `assertSame` here?

kelebra: Missed this one, thanks.

kelebra:
Is there anything required to merge this PR? 

Test cases were fixed (`assertSame` usage) and evidence of inlining was provided. I guess it is up to @mkeskells now.

julienrf:
Thanks!

mkeskells:
nothing is ever up to me - you overestimate my influence :-)

Issue #435, opened 2018-02-08
Array.transpose creates vectors


lrytz:
Should create arrays

```
scala> Array(Array(1.0,2.0), Array(3.0,4.0)).transpose
res3: scala.collection.immutable.IndexedSeq[scala.collection.immutable.IndexedSeq[Double]] = Vector(Vector(1.0, 3.0), Vector(2.0, 4.0))
```

julienrf:
Fixed by #452.

Issue #397, opened 2018-02-03
Mismatching signature of copyToArray


Ichoran:
The old collections have a `copyToArray` that takes two numeric parameters; the second one is the maximum number of elements to copy.  The new version is missing the second parameter.

```
List(1, 2, 3).copyToArray(new Array[Int](2), 1, 1)
```
should compile.

jiminhsieh:
If no one work on this one, I will try to give this a shot. :)

julienrf:
@jiminhsieh please go ahead!

julienrf:
Fixed by #452.

PR #452, opened 2018-02-15
Small improvements from my integration branch

lrytz:


collections/src/main/scala/strawman/collection/ArrayOps.scala
```diff
@@ -63,6 +68,8 @@ class ArrayOps[A](val xs: Array[A]) extends AnyVal
 
   def zip[B: ClassTag](that: Iterable[B]): Array[(A, B)] = fromTaggedIterable(View.Zip(toIterable, that))
 
+  def zipWithIndex(implicit ct: ClassTag[(A, Int)]): Array[(A, Int)] = fromTaggedIterable(View.ZipWithIndex(toIterable))
```
julienrf: How did you manage to solve the ambiguity with the inherited version?

lrytz: I misunderstood the error message, the ambiguity was on `WithFilter.map`, which has two overloads. I moved one in a parent trait in this PR.

collections/src/main/scala/strawman/collection/Iterator.scala
```diff
@@ -1123,6 +1123,52 @@ trait Iterator[+A] extends IterableOnce[A] { self =>
...
+   *
+   *  @note    Reuse: $consumesIterator
+   *
+   *  @usecase def copyToArray(xs: Array[A], start: Int): Unit
+   *
+   *    $willNotTerminateInf
+   */
+  @`inline` final def copyToArray[B >: A](xs: Array[B], start: Int = 0): xs.type = {
+    copyToArray(xs, start, xs.length)
+  }
```
julienrf: Actually what do you think of using a default parameter value for `len` as well instead of having an overload? (like #456)

julienrf:
Fixes #435. And Fixes #397.

collections/src/main/scala/strawman/collection/Iterator.scala
lrytz: Ah, that acutally points out that i didn't pick the correct default for `len`! It should have been `size`, not `xs.length`. However, we cannot use `size` (or `length`), as this consumes the iterator. So we have to use a different implementation, I'll push it here.

I think we should not use `size` in Iterable either, as we don't know the cost of the operation.

lrytz:
@julienrf updated `copyToArray` https://github.com/scala/collection-strawman/pull/452/commits/f3f370c2aad06a53fb28416457821b50084e865d

Issue #451, opened 2018-02-15
Sonatype's snapshot only has version 0.1


jiminhsieh:
It seems no longer publishing latest snapshot to Sonatype automatically. The only snapshot I can find on Sonatype is `0.1-SNAPSHOT`.
https://oss.sonatype.org/content/repositories/snapshots/ch/epfl/scala/collection-strawman_2.12/

julienrf:
Indeed, I don’t know why but our `after_success` hook is not called by Travis-CI…

I’ve just manually published a 0.10.0-SNAPSHOT but if someone knows what’s wrong with our .travis.yml file I’m happy to fix it.

SethTisue:
> if someone knows what’s wrong with our .travis.yml file 

maybe #453 will take care of it

PR #453, opened 2018-02-15
fix typo in .travis.yml

SethTisue:
perhaps this will make snapshots start publishing themselves again

references #451


julienrf:
good catch haha

Issue #421, opened 2018-02-05
LazyList.flatMap SOE


lrytz:
```
scala> val l = LazyList.from(1 to 10000)
l: scala.collection.immutable.LazyList[Int] = LazyList(?)

scala> l.flatMap(List(_)) == l
java.lang.StackOverflowError
  at scala.collection.Iterator$$anon$1.<init>(Iterator.scala:1150)
  at scala.collection.Iterator$.apply(Iterator.scala:1149)
  at scala.collection.View$Elems.iterator(View.scala:98)
  at scala.collection.mutable.Growable.addAll(Growable.scala:49)
  at scala.collection.mutable.Growable.addAll$(Growable.scala:48)
  at scala.collection.mutable.ListBuffer.addAll(ListBuffer.scala:34)
  at scala.collection.mutable.Growable.$plus$plus$eq(Growable.scala:57)
  at scala.collection.mutable.Growable.$plus$plus$eq$(Growable.scala:57)
  at scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:34)
  at scala.collection.mutable.ListBuffer$.from(ListBuffer.scala:284)
  at scala.collection.immutable.List$.from(List.scala:436)
  at scala.collection.immutable.List$.from(List.scala:429)
  at scala.collection.IterableFactoryLike.apply(Factory.scala:98)
  at scala.collection.IterableFactoryLike.apply$(Factory.scala:98)
  at scala.collection.immutable.List$.apply(List.scala:432)
  at .$anonfun$res0$1(<console>:1)
  at .$anonfun$res0$1$adapted(<console>:1)
  at scala.collection.immutable.LazyListOps.flatMap(LazyList.scala:347)
  at scala.collection.immutable.LazyListOps.flatMap$(LazyList.scala:342)
  at scala.collection.immutable.LazyList.flatMap(LazyList.scala:186)
  at scala.collection.immutable.LazyListOps.$anonfun$flatMap$1(LazyList.scala:355)
  at scala.collection.immutable.LazyListOps.lazyAppendAll(LazyList.scala:249)
  at scala.collection.immutable.LazyListOps.lazyAppendAll$(LazyList.scala:248)
  at scala.collection.immutable.LazyList.lazyAppendAll(LazyList.scala:186)
  at scala.collection.immutable.LazyListOps.$anonfun$lazyAppendAll$2(LazyList.scala:249)
[...]

PR #454, opened 2018-02-16
LazyList.force returns LazyList. 

### LINKED ISSUE #422: LazyList.force should not return an Option
It should just return a LazyList, and it should actually force the entire list (Not just the head)

optician:
Fixes #422.

* Add `force` tests for LazyList and Stream.
* Add `sameElements` tests for LazyList and Stream.
* Change `force` result type from `Option[(head, tail)]` to `CC[A]`.
* Force `force` to evaluate whole LazyList and Stream.
* Update methods depended on `force`: `unapply` and `sameElements`.
* Reimplement `sameElements`.
** Remove Tuple allocation.
** Check reference equality on every step.

### DIFF
--- a/collections/src/main/scala/strawman/collection/immutable/LazyList.scala
+++ b/collections/src/main/scala/strawman/collection/immutable/LazyList.scala
@@ -4,7 +4,7 @@ package immutable
 
 import strawman.collection.mutable.{ArrayBuffer, Builder}
 
-import scala.{Any, AnyRef, Boolean, Int, None, NoSuchElementException, noinline, Nothing, Option, PartialFunction, Some, StringContext, Unit, UnsupportedOperationException, deprecated}
+import scala.{Any, AnyRef, Boolean, Int, NoSuchElementException, None, Nothing, Option, PartialFunction, Some, StringContext, Unit, UnsupportedOperationException, deprecated, noinline}
 import scala.Predef.String
 import scala.annotation.tailrec
 import scala.annotation.unchecked.uncheckedVariance
@@ -235,8 +235,8 @@ sealed private[immutable] trait LazyListOps[+A, +CC[+X] <: LinearSeq[X] with Laz
 
   protected[this] def cons[T](hd: => T, tl: => CC[T]): CC[T]
 
-  /** Force the evaluation of both the head and the tail of this `LazyList` */
-  def force: Option[(A, CC[A])]
+  /** Force the evaluation of the whole `LazyList` */
+  def force: CC[A]
 
   override def nonEmpty: Boolean = !isEmpty
 
@@ -255,11 +255,12 @@ sealed private[immutable] trait LazyListOps[+A, +CC[+X] <: LinearSeq[X] with Laz
 
   override def sameElements[B >: A](that: IterableOnce[B]): Boolean = {
     @tailrec def lazyListEq(a: CC[_], b: CC[_]): Boolean =
-      if (a eq b) true else {
-        (a.force, b.force) match {
-          case (Some((ah, at)), Some((bh, bt))) => (ah == bh) && lazyListEq(at, bt)
-          case (None, None) => true
-          case _ => false
+      (a eq b) || {
+        if (a.nonEmpty && b.nonEmpty && a.head == b.head) {
+          lazyListEq(a.tail, b.tail)
+        }
+        else {
+          a.isEmpty && b.isEmpty
         }
       }
     that match {
@@ -441,7 +442,6 @@ sealed private[immutable] trait LazyListFactory[+CC[+X] <: LinearSeq[X] with Laz
     newCons(head, stream.tail.collect(pf))
   }
 
-  type Evaluated[+A] <: Option[(A, CC[A])]
 }
 
 /**
@@ -453,14 +453,12 @@ object LazyList extends LazyListFactory[LazyList] {
 
   protected[this] def newCons[T](hd: => T, tl: => LazyList[T]): LazyList[T] = new LazyList.Cons(hd, tl)
 
-  type Evaluated[+A] = Option[(A, LazyList[A])]
-
   object Empty extends LazyList[Nothing] {
     override def isEmpty: Boolean = true
     override def head: Nothing = throw new NoSuchElementException("head of empty lazy list")
     override def tail: LazyList[Nothing] = throw new UnsupportedOperationException("tail of empty lazy list")
-    def force: Evaluated[Nothing] = None
-    override def toString: String = "Empty"
+    override def force: LazyList[Nothing] = this
+    override def toString(): String = "Empty"
   }
 
   final class Cons[A](hd: => A, tl: => LazyList[A]) extends LazyList[A] {
@@ -475,8 +473,12 @@ object LazyList extends LazyListFactory[LazyList] {
       tlEvaluated = true
       tl
     }
-    def force: Evaluated[A] = Some((head, tail))
-    override def toString: String =
+    override def force: LazyList[A] = {
+      head
+      tail.force
+      this
+    }
+    override def toString(): String =
       if (hdEvaluated) s"$head #:: ${if (tlEvaluated) tail.toString else "?"}"
       else "LazyList(?)"
   }
@@ -506,7 +508,8 @@ object LazyList extends LazyListFactory[LazyList] {
   }
 
   object #:: {
-    def unapply[A](s: LazyList[A]): Evaluated[A] = s.force
+    def unapply[A](s: LazyList[A]): Option[(A, LazyList[A])] =
+      if (s.nonEmpty) Some(s.head, s.tail) else None
   }
 
   def from[A](coll: collection.IterableOnce[A]): LazyList[A] = coll match {
@@ -583,14 +586,12 @@ object Stream extends LazyListFactory[Stream] {
 
   protected[this] def newCons[T](hd: => T, tl: => Stream[T]): Stream[T] = new Stream.Cons(hd, tl)
 
-  type Evaluated[+A] = Option[(A, Stream[A])]
-
   object Empty extends Stream[Nothing] {
     override def isEmpty: Boolean = true
     override def head: Nothing = throw new NoSuchElementException("head of empty lazy list")
     override def tail: Stream[Nothing] = throw new UnsupportedOperationException("tail of empty lazy list")
-    def force: Evaluated[Nothing] = None
-    override def toString: String = "Empty"
+    override def force: Stream[Nothing] = this
+    override def toString(): String = "Empty"
   }
 
   final class Cons[A](override val head: A, tl: => Stream[A]) extends Stream[A] {
@@ -600,9 +601,13 @@ object Stream extends LazyListFactory[Stream] {
       tlEvaluated = true
       tl
     }
-    def force: Evaluated[A] = Some((head, tail))
-    override def toString: String =
+    override def force: Stream[A] = {
+      tail.force
+      this
+    }
+    override def toString(): String =
       s"$head #:: ${if (tlEvaluated) tail.toString else "?"}"
+
   }
 
   /** An alternative way of building and matching Streams using Stream.cons(hd, tl).
@@ -630,7 +635,8 @@ object Stream extends LazyListFactory[Stream] {
   }
 
   object #:: {
-    def unapply[A](s: Stream[A]): Evaluated[A] = s.force
+    def unapply[A](s: Stream[A]): Option[(A, Stream[A])] =
+      if (s.nonEmpty) Some(s.head, s.tail) else None
   }
 
   def from[A](coll: collection.IterableOnce[A]): Stream[A] = coll match {
--- a/test/junit/src/test/scala/strawman/collection/immutable/LazyListTest.scala
+++ b/test/junit/src/test/scala/strawman/collection/immutable/LazyListTest.scala
@@ -1,5 +1,6 @@
 package strawman.collection.immutable
 
+import strawman.collection.Iterator
 import org.junit.runner.RunWith
 import org.junit.runners.JUnit4
 import org.junit.{Test, Ignore}
@@ -139,4 +140,28 @@ class LazyListTest {
     assertEquals(LazyList(3), LazyList(1, 2, 3).drop(2))
     assertEquals(LazyList(3, 4), LazyList(1, 2, 3, 4).drop(2))
   }
+
+  @Test
+  def testForceReturnsEvaluatedLazyList() : Unit = {
+    var i = 0
+    def f: Int = { i += 1; i }
+    val xs = LazyList.from(Iterator.fill(3)(f))
+    assertEquals(0, i)
+    xs.force
+    assertEquals(3, i)
+    // it's possible to implement `force` with incorrect string representation
+    // (to forget about `tlEvaluated` update)
+    assertEquals( "1 #:: 2 #:: 3 #:: Empty", xs.toString())
+  }
+
+  @Test
+  def testSameElements(): Unit = {
+    assert(LazyList().sameElements(LazyList()))
+    assert(!LazyList().sameElements(LazyList(1)))
+    assert(LazyList(1,2).sameElements(LazyList(1,2)))
+    assert(!LazyList(1,2).sameElements(LazyList(1)))
+    assert(!LazyList(1).sameElements(LazyList(1,2)))
+    assert(!LazyList(1).sameElements(LazyList(2)))
+  }
+
 }
--- a/test/junit/src/test/scala/strawman/collection/immutable/StreamTest.scala
+++ b/test/junit/src/test/scala/strawman/collection/immutable/StreamTest.scala
@@ -125,10 +125,25 @@ class StreamTest {
   }
 
   @Test
-  def testStreamForcesHead: Unit = {
+  def testForceReturnsEvaluatedStream() : Unit = {
     var i = 0
     def f: Int = { i += 1; i }
-    val s = f #:: f #:: f #:: Stream.empty
+    val xs = f #:: f #:: f #:: Stream.empty
     assertEquals(1, i)
+    xs.force
+    assertEquals(3, i)
+    // it's possible to implement `force` with incorrect string representation
+    // (to forget about `tlEvaluated` update)
+    assertEquals( "1 #:: 2 #:: 3 #:: Empty", xs.toString())
+  }
+
+  @Test
+  def testSameElements(): Unit = {
+    assert(Stream().sameElements(Stream()))
+    assert(!Stream().sameElements(Stream(1)))
+    assert(Stream(1,2).sameElements(Stream(1,2)))
+    assert(!Stream(1,2).sameElements(Stream(1)))
+    assert(!Stream(1).sameElements(Stream(1,2)))
+    assert(!Stream(1).sameElements(Stream(2)))
   }
 }


collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -453,14 +453,12 @@ object LazyList extends LazyListFactory[LazyList] {
...
 
-  type Evaluated[+A] = Option[(A, LazyList[A])]
-
   object Empty extends LazyList[Nothing] {
     override def isEmpty: Boolean = true
     override def head: Nothing = throw new NoSuchElementException("head of empty lazy list")
     override def tail: LazyList[Nothing] = throw new UnsupportedOperationException("tail of empty lazy list")
-    def force: Evaluated[Nothing] = None
-    override def toString: String = "Empty"
+    override def force: LazyList[Nothing] = this
```
julienrf: I prefer you to omit `override` when you only *implement*

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -506,7 +508,8 @@ object LazyList extends LazyListFactory[LazyList] {
   }
 
   object #:: {
-    def unapply[A](s: LazyList[A]): Evaluated[A] = s.force
+    def unapply[A](s: LazyList[A]): Option[(A, LazyList[A])] =
+      if (s.nonEmpty) Some(s.head, s.tail) else None
```
julienrf: You should add extra parens to make it clear that `Some` takes one parameter (which is a pair)

julienrf:
Thanks @optician! That looks good, I left two comments about code style. We should probably wait for #446 to be merged and then you should rebase your work on top of it.

julienrf:
Well, maybe this PR will go first because the other one isn’t ready.

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
optician: Fixed.

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
optician: Fixed.

optician:
What bother me is `LazyListOps.sameElements` method. In fact it is the same as `LinearSeqOps.sameElements` (without a modification of reference equality). At the same time `LazyListOps` is the only successor of `LinearSeqOps`. 

Would it be reasonable to remove override? At first glance it's ok. LazyList doesn't evaluate totally if some elements don't equal.

julienrf:
Good point, I didn’t notice that. I think it would make sense to pull up the implementation of `sameElements` of `LazyList` and `Stream` to `LinearSeqOps` so that we don’t even have to override it. What do you think?

optician:
Agree. I'll do it.

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -235,8 +235,8 @@ sealed private[immutable] trait LazyListOps[+A, +CC[+X] <: LinearSeq[X] with Laz
 
   protected[this] def cons[T](hd: => T, tl: => CC[T]): CC[T]
 
-  /** Force the evaluation of both the head and the tail of this `LazyList` */
-  def force: Option[(A, CC[A])]
+  /** Force the evaluation of the whole `LazyList` */
+  def force: CC[A]
```
lrytz: should the return type be `this.type`?

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -475,8 +473,12 @@ object LazyList extends LazyListFactory[LazyList] {
       tlEvaluated = true
       tl
     }
-    def force: Evaluated[A] = Some((head, tail))
-    override def toString: String =
+    def force: LazyList[A] = {
+      head
+      tail.force
```
lrytz: 2.12 supports forcing cyclic streams:

```
scala> val s: Stream[Int] = List(1,2,3).toStream #::: s
s: Stream[Int] = Stream(1, ?)

scala> s.force
res6: scala.collection.immutable.Stream[Int] = Stream(1, 2, 3, ...)
```

We should aim for an iterative implementation instead of a (non-tail-) recursive one.

https://github.com/scala/scala/blob/v2.13.0-M3/src/library/scala/collection/immutable/Stream.scala#L254-L276

lrytz: Context: https://github.com/scala/scala/pull/3848. Don't worry about replicating the test cases of that PR here in the strawman repo, the test cases in scala/scala will cover it.

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
optician: Fixed

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
optician: Fixed.

optician:
Wrong message in last commit, sorry. 

Method `LazyList.from(coll: IterableOnce)` is called everywhere with `coll.iterator()` argument. `iterator()` is redundant with regard to `from` implementation. Fixed.

Other notes. 
* Three 'cons' methods attract attention: `LazyListOps.cons`, `LazyList.newCons`, `LazyList.cons.apply`. They do the same but I couldn't find a way to reduce their number.
* `IterableOnceExtensionMethods` and `IterableOps` duplicates some methods (for example `toStream`, `toList` and so on). The first is AnyVal wrapper over `IterableOnce` and the second is extension of `IterableOnce`. May be it's possible to remove one of them.

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -475,8 +465,26 @@ object LazyList extends LazyListFactory[LazyList] {
       tlEvaluated = true
       tl
     }
-    def force: Evaluated[A] = Some((head, tail))
-    override def toString: String =
+    def force: this.type = {
+      // Use standard 2x 1x iterator trick for cycle detection ("those" is slow one)
+      var these, those: LazyList[A] = this
+
+      def hareStep(): Unit = {
```
lrytz: DRY, I know, but this method causes boxing because it assigns the local variable `these`. So let's just inline it (manually)

```
$ scala -Xprint:cleanup
scala> def f = { var x = "hi"; def step() = { x += "-" }; step(); step(); x }
    def f(): String = {
      var x: runtime.ObjectRef = scala.runtime.ObjectRef.create("hi");
      iw.this.step$1(x);
      iw.this.step$1(x);
      x.elem.$asInstanceOf[String]()
    };
    final private[this] def step$1(x$1: runtime.ObjectRef): Unit = x$1.elem = x$1.elem.$asInstanceOf[String]().+("-");
```

See how `x` is boxed into an `ObjectRef`.

julienrf:
> IterableOnceExtensionMethods and IterableOps duplicates some methods (for example toStream, toList and so on). The first is AnyVal wrapper over IterableOnce and the second is extension of IterableOnce. May be it's possible to remove one of them.

The methods in `IterableOnceExtensionMethods` are deprecated anyway.

collections/src/main/scala/strawman/collection/LinearSeq.scala
```diff
@@ -127,20 +127,21 @@ trait LinearSeqOps[+A, +CC[X] <: LinearSeq[X], +C <: LinearSeq[A]] extends Any w
...
-    case that1: LinearSeq[B] =>
-      // Probably immutable, so check reference identity first (it's quick anyway)
-      (coll eq that1) || {
-        var these: LinearSeq[A] = coll
-        var those: LinearSeq[B] = that1
-        while (!these.isEmpty && !those.isEmpty && these.head == those.head) {
-          these = these.tail
-          those = those.tail
+  override def sameElements[B >: A](that: IterableOnce[B]): Boolean = {
+    @tailrec def lazyListEq(a: LinearSeq[B], b: LinearSeq[B]): Boolean =
```
julienrf: can you rename it to `linearSeqEq` for the sake of consistency?

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -246,28 +254,13 @@ sealed private[immutable] trait LazyListOps[+A, +CC[+X] <: LinearSeq[X] with Laz
     * @return The lazy list containing elements of this lazy list and the iterable object.
     */
   def lazyAppendAll[B >: A](suffix: => collection.IterableOnce[B]): CC[B] =
-    if (isEmpty) iterableFactory.from(suffix.iterator()) else cons[B](head, tail.lazyAppendAll(suffix))
+    if (isEmpty) iterableFactory.from(suffix) else cons[B](head, tail.lazyAppendAll(suffix))
```
julienrf: Actually if `suffix` is already a `LazyList`, then `iterableFactory.from(suffix)` is a no-op, so I would keep using just `suffix`.

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -344,11 +337,11 @@ sealed private[immutable] trait LazyListOps[+A, +CC[+X] <: LinearSeq[X] with Laz
   else {
     // establish !prefix.isEmpty || nonEmptyPrefix.isEmpty
     var nonEmptyPrefix: CC[A] = coll
-    var prefix = iterableFactory.from(f(nonEmptyPrefix.head).iterator())
+    var prefix = iterableFactory.from(f(nonEmptyPrefix.head))
     while (!nonEmptyPrefix.isEmpty && prefix.isEmpty) {
       nonEmptyPrefix = nonEmptyPrefix.tail
       if(!nonEmptyPrefix.isEmpty)
-        prefix = iterableFactory.from(f(nonEmptyPrefix.head).iterator())
```
julienrf: Same comment here about `iterableFactory.from`

test/junit/src/test/scala/strawman/collection/immutable/LazyListTest.scala
```diff
@@ -139,4 +140,28 @@ class LazyListTest {
...
+  @Test
+  def testSameElements(): Unit = {
+    assert(LazyList().sameElements(LazyList()))
+    assert(!LazyList().sameElements(LazyList(1)))
+    assert(LazyList(1,2).sameElements(LazyList(1,2)))
+    assert(!LazyList(1,2).sameElements(LazyList(1)))
+    assert(!LazyList(1).sameElements(LazyList(1,2)))
+    assert(!LazyList(1).sameElements(LazyList(2)))
+  }
+
```
julienrf: Can you also add a test with a cyclic collection?

collections/src/main/scala/strawman/collection/LinearSeq.scala
optician: Fixed.

test/junit/src/test/scala/strawman/collection/immutable/LazyListTest.scala
optician: I added 2 tests. 
At the same time comparison of equal collections but with different references is infinite. I didn't catch OOM. Strange feeling but it seems ok - collection doesn't have unnecessary overhead and indeed infinite. But should I add test for absent of OOM? And if so how is it possible to implement?
```
val cycle1: LazyList[Int] = 1 #:: 2 #:: cycle1
val cycle2: LazyList[Int] = 1 #:: 2 #:: 3 #:: cycle2
val cycle3: LazyList[Int] = cycle2 #::: cycle3
val cycle4: LazyList[Int] = cycle2 #::: cycle4
assert(!cycle3.sameElements(cycle4))
```

One more thing (applies for old collections too). `val cycle: LazyList[Int] = LazyList.Empty #::: cycle` evaluates into 
```
java.lang.NullPointerException
  at strawman.collection.immutable.LazyList$.from(LazyList.scala:523)
  at strawman.collection.immutable.LazyList$.from(LazyList.scala:444)
  at strawman.collection.immutable.LazyListOps.lazyAppendAll(LazyList.scala:257)
  at strawman.collection.immutable.LazyListOps.lazyAppendAll$(LazyList.scala:256)
  at strawman.collection.immutable.LazyList.lazyAppendAll(LazyList.scala:186)
  at strawman.collection.immutable.LazyList$Deferrer.$hash$colon$colon$colon(LazyList.scala:513)
  ... 40 elided
```
Didn't have enough time to investigate. Is it ok (look back to the old collection) or should be fixed?

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
optician: Fixed. 
@lrytz please, can you give me a direction to learn more about purpose of `ObjectRef` and why wrapping in the case required?

optician:
> The methods in IterableOnceExtensionMethods are deprecated anyway.

@julienrf the same is true for methods in `IterableOps`. For example
```
@deprecated("Use LazyList.from(it) instead of it.toStream", "2.13.0")
@`inline` final def toStream: immutable.LazyList[A] = immutable.LazyList.from(this)
```

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
optician: It seems it fixes #421.

julienrf: Interesting, do you have any insight of why? Looking at the code this is not clear to me.

optician: With `suffix.iterator()` stacktrace is (differs from original issue)
```
java.lang.StackOverflowError
  at strawman.collection.immutable.LazyList$Cons.head$lzycompute(LazyList.scala:462)
  at strawman.collection.immutable.LazyList$Cons.head(LazyList.scala:460)
  at strawman.collection.LinearSeqOps$$anon$1.next(LinearSeq.scala:25)
  at strawman.collection.Iterator$$anon$14.next(Iterator.scala:891)
  at strawman.collection.immutable.LazyList$.evaluatedElem$lzycompute$1(LazyList.scala:538)
  at strawman.collection.immutable.LazyList$.evaluatedElem$1(LazyList.scala:538)
  at strawman.collection.immutable.LazyList$.$anonfun$fromIterator$1(LazyList.scala:539)
  at strawman.collection.immutable.LazyList$Cons.head$lzycompute(LazyList.scala:462)
  at strawman.collection.immutable.LazyList$Cons.head(LazyList.scala:460)
  at strawman.collection.immutable.LazyListOps.flatMap(LazyList.scala:340)
  at strawman.collection.immutable.LazyListOps.flatMap$(LazyList.scala:335)
  at strawman.collection.immutable.LazyList.flatMap(LazyList.scala:186)
  at strawman.collection.immutable.LazyListOps.$anonfun$flatMap$1(LazyList.scala:348)
  at strawman.collection.immutable.LazyListOps.lazyAppendAll(LazyList.scala:257)
  at strawman.collection.immutable.LazyListOps.lazyAppendAll$(LazyList.scala:256)
  at strawman.collection.immutable.LazyList.lazyAppendAll(LazyList.scala:186)
  at strawman.collection.immutable.LazyListOps.$anonfun$lazyAppendAll$2(LazyList.scala:257)
  at strawman.collection.immutable.LazyList$Cons.tail$lzycompute(LazyList.scala:466)
  at strawman.collection.immutable.LazyList$Cons.tail(LazyList.scala:464)
  at strawman.collection.immutable.LazyList$Cons.tail(LazyList.scala:456)
  at strawman.collection.LinearSeqOps$$anon$1.next(LinearSeq.scala:25)
  at strawman.collection.immutable.LazyList$.evaluatedElem$lzycompute$1(LazyList.scala:538)
  at strawman.collection.immutable.LazyList$.evaluatedElem$1(LazyList.scala:538)
  at strawman.collection.immutable.LazyList$.$anonfun$fromIterator$1(LazyList.scala:539)
  at strawman.collection.immutable.LazyList$Cons.head$lzycompute(LazyList.scala:462)
  at strawman.collection.immutable.LazyList$Cons.head(LazyList.scala:460)
  at strawman.collection.LinearSeqOps$$anon$1.next(LinearSeq.scala:25)
  at strawman.collection.immutable.LazyList$.evaluatedElem$lzycompute$1(LazyList.scala:538)
  at strawman.collection.immutable.LazyList$.evaluatedElem$1(LazyList.scala:538)
  at strawman.collection.immutable.LazyList$.$anonfun$fromIterator$1(LazyList.scala:539)
  at strawman.collection.immutable.LazyList$Cons.head$lzycompute(LazyList.scala:462)
  at strawman.collection.immutable.LazyList$Cons.head(LazyList.scala:460)
```
If I understood it correctly then methods are called in such order (unused lines are hid under comments):
```scala
def flatMap[B](f: A => IterableOnce[B]): CC[B] =
  // if (isEmpty) iterableFactory.empty
  // else 
  {
    var nonEmptyPrefix: CC[A] = coll
    var prefix = iterableFactory.from(f(nonEmptyPrefix.head))
    while (!nonEmptyPrefix.isEmpty && prefix.isEmpty) {
      // nonEmptyPrefix = nonEmptyPrefix.tail
      // if(!nonEmptyPrefix.isEmpty)
      //   prefix = iterableFactory.from(f(nonEmptyPrefix.head))
    }

    // if (nonEmptyPrefix.isEmpty) iterableFactory.empty
    // else 
    prefix.lazyAppendAll(nonEmptyPrefix.tail.flatMap(f))
  }

def lazyAppendAll[B >: A](suffix: => collection.IterableOnce[B]): CC[B] =
    if (isEmpty) iterableFactory.from(suffix.iterator()) else cons[B](head, tail.lazyAppendAll(suffix))

def fromIterator[A](it: Iterator[A]): LazyList[A] =
  // if (it.hasNext) {
    lazy val evaluatedElem = it.next()
    new LazyList.Cons(evaluatedElem, { evaluatedElem; fromIterator(it) })
  // } else LazyList.Empty
```
`fromIterator` is recursive (without 'tail-') and it's a cause of SOE.

julienrf: I see, then we should fix these operations to be stack-safe. But I think that should be done in a separate PR. In the meantime, I continue thinking that it’s better to use `suffix` rather than `suffix.iterator()`.

optician: I think `.iterator()`  is redundant in any case. That's the reason I removed it.

Should I open issue for `fromIterator`?

julienrf: > I think .iterator() is redundant in any case. That's the reason I removed it.

OK, actually I don’t know why I read the diff in the opposite side: I thought you had added the `.iterator` calls… But I was wrong, your PR is correct :)

> Should I open issue for fromIterator?

Sure!

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
lrytz: In the JVM there are no nested methods. So if a nested method in Scala accesses a local `val` of the enclosing method, that val gets passed as argument to the lifted nested method:

```
$ scala -Xprint:cleanup
scala> def f = { val x = 1; def g = x; g }
    def f(): Int = {
      val x: Int = 1;
      iw.this.g$1(x)
    };
    final private[this] def g$1(x$1: Int): Int = x$1;
```

Now if a nested method *modifies* a local variable of the enclosing method, passing the variable as argument is not enough, as the lifted method could not modify the local variable of the other method. So the compiler wraps the local variable in a mutable box.

```
scala> def f = { var x = 1; def g = {x = 2}; g; x }
    def f(): Int = {
      var x: runtime.IntRef = scala.runtime.IntRef.create(1);
      iw.this.g$1(x);
      x.elem
    };
    final private[this] def g$1(x$1: runtime.IntRef): Unit = x$1.elem = 2;
```

optician: Thx a lot! :)

test/junit/src/test/scala/strawman/collection/immutable/LazyListTest.scala
lrytz: let's create a ticket to keep track of it

dhanesharole:
@julienrf and @optician. #446 is ready. I have made requested changes. Can u fools have a look. Since, it's a small one and is kind of blocker for this one.

dhanesharole:
*folks..

dhanesharole:
Extension sorry. Stupid autocorrect

Ichoran:
@dhanesharole - You can edit previous comments (click on the three dots at the top right of the message) if they don't turn out the way you intended or you notice a mistake.  (Won't change the email notifications, but still.)

dhanesharole:
@optician @julienrf @Ichoran , extremely sorry. I genuinely din't mean that, I was writing from mobile in a metro.

optician:
@dhanesharole no problem, funny typo)

PR #456, opened 2018-02-17
[WIP] Patch missing method (#397)

### LINKED ISSUE #397: Mismatching signature of copyToArray

jiminhsieh:
Fix #397.

I found the `copyToArray` at 2.12.x has the method loading. 
1. `def copyToArray[B >: A](xs: Array[B], start: Int, len: Int): Unit`
2. `def copyToArray[B >: A](xs: Array[B], start: Int): Unit`
3. `def copyToArray[B >: A](xs: Array[B]): Unit`

Since the original method can achieve 2. and 3., I patched this for 1.


jiminhsieh:
I am thinking of patching test cases to `junit`'s `IterableTest`.

jiminhsieh:
#397 had fixed by #452.

Issue #408, opened 2018-02-04
Mutable maps and sets don't mapInPlace properly


Ichoran:
They end up behaving like `xs ++= xs.map(f)` instead of `xs = xs.map(f)`.

Example code:

```
val xs = mutable.HashSet(1)
xs.mapInPlace(_ + 1)
xs
```

xs is now `Set(1, 2)`

optician:
It seems issue already have been fixed. Scala project, branch 2.13.x-new-collections.
```
[2.13.x-new-collections] root: console
[info] Starting scala interpreter...
[info]
Welcome to Scala 2.13.0-M4-pre-20d3c21 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_121).
Type in expressions for evaluation. Or try :help.

scala> :paste
// Entering paste mode (ctrl-D to finish)

import scala.collection._
val xs = mutable.HashSet(1)
xs.mapInPlace(_ + 1)
xs

// Exiting paste mode, now interpreting.

import scala.collection._
xs: scala.collection.mutable.HashSet[Int] = HashSet(2)
```

optician:
Tests exist already.
https://github.com/scala/scala/blob/2.13.x/test/junit/scala/collection/mutable/HashSetTest.scala#L39
https://github.com/scala/scala/blob/2.13.x/test/junit/scala/collection/mutable/HashMapTest.scala#L52

julienrf:
Thanks @optician to have double checked that! I’m closing the issue.

PR #459, opened 2018-02-18
Fixed mapInPlace on mutable maps and sets

ruhly:
addresses #408


ruhly:
Once the mapInPlace issue is resolved, it probably needs to be propagated to flatMapInPlace because it looks like those methods suffer from the same issue

collections/src/main/scala/strawman/collection/mutable/Map.scala
```diff
@@ -121,15 +121,10 @@ trait MapOps[K, V, +CC[X, Y] <: MapOps[X, Y, CC, _], +C <: MapOps[K, V, CC, C]]
...
-      val mapped = f(elem)
-      if (!contains(mapped._1)) {
-        toAdd += mapped
-        toRemove -= elem._1
-      }
+        toAdd += f(elem)
     }
-    for (elem <- toRemove) coll -= elem
+    coll.clear()
     for (elem <- toAdd) coll += elem
```
julienrf: You can even call `coll ++= toAdd`, which should be faster.

szeiger:
I didn't even realize we had these methods. If any such method makes sense for `Map` then IMHO it is `mapValuesInPlace`.

ruhly:
also see issue #473.

julienrf:
Thanks

PR #463, opened 2018-02-19
Fix slice behavior difference of ImmutableArray

ackratos:
Fix ImmutableArray slice behavior difference with 2.13 and add test case.

Following code won't throw exception anymore

```
package strawman.collection.immutable

object Test extends App {
  val array = ImmutableArray(1, 2, 3)
  val newArray = array.slice(1, 5)
  println(newArray)
}
```


ackratos:
@julienrf @szeiger thanks for merging!

Issue #335, opened 2018-01-09
Provide default implementations of factory methods when possible


julienrf:
The `fromSpecificIterable`, `newSpecificBuilder`, and other factory methods that are currently implemented in concrete collections only could probably have default implementations earlier in the hierarchy.

julienrf:
In #379, @lrytz wrote the following:

In 2.12:

```
scala>   class K extends Iterable[Int] {
     |     var x = 10
     |     def iterator: Iterator[Int] = new Iterator[Int] {
     |       def hasNext = x > 0
     |       def next() = { x -= 1; x}
     |     }
     |   }
defined class K

scala>

scala> (new K).toList
res0: List[Int] = List(9, 8, 7, 6, 5, 4, 3, 2, 1, 0)
```

In strawman I need to add

```
    def iterableFactory: collection.IterableFactoryLike[Iterable] = Iterable
    protected[this] def fromSpecificIterable(coll: scala.Iterable[Int]): Iterable[Int] = iterableFactory.from(coll)
    protected[this] def newSpecificBuilder(): collection.mutable.Builder[Int, Iterable[Int]] = iterableFactory.newBuilder()
```

It's pretty mechanical (and also not obvious to newcomers) in the basic case when you don't need specific return types for operations. Should we provide some helper classes with default implementations?

szeiger:
The approach I had in mind runs into a bug (or a more fundamental language problem?) in Scala (but would work in Dotty): https://github.com/scala/bug/issues/10725

PR #468, opened 2018-02-20
Remove the boilerplate associated with the new design (and more)

szeiger:
- The respective `CC` types are exposed (to `protected[this]`) through
  aliases `IterableCC`, `SortedIterableCC`, `MapCC` and `SortedMapCC`
  in the respective `Ops` traits, plus a `BitSetC` for the shared
  `BitSetOps` trait.

- This allows defining `fromSpecificIterable`, `newSpecificBuilder`
  and `empty` (for all but `Iterable`) in the respective collection
  traits `Iterable`, `SortedSet`, `Map` and `SortedMap` (*not* in their
  `Ops` traits) where the `CC` type is visible and set to a concrete
  type but can still be refined in subtypes. This gives us a valid
  implementation until a concrete collection type with a refined `CC`
  gets defined. In this case the inherited implementations have the
  wrong type, so the user is forced to override them.

- Implementations of `fromSpecificIterable`, `newSpecificBuilder` and
  `empty` can be removed from almost all collection implementations
  except the ones where they need to be refined even further than what
  a factory can provide (e.g. `IntMap`, `PriorityQueue`).

- Factories are generally overridden in abstract collection types that
  refine a `CC` (e.g. `Iterable` -> `Seq` -> `immutable.Set` ->
  `immutable.IndexedSeq`) so that concrete implementations without
  further refinement do not need to override them.

- DefaultMap is deprecated because there is no more boilerplate left
  that it could implement.

- `sortedFromIterable`, `mapFromIterable` and `sortedMapFromIterable`
  are treated the same way as `fromIterable`. They are implemented as
  `inline final` calls to the respective factory method.

Fixes https://github.com/scala/collection-strawman/issues/335

### DIFF
--- a/collections/src/main/scala/strawman/collection/SortedSet.scala
+++ b/collections/src/main/scala/strawman/collection/SortedSet.scala
@@ -1,20 +1,32 @@
 package strawman.collection
 
-import scala.{Boolean, Ordering, `inline`, None, Option, Some}
+import scala.{Boolean, Ordering, `inline`, None, Option, Some, Any}
 import scala.annotation.unchecked.uncheckedVariance
 
 /** Base type of sorted sets */
 trait SortedSet[A] extends Set[A] with SortedSetOps[A, SortedSet, SortedSet[A]] {
   def unsorted: Set[A] = this
+
+  override protected[this] def fromSpecificIterable(coll: Iterable[A]): SortedIterableCC[A] = sortedIterableFactory.from(coll)
+  override protected[this] def newSpecificBuilder(): mutable.Builder[A, SortedIterableCC[A]] = sortedIterableFactory.newBuilder[A]()
+
+  def sortedIterableFactory: SortedIterableFactory[SortedIterableCC] = SortedSet
+
+  override def empty: SortedIterableCC[A] = sortedIterableFactory.empty
 }
 
 trait SortedSetOps[A, +CC[X] <: SortedSet[X], +C <: SortedSetOps[A, CC, C]]
   extends SetOps[A, Set, C]
      with SortedOps[A, C] {
 
-  def sortedIterableFactory: SortedIterableFactory[CC]
+  protected[this] type SortedIterableCC[X] = CC[X]
 
-  protected[this] def sortedFromIterable[B: Ordering](it: Iterable[B]): CC[B]
+  def sortedIterableFactory: SortedIterableFactory[SortedIterableCC]
+
+  /** Similar to `fromSpecificIterable`, but for a (possibly) different type of element.
+    * Note that the return type is now `CC[B]` aka `SortedIterableCC[B]` rather than `IterableCC[B]`.
+    */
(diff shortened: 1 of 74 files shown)


collections/src/main/scala/strawman/collection/SortedSet.scala
```diff
@@ -1,20 +1,32 @@
...
      with SortedOps[A, C] {
 
-  def sortedIterableFactory: SortedIterableFactory[CC]
+  protected[this] type SortedIterableCC[X] = CC[X]
 
-  protected[this] def sortedFromIterable[B: Ordering](it: Iterable[B]): CC[B]
+  def sortedIterableFactory: SortedIterableFactory[SortedIterableCC]
+
+  /** Similar to `fromSpecificIterable`, but for a (possibly) different type of element.
+    * Note that the return type is know `CC[E]`.
```
julienrf: typo: know => now. Also CC[E] => CC[B]

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -28,17 +28,17 @@ sealed abstract class ImmutableArray[+A]
...
+  override def iterableFactory: SeqFactory[ImmutableArray] = ImmutableArray.untagged
 
   /** The wrapped mutable `Array` that backs this `ImmutableArray`. Any changes to this array will break
     * the expected immutability. */
   def unsafeArray: Array[A @uncheckedVariance]
   // uncheckedVariance should be safe: Array[A] for reference types A is covariant at the JVM level. Array[A] for
   // primitive types A can only be widened to Array[Any] which erases to Object.
 
-  protected[this] def fromSpecificIterable(coll: strawman.collection.Iterable[A]): ImmutableArray[A] = fromIterable(coll)
+  override protected[this] def fromSpecificIterable(coll: strawman.collection.Iterable[A]): ImmutableArray[A] = fromIterable(coll)
```
julienrf: Here what’s the difference with the base implementation?

julienrf:
You will have to update collections-contrib as well.

lrytz:
This is really great!!

collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
szeiger: This is wrong and was equally wrong before. We need to build using `elemTag` here.

collections/src/main/scala/strawman/collection/SortedSet.scala
szeiger: I'll update these comments to make more sense for the new design.

szeiger:
The new type members are unavoidable. They are merely aliases for existing type parameters, not abstract types. They ensure that the collection traits can declare the correct return types depending on some `CC` even though `CC` is fixed at that point (but can be covariantly refined). This allows us to provide concrete implementations even though the types can be changed in incompatible ways in subclasses.

This started as a plan B after running into https://github.com/scala/bug/issues/10725 but even without that problem my original approach would not have worked because concrete definitions win over abstract ones. It might still work in Dotty if it has gotten rid of that rule in favor of proper type unification.

PR #470, opened 2018-02-20
Fix the iterableFactory of mutable Sets

szeiger:
This points to a soundness bug in the compiler. These errors should be
caught. Omitting a necessary override leads to the expected errors in
some cases but not in all.

### DIFF
--- a/collections/src/main/scala/strawman/collection/Set.scala
+++ b/collections/src/main/scala/strawman/collection/Set.scala
@@ -28,6 +28,8 @@ trait Set[A]
 
   override def hashCode(): Int = Set.setHash(toIterable)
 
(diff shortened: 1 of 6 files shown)


julienrf:
> Omitting a necessary override leads to the expected errors in
> some cases but not in all.

I hope we can fix that in the compiler :(

szeiger:
Looks like a similar issue to https://github.com/scala/bug/issues/7093, present in Scala and Dotty, with no solution since it was opened 5 years ago, so I wouldn't hold my breath.

szeiger:
I could have held my breath after all because Dotty fixed it: https://github.com/lampepfl/dotty/pull/4013. See my comment there. Not only are we exploiting unsound `protected[this]` in my design changes, we already did before, just as the old collections library does.

PR #478, opened 2018-02-21
Introduce typeclasses making it easier to write generic extension methods

### LINKED ISSUE #294: Improve the way we can generically implement decorators on collections
By generically I mean “that works on any collection type and preserves that collection type”. In `collections-contrib` there are a couple of decorators defined by their type signatures are a bit complex and don’t always follow the same pattern. Also, they might not work with `View`s, `String` or `Array[_]`.

julienrf:
Fixes #294.


collections-contrib/src/test/scala/strawman/collection/decorators/SeqDecoratorTest.scala
```diff
@@ -17,4 +18,20 @@ class SeqDecoratorTest {
...
+    val array = Array(1, 2, 3)
+    val string = "foo"
+    list.intersperse(0)
+    list.view.intersperse(0)
+    vector.intersperse(0)
+    vector.view.intersperse(0)
+    array.intersperse(0)
+    array.view.intersperse(0)
+    string.intersperse(' ')
+    string.view.intersperse(' ')
```
julienrf: This approach works with `List`, `Vector` but also views, `Array` and `String`.

However, there seems to be some type information lost somewhere:

~~~ scala
    val result = list.intersperse(0)
    result: List[Int]
~~~

julienrf:
Here is a design that make it possible to define a decorator that will work for any collection-like type, whether it is a collection (e.g. `Seq[Int]`), a view (e.g. `SeqView[Int]`) or something that can be seen as a collection (e.g. `String`). It is possible to program generically against this decorated type and extract type information such as the type of elements or the return type of transformation operations.

1. A decorator is an implicit class that takes an implicit `HasXxxOps`, where `Xxx` can be `Set`, `Seq`, etc.:

~~~ scala
implicit class SeqDecorator[C, S <: HasSeqOps[C]](coll: C)(implicit val seq: S) { … }
~~~

(note that if https://github.com/lampepfl/dotty/issues/3964 or https://github.com/scala/bug/issues/5712 were solved we could just write the following: `implicit class SeqDecorator[C](coll: C)(implicit val seq: HasSeqOps[C])`)

The definition of `HasSeqOps` is the following:

~~~ scala
trait HasSeqOps[Repr] {
  type A       // element type
  type CC[_]   // collection’s type constructor
  type C       // collection type
  val conversion: Repr => SeqOps[A, CC, C]
}
~~~

The type `HasSeqOps` contains type members referring to the type of elements, the collection type constructor and the collection type of the decorated collection. For instance, if we decorate a `List[Int]` we would get `seq.A =:= Int`, `seq.CC =:= List` and `seq.C =:= List[Int]`. Here is an example of use of the type of elements `A` to set a lower bound on type a parameter taken by the extension method:

~~~ scala
  def intersperse[B >: seq.A](sep: B) = …
~~~

The fact that these types (`A`, `CC` and `C`) are type members avoid users of the `HasSeqOps` trait to take type parameters. This saves a lot of boilerplate because not only there are three parameters, but also because they sometimes have upper bound constraints. By capturing them as type members, users can still use some of them if they want to, but don’t have to define a type parameter for all of them.

2. Extension methods can use `BuildFrom` to compute their result type. We might be tempted to use `seq.CC` or `seq.C` instead, but we would get an `IndexedSeq` instead of an `Array`, in case of usage with arrays.

~~~ scala
  def intersperse[B >: seq.A, That](sep: B)(implicit bf: BuildFrom[C, B, That]): That =
    bf.fromSpecificIterable(coll)(new View.Intersperse(seq.conversion(coll), sep))
~~~

The drawbacks of this design are:

- decorators take two parameters, and thus can not be made value classes,
- it is often hard to understand *why* an implicit `HasXxxOps[C]` instance has not be found for a given type `C`.

In my opinion this framework is a bit too complicated. What do you think?

collections-contrib/src/main/scala/strawman/collection/decorators/HasImmutableMapOps.scala
```diff
@@ -0,0 +1,43 @@
...
+/**
+  * Type class witnessing that a collection type `C` has
+  * a conversion to `immutable.MapOps[K, V, _, _]`.
+  *
+  * @see [[scala.collection.decorators.HasIterableOps]]
+  */
+trait HasImmutableMapOps[C] extends HasMapOps[C] {
+
+  // Convenient intermediate type definitions to satisfy type bounds
+  protected type _CC[X, +Y] <: immutable.MapOps[X, Y, _CC, _]
```
Ichoran: If the user ever needs to interact with this type directly, I would change the name to something more memorable, e.g. `MapOf`.

collections-contrib/src/main/scala/strawman/collection/decorators/HasImmutableMapOps.scala
```diff
@@ -0,0 +1,43 @@
...
+  protected type _C <: immutable.MapOps[K, V, _CC, _C]
```
Ichoran: Again, if the user needs to interact with this it should have a better name, e.g. `MapType`.

collections-contrib/src/main/scala/strawman/collection/decorators/HasImmutableMapOps.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  /** A conversion from the type `C` to `immutable.MapOps[K, V, _, _]` */
+  val conversion: C => immutable.MapOps[K, V, _CC, _C]
```
Ichoran: Why is this a val?  I would think this would be easiest to use as a def apply, so you can `(implicit ops: M)` and then in the code `ops(coll)`.

Ichoran:
The strategy looks sound overall, though I do worry that the complexity will seem daunting to people even if mostly it "just works".  We'll need some good documentation.

Also, I think we could improve the usability a bit (as indicated on my comments for `HasImmutableMapOps`, but they apply to everything).

collections-contrib/src/main/scala/strawman/collection/decorators/HasImmutableMapOps.scala
julienrf: This is an implementation detail which is not supposed to be used by users. I wish I could get rid of it but couldn’t figure out how.

collections-contrib/src/main/scala/strawman/collection/decorators/HasImmutableMapOps.scala
julienrf: I agree. This is fixed.

julienrf:
In any case, this framework mainly targets advanced users.

A first support of generic decorators is already provided by `BuildFrom` (exactly like we used to do with `CanBuildFrom` in the old collections). This allows users to generically define extension methods for any collection type but views, `String` and `Array`. This should be fine for most cases.

If one needs more power (ie the ability to also abstract over views, `String` and `Array`), then `HasIterableOps` and friends are the right tool to use.

Last, I’ve put this machinery in the `collections-contrib` module. We don’t have to have it in the core.

julienrf:
Rebased

julienrf:
@szeiger any objection to merge this one?

Issue #465, opened 2018-02-19
Enforce consistency between the available operations of IterableOnce and Iterable


julienrf:
Currently these collection types don’t have a common super type.

@lrytz suggested that we write some tests that use reflection to check that the operations and type signatures are consistent between them.

lrytz:
Here's a the current diff of Iterable vs Iterator operations: https://gist.github.com/lrytz/39bdb1b25c23e204c3263809534ba6a0/revisions#diff-5b3b7587cc6e69d92ccb282bab852b31

lrytz:
See also https://github.com/scala/collection-strawman/issues/419

szeiger:
This sounds a lot like the `IterableMonoTransforms` we had in earlier iterations of the design.

szeiger:
I had a look at the diff and most of what I saw looks reasonable. `Iterator` does have methods like `dropRight` that would require forcing and caching everything. OTOH, because an `Iterator` represents a single traversal, it has some `Seq`-like features that are missing from `Iterable`.

PR #479, opened 2018-02-21
Don't extend collection types in ArrayOps

szeiger:
+ more ArrayOps methods and optimizations

See commit comments for details.

### DIFF
--- a/collections/src/main/scala/strawman/collection/ArrayOps.scala
+++ b/collections/src/main/scala/strawman/collection/ArrayOps.scala
@@ -1,11 +1,12 @@
 package strawman
 package collection
 
-import scala.{AnyVal, Array, Char, Int, throws, Boolean, Serializable, Unit, `inline`, Option, Some, None, PartialFunction}
+import scala.{AnyVal, Array, Char, Int, throws, Boolean, Serializable, Unit, `inline`, Option, Some, None, PartialFunction, ArrayIndexOutOfBoundsException, NoSuchElementException, AnyRef}
+import scala.Predef.{implicitly, classOf}
 import java.lang.String
 import mutable.ArrayBuilder
 import scala.reflect.ClassTag
-import scala.math.{max, min}
+import scala.math.{max, min, Ordering}
 
 object ArrayOps {
   /** A lazy filtered array. No filtering is applied until one of `foreach`, `map` or `flatMap` is called. */
@@ -69,6 +70,28 @@ object ArrayOps {
     /** Creates a new non-strict filter which combines this filter with the given predicate. */
     def withFilter(q: A => Boolean): WithFilter[A] = new WithFilter[A](a => p(a) && q(a), xs)
   }
+
+  private class ArrayIterator[A](private[this] val xs: Array[A]) extends Iterator[A] {
+    private[this] var pos = 0
+    def hasNext: Boolean = pos < xs.length
+    def next(): A = try {
+      val r = xs(pos)
+      pos += 1
+      r
+    } catch { case _: ArrayIndexOutOfBoundsException => throw new NoSuchElementException }
+  }
+
+  private class GroupedIterator[A](xs: Array[A]) extends Iterator[Array[A]] {
+    private[this] var pos = 0
+    def hasNext: Boolean = pos < xs.length
+    def next(): Array[A] = {
+      if(pos >= xs.length) throw new NoSuchElementException
+      val r = xs.slice(pos, pos+size)
+      pos += size
+      r
+    }
+  }
+
 }
 
 /** This class serves as a wrapper for `Array`s with many of the operations found in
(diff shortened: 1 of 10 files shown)


lrytz:
Nice! What are your thoughts for keeping the "interface" (available methods) in synch? https://github.com/scala/collection-strawman/issues/465

julienrf:
This PR makes type inference fail on arrays usage. Also, what do you think of implementing the optimized versions of reduction operations on `WrappedArray` instead (which inherits from `IterableOps`)?

szeiger:
There's one type inference failure in the tests. I have not been able to figure out why this change makes it fail. The example we have is something like `Array("") ++ Array("", null)`. ArrayOps should have exactly the same methods as before (i.e. a single `++` method that takes a `ClassTag`; the old implementation had another without the `ClassTag` which was defined in a superclass and therefore shadowed) but the failure is new. Note that the general problem already existed before this change, we just didn't have any test for it: `Iterable("") ++ Array("", null)` fails in the same way even though it does not involve `ArrayOps`.

szeiger:
We could additionally make `WrappedArray` use more optimized implementations but the primary place for them should be `ArrayOps` because it is a pure value class. Because of the specialization `WrappedArray` always needs to allocate an instance (unless you implement a method in all of the specialized subclasses, which was probably the reason for the duplicate `apply` and `update` implementations in the old `ArrayOps`).

szeiger:
Something weird is going on with the `Array.apply` inference problem. The good news is that it works when we're using `Predef` and the `scala` namespace:

```
[info] Running scala.tools.nsc.MainGenericRunner -usejavacp
Welcome to Scala 2.13.0-20180220-202434-37b204e (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_102).
Type in expressions for evaluation. Or try :help.

scala> Array("") ++ Array("", null)
res0: Array[String] = Array("", "", null)

scala> Iterable("") ++ Array("", null)
res1: Iterable[String] = List("", "", null)
```

The problem only occurs when testing in the strawman namespace. My guess would be that some special case handling in the compiler is getting in the way. For example, you can see in `-Ytyper-debug` output that `arrayToWrappedArray` is always considered as an implicit conversion even when it is not in scope and not imported.

szeiger:
Ticket for the overload problem: https://github.com/scala/bug/issues/10746

The reason why it works from Predef is that we still have the separate implicits for converting primitive and reference arrays to `WrappedArray`, whereas in strawman there is only a single conversion method.

collections/src/main/scala/strawman/collection/mutable/ArrayBuilder.scala
```diff
@@ -69,12 +96,13 @@ object ArrayBuilder {
...
-  final class ofRef[T <: AnyRef : ClassTag] extends ArrayBuilder[T] {
+  final class ofRef[T <: AnyRef](implicit ct: ClassTag[T]) extends ArrayBuilder[T] {
 
-    private var elems: Array[T] = _
+    protected var elems: Array[T] = _
 
     private def mkArray(size: Int): Array[T] = {
-      val newelems = new Array[T](size)
+      // Anything short of this will make Dotty create an Array[Object]:
+      val newelems = java.lang.reflect.Array.newInstance(ct.runtimeClass, size).asInstanceOf[Array[T]]
```
szeiger: This is a strange one. I was unable to come up with a reproduction in the REPL but it fails consistently here. The `ClassTag` is the one for `Tuple2` but I end up with an `Array[Object]` when I use `new Array` or `Array.ofDim` (even passing the `ClassTag` explicitly).

collections/src/main/scala/strawman/collection/ArrayOps.scala
```diff
@@ -79,24 +536,138 @@ class ArrayOps[A](val xs: Array[A]) extends AnyVal
...
+      b(i) = ((xs(i), i))
+      i += 1
+    }
+    b
+  }
+
+  /** A copy of this array with an element appended. */
+  def appended[B >: A : ClassTag](x: B): Array[B] = {
+    val dest = new Array[B](xs.length + 1)
+    Array.copy(xs, 0, dest, 0, xs.length)
```
lrytz: @mkeskells suggests using `Arrays.copyOf(xs, xs.length+1, classTag[B])` (https://github.com/scala/scala/pull/6105#issuecomment-345505209)

szeiger: Good idea. We can also optimize all `slice`-based operations with `copyOfRange`.

lrytz: Do you think it's possible to optimize for the case where `B` is `A`, could we have an overload that avoids materializing the ClassTag?

collections/src/main/scala/strawman/collection/mutable/ArrayBuilder.scala
allanrenucci: I opened lampepfl/dotty#4037

PR #480, opened 2018-02-22
Introduce IterableOnceOps as common abstraction for Iterable & Iterator

### LINKED ISSUE #465: Enforce consistency between the available operations of IterableOnce and Iterable

szeiger:
Fixes #465

### DIFF
--- a/collections-contrib/src/main/scala/strawman/collection/immutable/MultiSet.scala
+++ b/collections-contrib/src/main/scala/strawman/collection/immutable/MultiSet.scala
@@ -10,10 +10,38 @@ import strawman.collection.decorators.ImmutableMapDecorator
   * An immutable multiset
   * @tparam A the element type of the collection
   */
-class MultiSet[A] private (elems: Map[A, Int])
+trait MultiSet[A]
   extends collection.MultiSet[A]
     with Iterable[A]
-    with collection.MultiSetOps[A, MultiSet, MultiSet[A]] {
+    with MultiSetOps[A, MultiSet, MultiSet[A]] {
+
+  override def iterableFactory: IterableFactory[MultiSet] = MultiSet
+}
+
+trait MultiSetOps[A, +CC[X] <: MultiSet[X], +C <: MultiSet[A]] extends collection.MultiSetOps[A, CC, C] {
+  /**
+    * @return an immutable multiset containing all the elements of this multiset
+    *         and one more occurrence of `elem`
+    * @param elem the element to add
+    */
+  def incl(elem: A): C
+
+  /** Alias for `incl` */
+  @`inline` final def + (elem: A): C = incl(elem)
+
+  /**
+    * @return an immutable multiset containing all the elements of this multiset
+    *         and one occurrence less of `elem`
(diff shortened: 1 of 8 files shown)


szeiger:
Looks like an actual unsoundness in `SortedMultiSet` slipped through the cracks of the compiler(s) so far but gets flagged now that `zipWithIndex` was moved to a supertrait.

collections-contrib/src/main/scala/strawman/collection/immutable/MultiSet.scala
```diff
@@ -10,10 +10,38 @@ import strawman.collection.decorators.ImmutableMapDecorator
...
   */
-class MultiSet[A] private (elems: Map[A, Int])
+trait MultiSet[A]
   extends collection.MultiSet[A]
     with Iterable[A]
-    with collection.MultiSetOps[A, MultiSet, MultiSet[A]] {
+    with MultiSetOps[A, MultiSet, MultiSet[A]] {
+
+  override def iterableFactory: IterableFactory[MultiSet] = MultiSet
+}
```
julienrf: Can you elaborate on the problem that is solved by introducing the separation `MultiSet`/`MultiSetImpl`?

julienrf: Oh maybe that’s for making `SortedMultiSet` a subtype of `MultiSet`?

collections/src/main/scala/strawman/collection/IterableOnce.scala
```diff
@@ -135,3 +138,831 @@ object IterableOnce {
   @`inline` implicit def iterableOnceExtensionMethods[A](it: IterableOnce[A]): IterableOnceExtensionMethods[A] =
     new IterableOnceExtensionMethods[A](it)
 }
+
+/** This implementation trait can be mixed into an `IterableOnce` to get the basic methods that are shared between
+  * `Iterator` and `Iterable`. The `IterableOnce` must support multiple calls to `iterator()` but may or may not
+  * return the same `Iterator` every time.
+  */
```
julienrf: You should define the documentation variables that are not inherited but used in the operations’ scaladoc.

collections-contrib/src/main/scala/strawman/collection/immutable/MultiSet.scala
szeiger: Yes, you want the usual diamond pattern with all of:

- mutable.SortedMultiSet <: collection.SortedMultiSet
- immutable.SortedMultiSet <: collection.SortedMultiSet
- mutable.MultiSet <: collection.MultiSet
- immutable.MultiSet <: collection.MultiSet

- collection.SortedMultiSet <: collection.MultiSet
- mutable.SortedMultiSet <: mutable.MultiSet
- immutable.SortedMultiSet <: immutable.MultiSet

This requires that `MultiSet` is a trait.

PR #482, opened 2018-02-24
 Override HashMap.default in Seq.occCounts instead of using Map.withDefaultValue

marcelocenerine:
Reverting a small change I made [here](https://github.com/scala/collection-strawman/pull/444/files#diff-d2679eddf2a5dee23263f82189f93c81). It was not needed for the purpose of that PR (turns out the anonymous class is only used internally) and can slightly increase the running time of `diff` and `intersect`. This is likely caused by the fact that `Map.withDefaultValue` wraps the default value in a function causing an additional dispatch every time a key is missing.

image


julienrf:
Your benchmark seems to show that performance was slightly better before the PR, no?

marcelocenerine:
sorry @julienrf ,
Red line is before #444
Blue line is after #444 (which is what this PR aims to revert)

julienrf:
Ah, thanks for the clarification!

PR #483, opened 2018-02-24
Upgrade from 2.13.0-M2 to 2.13.0-M3

szeiger:
- PagedSeq no longer exists in M3 (it was deprecated since 2.11)


szeiger:
Looks like there's no scala-js for M3 yet. The type errors in contrib are more worrying though.

julienrf:
Hey Stefan, we already have had #390 and this has been taken over by #345.

szeiger:
To be sorted out in scala/scala. We won't be able to bootstrap that with M3 anyway.

PR #486, opened 2018-02-27
Add size methods to VectorBuilder

NthPortal:
Fixes scala/bug#9904


szeiger:
Should `size` be defined in `Builder` or even in `Growable`? Do we have any cases where adding to a `Builder` is cheap but keeping track of the current size would incur an overhead?

NthPortal:
It doesn't sound like a terrible idea to define `size` on `Builder` (my gut says that `Growable` is too general to reliably track size, but I could be completely wrong there). I don't really know enough to say definitively one way or another - I was just trying to fix the open issue

PR #487, opened 2018-02-27
Make MapWrapper.Entry's hashCode conform to the contract in java.util.Map.Entry's documentation

isnotinvain:
This is a port of https://github.com/scala/scala/pull/6233 , which fixes scala/bug#10663

In the above PR, I also deleted a test (test/files/run/t5880.scala) that was testing for good spread in the (better but incorrect) hashcode implementation. I don't see that same test anywhere in this repo however.


isnotinvain:
Thanks!

PR #354, opened 2018-01-20
Optimize slice operation in ImmutableArray

ackratos:
baseline performance (0a000621122dc58cccee08f41355e3bb940f0852):
```
[info] Benchmark                              (size)  Mode  Cnt        Score         Error  Units
[info] ImmutableArrayBenchmark.access_slice        0  avgt    8       38.393 ±       0.416  ns/op
[info] ImmutableArrayBenchmark.access_slice        1  avgt    8       39.797 ±       0.606  ns/op
[info] ImmutableArrayBenchmark.access_slice        2  avgt    8       38.203 ±       0.261  ns/op
[info] ImmutableArrayBenchmark.access_slice        3  avgt    8       37.194 ±       0.369  ns/op
[info] ImmutableArrayBenchmark.access_slice        4  avgt    8       41.499 ±       0.251  ns/op
[info] ImmutableArrayBenchmark.access_slice        7  avgt    8       40.420 ±       0.485  ns/op
[info] ImmutableArrayBenchmark.access_slice        8  avgt    8       37.346 ±       0.207  ns/op
[info] ImmutableArrayBenchmark.access_slice       15  avgt    8       41.066 ±       0.387  ns/op
[info] ImmutableArrayBenchmark.access_slice       16  avgt    8       38.310 ±       0.326  ns/op
[info] ImmutableArrayBenchmark.access_slice       17  avgt    8       40.797 ±       0.313  ns/op
[info] ImmutableArrayBenchmark.access_slice       39  avgt    8       42.464 ±       0.477  ns/op
[info] ImmutableArrayBenchmark.access_slice      282  avgt    8      107.206 ±       2.077  ns/op
[info] ImmutableArrayBenchmark.access_slice     4096  avgt    8     1483.075 ±      36.826  ns/op
[info] ImmutableArrayBenchmark.access_slice   131070  avgt    8    55750.166 ±    1819.028  ns/op
[info] ImmutableArrayBenchmark.access_slice  7312102  avgt    8  6388168.153 ± 7012236.507  ns/op
```

This PR improvement:
```
[info] Benchmark                              (size)  Mode  Cnt       Score       Error  Units
[info] ImmutableArrayBenchmark.access_slice        0  avgt    8       7.399 ±     0.096  ns/op
[info] ImmutableArrayBenchmark.access_slice        1  avgt    8       6.912 ±     0.084  ns/op
[info] ImmutableArrayBenchmark.access_slice        2  avgt    8       6.996 ±     0.021  ns/op
[...]

PR #493, opened 2018-03-06
Fixes from scala/scala repository

### LINKED ISSUE #214: Backport work that has been made on scala/scala
For instance: https://github.com/scala/scala/pull/5972

Here is the history of the collections: https://github.com/scala/scala/commits/2.13.x/src/library/scala/collection
Here is the history of the JUnit tests: https://github.com/scala/scala/commits/2.13.x/test/junit/scala/collection

julienrf:
I went through the history of the collections on the scala/scala repo, for the last year, and backported all the things that were missing in the strawman.

Fixes #214.

### DIFF
--- a/collections/src/main/scala/strawman/collection/Iterable.scala
+++ b/collections/src/main/scala/strawman/collection/Iterable.scala
@@ -361,6 +361,13 @@ trait IterableOps[+A, +CC[_], +C] extends Any with IterableOnce[A] with Iterable
     */
   def splitAt(n: Int): (C, C) = (take(n), drop(n))
 
+  /** Selects first ''n'' elements.
+    *  $orderDependent
+    *  @param  n    the number of elements to take from this $coll.
+    *  @return a $coll consisting only of the first `n` elements of this $coll,
+    *          or else the whole $coll, if it has less than `n` elements.
+    *          If `n` is negative, returns an empty $coll.
+    */
   def take(n: Int): C = fromSpecificIterable(new View.Take(this, n))
 
   /** A collection containing the last `n` elements of this collection. */
@@ -381,6 +388,13 @@ trait IterableOps[+A, +CC[_], +C] extends Any with IterableOnce[A] with Iterable
 
   def span(p: A => Boolean): (C, C) = (takeWhile(p), dropWhile(p))
 
+  /** Selects all elements except first ''n'' ones.
+    *  $orderDependent
+    *  @param  n    the number of elements to drop from this $coll.
+    *  @return a $coll consisting of all elements of this $coll except the first `n` ones, or else the
+    *          empty $coll, if this $coll has less than `n` elements.
+    *          If `n` is negative, don't drop any elements.
+    */
   def drop(n: Int): C = fromSpecificIterable(new View.Drop(this, n))
 
   /** The rest of the collection without its `n` last elements. For
--- a/collections/src/main/scala/strawman/collection/Iterable.scala
+++ b/collections/src/main/scala/strawman/collection/Iterable.scala
@@ -361,13 +361,6 @@ trait IterableOps[+A, +CC[_], +C] extends Any with IterableOnce[A] with Iterable
     */
   def splitAt(n: Int): (C, C) = (take(n), drop(n))
 
-  /** Selects first ''n'' elements.
-    *  $orderDependent
-    *  @param  n    the number of elements to take from this $coll.
-    *  @return a $coll consisting only of the first `n` elements of this $coll,
-    *          or else the whole $coll, if it has less than `n` elements.
-    *          If `n` is negative, returns an empty $coll.
-    */
   def take(n: Int): C = fromSpecificIterable(new View.Take(this, n))
 
   /** A collection containing the last `n` elements of this collection. */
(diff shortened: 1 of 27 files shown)


julienrf:
I’m wondering what happened to https://github.com/scala/scala/pull/5664, it seems to have been reverted (https://github.com/scala/scala/commit/a11918d1053f3b0af32d7bd7fb5b75acb31724ff, https://github.com/scala/scala/commit/76babbb6728e5daf4fb95444273f6f8e0a4098d7 and https://github.com/scala/scala/commit/0365d58f62e5f21a2723365cb21636f382474805). Also, it seems that we have a similar PR (#354). What should we do about this stuff?

lrytz:
https://github.com/scala/scala/pull/5664 was partially reverted because it was binary incompatible. We don't need to worry about that here.

szeiger:
#354 is the intended replacement for the parts in https://github.com/scala/scala/pull/5664 that had to be reverted due to binary compatibility constraints.

collections/src/main/scala/strawman/collection/LinearSeq.scala
```diff
@@ -171,4 +171,8 @@ trait LinearSeqOps[+A, +CC[X] <: LinearSeq[X], +C <: LinearSeq[A]] extends Any w
     }
     last
   }
+
+  override def tails: Iterator[C] =
+    Iterator.iterate(coll)(_.tail.asInstanceOf[C]).takeWhile(_.nonEmpty) ++ Iterator(newSpecificBuilder().result())
```
szeiger: Is the cast necessary?

collections/src/main/scala/strawman/collection/package.scala
```diff
@@ -121,8 +121,8 @@ package object collection extends LowPriority {
 
   /** An extractor used to init/last deconstruct sequences. */
   object :+ {
-    /** Splits a sequence into init :+ last.
-      * @return Some((init, last)) if sequence is non-empty. None otherwise.
+    /** Splits a sequence into init :+ tail.
+      * @return Some((init, tail)) if sequence is non-empty. None otherwise.
```
szeiger: This is wrong. `:+` does split the sequence into `(init, last)`.

collections/src/main/scala/strawman/collection/Iterable.scala
```diff
@@ -361,6 +361,13 @@ trait IterableOps[+A, +CC[_], +C] extends Any with IterableOnce[A] with Iterable
     */
   def splitAt(n: Int): (C, C) = (take(n), drop(n))
 
+  /** Selects first ''n'' elements.
+    *  $orderDependent
+    *  @param  n    the number of elements to take from this $coll.
+    *  @return a $coll consisting only of the first `n` elements of this $coll,
+    *          or else the whole $coll, if it has less than `n` elements.
+    *          If `n` is negative, returns an empty $coll.
+    */
```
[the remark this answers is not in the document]
julienrf: I’ve put them here to override `IterableOnce`’s comments which give information about the fact that these operation consume the `Iterator`. This information is not relevant in `Iterable`.

But maybe we could organize the scaladoc differently…

collections/src/main/scala/strawman/collection/LinearSeq.scala
julienrf: Yes, the static return type of `tail` is `LinearSeq[A]`, whereas we want a `C`.

collections/src/main/scala/strawman/collection/package.scala
julienrf: Oops, indeed…

julienrf:
> #354 is the intended replacement for the parts in scala/scala#5664 that had to be reverted due to binary compatibility constraints.

OK, so that’s fine to not have it in this PR.

collections/src/main/scala/strawman/collection/Iterable.scala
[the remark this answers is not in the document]
szeiger: The iterator comsumption stuff is contained in templates (or whatever you call those scaladoc variables). They should be defined as empty in `Iterable` and with the relevant text in `Iterator`.

collections/src/main/scala/strawman/collection/LinearSeq.scala
szeiger: Hm, that's not nice. We lose the precise type on repeated applications of `tail`. `this.type.tail` results in C but `C.type` is only `LinearSeq[A]`.

julienrf: Having a more precise type would require a more complex upper bound on `C`.

szeiger: The problem lies in the definition of `LinearSeqOps`. We need to make `C` recursive:
```scala
trait LinearSeqOps[+A, +CC[X] <: LinearSeq[X], +C <: LinearSeq[A] with LinearSeqOps[A, CC, C]] extends Any with SeqOps[A, CC, C]
```
And the same for `immutable.LinearSeqOps` and `LazyListOps`. With these changes the cast is no longer needed.

julienrf: Yes, that’s what I was trying to say in my previous comment. The drawback is that it becomes harder to abstract over _some_ `LinearSeqOps`. See how it is done for `immutable.MapOps`, which has similar bounds: https://github.com/scala/collection-strawman/blob/master/collections-contrib/src/main/scala/strawman/collection/decorators/HasImmutableMapOps.scala#L12-L17

szeiger: I don't think this is a problem for `LinearSeqOps`. The only changes I had to do downstream was to repeat the bounds in `immutable.LinearSeqOps` and `LazyListOps`. `LinearSeqOps` is just an implementation detail. You'd abstract over `SeqOps`, possibly restricted to subtypes of `LinearSeq` if you care about performance characteristics.

collections/src/main/scala/strawman/collection/Iterable.scala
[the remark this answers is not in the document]
julienrf: In ab2d3f38c5f2357dc23829d236e5c38e795de797 I moved the documentation to `IterableOnceOps` but there is still an issue with the documentation of `IterableOnceOps`. I’ve opened #510 to keep track of it.

collections/src/main/scala/strawman/collection/LinearSeq.scala
julienrf: Allright, I’ve fixed that in a304aaf.

PR #494, opened 2018-03-07
Make LinkedList.from non recursive.

julienrf:
Fixes scala/collection-strawman#441

Not sure we are going to keep `LinkedList`, though… (see #490)


julienrf:
`LinkedList` has been removed.

Issue #437, opened 2018-02-08
Nil.indexOfSlice(Nil) is -1


lrytz:
was `0` in 2.12, currently `-1`

PR #496, opened 2018-03-07
Override knownSize in empty collections

julienrf:
Fixes scala/collection-strawman#437


lrytz:
If I understand correctly, those empty collections would inherit `def knownSize: Int = -1`. It seems that this should not lead to correctness bugs, i.e., `indexOfSlice` is using `knownSize` wrongly - no?

julienrf:
That’s a good point, actually `indexOfSlice`’s implementation is missing a test that makes it incorrect in the case the receiver is empty (but `knownSize` is `-1`). I’ve just added it.

collections/src/main/scala/strawman/collection/Seq.scala
```diff
@@ -366,29 +366,31 @@ trait SeqOps[+A, +CC[_], +C] extends Any
...
-    val tl = that.knownSize
-    if (l >= 0 && tl >= 0) {
-      val clippedFrom = math.max(0, from)
-      if (from > l) -1
-      else if (tl < 1) clippedFrom
-      else if (l < tl) -1
-      else SeqOps.kmpSearch(toSeq, clippedFrom, l, that, 0, tl, forward = true)
-    }
+  def indexOfSlice[B >: A](that: Seq[B], from: Int = 0): Int =
+    if (isEmpty && that.isEmpty && from == 0) 0
```
lrytz: This could be `if (that.isEmpty && from == 0) 0`

julienrf: Fixed!

Issue #433, opened 2018-02-08
unzip3 is missing


lrytz:
2.12 ArrayOps provides
```
  def unzip3[T1, T2, T3](implicit asTriple: T => (T1, T2, T3), ct1: ClassTag[T1], ct2: ClassTag[T2],
    ct3: ClassTag[T3]): (Array[T1], Array[T2], Array[T3])
```
Should we just bring it back?

marcelocenerine:
same for `Iterable`

PR #497, opened 2018-03-08
Add unzip3 to IterableOps

julienrf:
Fixes scala/collection-strawman#433


collections/src/main/scala/strawman/collection/StrictOptimizedIterableOps.scala
```diff
@@ -53,6 +53,20 @@ trait StrictOptimizedIterableOps[+A, +CC[_], +C]
     (first.result(), second.result())
   }
 
+  override def unzip3[A1, A2, A3](implicit asTriple: A => (A1, A2, A3)): (CC[A1], CC[A2], CC[A3]) = {
+    val b1 = iterableFactory.newBuilder[A1]()
+    val b2 = iterableFactory.newBuilder[A2]()
+    val b3 = iterableFactory.newBuilder[A3]()
+
+    foreach { xyz =>
+      val (x, y, z) = asTriple(xyz)
```
szeiger: We should avoid destructuring assigments in performance-critical code: https://github.com/scala/bug/issues/10767

Issue #405, opened 2018-02-03
Consider fallback to sorted set type for BitSet


Ichoran:
BitSet can only be mapped if the output is also an integer; otherwise there's an ambiguous reference between StrictOptimizedIterableOps and SortedSetOps.  It's easy to pick the former with `unsorted`.  But the latter doesn't seem to have any easy way to select.  It should either be the default, if possible, or have a method that would upcast to a generic sorted set type (if there is a method name that seems obvious enough)

ruhly:
It appears the same ambiguous reference issue exists with `flatMap`, `collect`, and `zip` as well?

szeiger:
This appears to be already resolved. The test case from #500 works in master without the explicit `toSortedSet` call.

PR #500, opened 2018-03-08
Add toSortedSet operation to BitSet

julienrf:
Fixes scala/collection-strawman#405

### DIFF
--- a/collections/src/main/scala/strawman/collection/BitSet.scala
+++ b/collections/src/main/scala/strawman/collection/BitSet.scala
@@ -22,6 +22,7 @@ trait BitSet extends SortedSet[Int] with BitSetOps[BitSet] {
   override protected[this] def fromSpecificIterable(coll: Iterable[Int]): BitSetC = bitSetFactory.fromSpecific(coll)
   override protected[this] def newSpecificBuilder(): Builder[Int, BitSetC] = bitSetFactory.newBuilder()
   override def empty: BitSetC = bitSetFactory.empty
+  def toSortedSet: SortedSet[Int] = this
 }
 
 object BitSet extends SpecificIterableFactory[Int, BitSet] {
@@ -155,6 +156,9 @@ trait BitSetOps[+C <: BitSet with BitSetOps[C]]
   }
 
   @`inline` final def ^ (other: BitSet): C = xor(other)
+
+  /** Upcasts this `BitSet` to a `SortedSet[Int]` */
+  def toSortedSet: SortedSet[Int]
 
   /**
     * Builds a new bitset by applying a function to all elements of this bitset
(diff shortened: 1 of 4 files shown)


collections/src/main/scala/strawman/collection/BitSet.scala
```diff
@@ -156,6 +157,9 @@ trait BitSetOps[+C <: BitSet with BitSetOps[C]]
 
   @`inline` final def ^ (other: BitSet): C = xor(other)
 
+  /** Upcasts this `BitSet` to a `SortedSet[Int]` */
+  def toSortedSet: SortedSet[Int]
```
julienrf: I’ve added a `toSortedSet` method to `BitSet`.

collections/src/main/scala/strawman/collection/immutable/BitSet.scala
```diff
@@ -49,6 +49,9 @@ sealed abstract class BitSet
   /** Update word at index `idx`; enlarge set if `idx` outside range of set.
     */
   protected def updateWord(idx: Int, w: Long): BitSet
+
+  override def toSortedSet: SortedSet[Int] = this
```
julienrf: I override it in `immutable.BitSet` to return an `immutable.SortedSet[Int]`

collections/src/main/scala/strawman/collection/mutable/BitSet.scala
```diff
@@ -138,6 +136,9 @@ class BitSet(protected[collection] final var elems: Array[Long])
     new BitSet(java.util.Arrays.copyOf(elems, elems.length))
 
   def toImmutable: immutable.BitSet = immutable.BitSet.fromBitMask(elems)
+
+  override def toSortedSet: SortedSet[Int] = this
```
julienrf: Same override in `mutable.BitSet`, but which returns a `mutable.SortedSet[Int]`

collections/src/main/scala/strawman/collection/mutable/BitSet.scala
```diff
@@ -85,8 +85,6 @@ class BitSet(protected[collection] final var elems: Array[Long])
 
   def get(elem: Int): Option[Int] = if (contains(elem)) Some(elem) else None
 
-  def unconstrained: collection.Set[Int] = this
```
julienrf: I removed this method that doesn’t exist in the old collection and for which we already have `unsorted`.

szeiger:
Requiring an explicit `toSortedSet` call seems wrong. I think the correct solution for #405 is to get the priorities of the overloads right so you don't need this method. Making `toSortedSet` an upcast is also inconsistent with other `to...` methods where we often refine the return types further (i.e. `BitSet.toSortedSet` could be expected to return a `BitSet`).

PR #511, opened 2018-03-12
Remove Traversable

szeiger:
It was only there to allow bootstrapping on top of 2.13.0-M3. It is no
longer necessary after https://github.com/scala/scala/pull/6364.

### DIFF
--- a/collections/src/main/scala/strawman/collection/package.scala
+++ b/collections/src/main/scala/strawman/collection/package.scala
@@ -5,6 +5,8 @@ import scala.Predef.{String, ArrowAssoc}
 
 package object collection extends LowPriority {
   @deprecated("Use Iterable instead of Traversable", "2.13.0")
+  type Traversable[+X] = Iterable[X]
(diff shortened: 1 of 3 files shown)


collections/src/main/scala/strawman/collection/package.scala
```diff
@@ -4,6 +4,8 @@ import scala.{Any, AnyVal, Array, Boolean, Char, IllegalArgumentException, Index
 import scala.Predef.{String, ArrowAssoc}
 
 package object collection extends LowPriority {
+  @deprecated("Use Iterable instead of Traversable", "2.13.0")
+  type Traversable[+X] = Iterable[X]
```
julienrf: Maybe we still need a type alias for backward compatibility?

szeiger: WDYM? This is the type alias.

julienrf: Oops, I must have been color blind when I reviewed the PR…

PR #512, opened 2018-03-12
Make toString in LazyList consistent with other collections.

julienrf:
Fixes scala/collection-strawman#376
Fixes scala/collection-strawman#420

This is a rebased and squashed version of #446 (authored by @dhanesharole).

### DIFF
--- a/collections/src/main/scala/strawman/collection/immutable/LazyList.scala
+++ b/collections/src/main/scala/strawman/collection/immutable/LazyList.scala
@@ -4,10 +4,11 @@ package immutable
 
 import strawman.collection.mutable.{ArrayBuffer, Builder}
 
-import scala.{Any, AnyRef, Boolean, Int, NoSuchElementException, None, Nothing, Option, PartialFunction, Some, StringContext, Unit, UnsupportedOperationException, deprecated, noinline}
+import scala.{Any, AnyRef, Boolean, Int, NoSuchElementException, None, Nothing, Option, PartialFunction, Some, StringBuilder, StringContext, Unit, UnsupportedOperationException, deprecated, noinline}
 import scala.Predef.String
 import scala.annotation.tailrec
 import scala.annotation.unchecked.uncheckedVariance
+import scala.collection.mutable.StringBuilder
 
 /**  The class `LazyList` implements lazy lists where elements
   *  are only evaluated when they are needed. Here is an example:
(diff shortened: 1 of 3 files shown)


collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -220,6 +221,119 @@ sealed abstract class LazyList[+A] extends LinearSeq[A] with LazyListOps[A, Lazy
...
+    *         Examples:
+    *
+    *           - `"LazyList(_, ?)"`, a non-empty lazy list, whose head has not been
+    *             evaluated ;
+    *           - `"LazyList(_, 1, _, ?)"`, a lazy list with at least three elements,
+    *             the second one has been evaluated ;
+    *           - `"LazyList(1, 2, 3, ...)"`, an infinite lazy list that contains
+    *             a cycle at the fourth element.
+    */
+  override def toString: String = {
```
julienrf: I finally changed my mind and agreed with @SethTisue [suggestion](https://github.com/scala/collection-strawman/issues/376#issuecomment-364211252) of using different characters to represent undefined elements and undefined tail.

collections/src/main/scala/strawman/collection/immutable/LazyList.scala
```diff
@@ -575,6 +691,100 @@ sealed abstract class Stream[+A] extends LinearSeq[A] with LazyListOps[A, Stream
     if (this.isEmpty) z
     else tail.foldLeft(op(z, head))(op)
   }
+
+
+  protected def tailDefined: Boolean
+
+  override def toString: String = {
```
szeiger: Can this implementation be shared between LazyList and Stream?

julienrf: I thought this wasn’t useful since `Stream` is deprecated anyway.

julienrf: OK, I’ve factored the implementation out in `LazyListOps`.

szeiger:
Oh, I just wondered why my commit showed up in this branch but it doesn't. We both did the same change (https://github.com/scala/collection-strawman/commit/ec281081cb0cd8753ba8a6b1efa1cb29fc29a5b3)

julienrf:
Oops, I pushed `--force`

szeiger:
We could even consider making this the default implementation of `mkString` in `IterableOps` and overriding it with the current default implementation in `StrictOptimizedIterableOps`. This would be more consistent with other default implementations where the lazy version is in `IterableOps`

julienrf:
We could maybe do it in `SeqOps` (because it relies on `head` and `tail`). But I think that’s out of the scope of this PR.

szeiger:
Right, you'd need at least a `LinearSeq` for proper head/tail destructuring. Iterators couldn't express the `headDefined`/`tailDefined` semantics.

Issue #506, opened 2018-03-11
Queue/ArrayDeque checks bounds too aggressively


Ichoran:
`ArrayDeque` throws exceptions in a variety of cases when empty when there is a very sensible behavior instead.  For instance,

```
scala> val xs = strawman.collection.mutable.Queue.empty[Int]
xs: strawman.collection.mutable.Queue[Int] = ArrayDeque()

scala> xs.toArray
java.lang.IndexOutOfBoundsException: 0
  at strawman.collection.mutable.ArrayDeque.requireBounds(ArrayDeque.scala:487)
  at strawman.collection.mutable.ArrayDeque.copySliceToArray(ArrayDeque.scala:447)
  at strawman.collection.mutable.ArrayDeque.toArray(ArrayDeque.scala:434)
  ... 40 elided

scala> xs.insert(0, 0)
java.lang.IndexOutOfBoundsException: 0
  at strawman.collection.mutable.ArrayDeque.requireBounds(ArrayDeque.scala:487)
  at strawman.collection.mutable.ArrayDeque.insert(ArrayDeque.scala:132)
  ... 40 elided
```

There are also problems with `copyToArray` and `patchInPlace`.  These issues make it impossible to test `ArrayDeque` (and thus `mutable.Queue`) properly with collections-laws, so there are likely other undiscovered bugs as well.

PR #513, opened 2018-03-12
Fix bounds checks, etc. in ArrayDeque

szeiger:
This also affects subclasses like mutable.Queue.

- Remove unnecessary `from` paramter in `requireBounds`.
- Allow `copyToSlice` directly beyond the end of a target array for a
  no-op. This is required to make it work for empty arrays.
- Allow `insert` and `insertAll` directly beyond the end of the
  `ArrayDeque` for appending to the end.
- Remove explicit `className` in `ArrayDeque` to have it computed
  correctly for subclasses like `Queue`.

Fixes https://github.com/scala/collection-strawman/issues/506

### DIFF
--- a/collections/src/main/scala/strawman/collection/mutable/ArrayDeque.scala
+++ b/collections/src/main/scala/strawman/collection/mutable/ArrayDeque.scala
@@ -129,7 +129,7 @@ class ArrayDeque[A] protected (
   }
 
   def insert(idx: Int, elem: A): Unit = {
-    requireBounds(idx)
+    requireBounds(idx, length+1)
     val n = length
     if (idx == 0) {
       prepend(elem)
@@ -164,7 +164,7 @@ class ArrayDeque[A] protected (
   }
 
   def insertAll(idx: Int, elems: IterableOnce[A]): Unit = {
-    requireBounds(idx)
+    requireBounds(idx, length+1)
     val n = length
     if (idx == 0) {
       prependAll(elems)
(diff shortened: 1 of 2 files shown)


collections/src/main/scala/strawman/collection/mutable/ArrayDeque.scala
```diff
@@ -428,8 +428,6 @@ class ArrayDeque[A] protected (
   override def copyToArray[B >: A](dest: Array[B], destStart: Int, len: Int): dest.type =
     copySliceToArray(srcStart = 0, dest = dest, destStart = destStart, maxItems = len)
 
-  override def className = "ArrayDeque"
```
julienrf: Why remove that?

szeiger: Because it's not needed. If you define it anyway all subclasses have to do the same (which they don't). Queue.toString used to print "ArrayDeque".

julienrf: I think that’s fine to define it in all subclasses.

Also, the scaladoc says:

> It is recommended to overwrite this method even if the default implementation returns the correct name, to avoid the implementation using reflection.

Which I agree with.

szeiger: The old collections library doesn't have that notice. The only use of this method is for debug output, it's never in a hot code path. The default implementation is perfectly fine for that. IMHO the recommendation should be not to override it unless necessary.

julienrf: OK then.

PR #514, opened 2018-03-13
More fixes from latest rebasing of Scala 2.13.x integration build

szeiger:
- Restore `IterableFactory.toBuildFrom` which was erroneously removed
  in https://github.com/scala/collection-strawman/pull/495
- Fix `ArrayOps.ReverseIterator`
- More explicit syntax in `AnyRefMap.collect` implementation to make it
  compile with and without Predef
- Add a deprecated `front` method to `mutable.Queue`

### DIFF
--- a/collections/src/main/scala/strawman/collection/Factory.scala
+++ b/collections/src/main/scala/strawman/collection/Factory.scala
@@ -245,6 +245,12 @@ object IterableFactory {
       def newBuilder(): Builder[A, CC[A]] = factory.newBuilder[A]()
     }
 
+  implicit def toBuildFrom[A, CC[_]](factory: IterableFactory[CC]): BuildFrom[Any, A, CC[A]] =
+    new BuildFrom[Any, A, CC[A]] {
+      def fromSpecificIterable(from: Any)(it: Iterable[A]) = factory.from(it)
+      def newBuilder(from: Any) = factory.newBuilder()
+    }
+
   class Delegate[CC[_]](delegate: IterableFactory[CC]) extends IterableFactory[CC] {
     def empty[A]: CC[A] = delegate.empty
     def from[E](it: IterableOnce[E]): CC[E] = delegate.from(it)
--- a/collections/src/main/scala/strawman/collection/mutable/Queue.scala
+++ b/collections/src/main/scala/strawman/collection/mutable/Queue.scala
@@ -97,14 +97,13 @@ class Queue[A] protected (array: Array[AnyRef], start: Int, end: Int)
     *  @param f   the predicate used for choosing elements
     *  @return
     */
-  def dequeueAll(f: A => Boolean): strawman.collection.Seq[A] = removeHeadWhile(f)
+  def dequeueAll(f: A => Boolean): strawman.collection.immutable.Seq[A] = removeHeadWhile(f)
 
   /** Returns the first element in the queue, or throws an error if there
     *  is no element contained in the queue.
     *
     *  @return the first element.
     */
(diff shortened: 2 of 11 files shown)


szeiger:
This should be the last changes we need before moving to scala/scala, but let's wait with merging until I've rebased Lukas' branch (with the partest tests) on top of mine.

collections/src/main/scala/strawman/collection/Factory.scala
```diff
@@ -245,6 +245,12 @@ object IterableFactory {
       def newBuilder(): Builder[A, CC[A]] = factory.newBuilder[A]()
     }
 
+  implicit def toBuildFrom[A, CC[_]](factory: IterableFactory[CC]): BuildFrom[Any, A, CC[A]] =
+    new BuildFrom[Any, A, CC[A]] {
+      def fromSpecificIterable(from: Any)(it: Iterable[A]) = factory.from(it)
+      def newBuilder(from: Any) = factory.newBuilder()
+    }
```
julienrf: Perfect, I’ve just added some tests for that, to not repeat the mistake.

collections/src/main/scala/strawman/collection/mutable/Queue.scala
```diff
@@ -97,14 +97,13 @@ class Queue[A] protected (array: Array[AnyRef], start: Int, end: Int)
...
-  def dequeueAll(f: A => Boolean): strawman.collection.Seq[A] = removeHeadWhile(f)
+  def dequeueAll(f: A => Boolean): strawman.collection.immutable.Seq[A] = removeHeadWhile(f)
 
   /** Returns the first element in the queue, or throws an error if there
     *  is no element contained in the queue.
     *
     *  @return the first element.
     */
-  @deprecated("Use Queue.head instead of Queue.front", "2.13.0")
   @`inline` final def front: A = head
```
julienrf: I don’t think it’s a good thing to have several method names to perform the same operations. I’m in favor of keeping the deprecation warning.

collections/src/main/scala/strawman/collection/mutable/Stack.scala
```diff
@@ -38,46 +39,63 @@ class Stack[A] protected (array: Array[AnyRef], start: Int, end: Int)
...
     *
     *  @param elems the traversable object.
     *  @return the stack with the new elements on top.
     */
-  def pushAll(elems: strawman.collection.IterableOnce[A]): this.type = this ++= elems
+  def pushAll(elems: strawman.collection.IterableOnce[A]): this.type =
+    prependAll(elems match {
+      case it: strawman.collection.Seq[A] => it.view.reverse
+      case it => IndexedSeq.from(it).view.reverse
+    })
```
julienrf: We see here that it’s less efficient to choose this order (pushed elements are prepended rather than appended). Is it specified somewhere that it should be done that way?

szeiger: It's not less efficient in general but the implementation is not ideal. Note that `prependAll` converts to an `IndexedSeq` anyway if the size is not known, so the amount of copying shouldn't change, it's only an extra layer of indirection through the view.

collections/src/main/scala/strawman/collection/mutable/Queue.scala
szeiger: Adding `Stack.top` convinced me that we should also have `Queue.top`. It's only a matter of convention how the order of the `Seq` is related to the `Queue` or `Stack` semantics. `front` and `top` make it clear what you want in terms of queue or stack.

PR #517, opened 2018-03-19
Dotty #4013

allanrenucci:
This version of Dotty fixes several unsoundness issues and remove the unsound protected[this] escape hatch for variance checking. See lampepfl/dotty#4013

@julienrf You might want to play with this branch and make it compile with Dotty

### DIFF
--- a/collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
+++ b/collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
@@ -125,8 +125,8 @@ sealed abstract class ImmutableArray[+A]
   * @define coll immutable array
   * @define Coll `ImmutableArray`
   */
-object ImmutableArray extends StrictOptimizedClassTagSeqFactory[ImmutableArray] { self =>
-  val untagged: SeqFactory[ImmutableArray] = new ClassTagSeqFactory.AnySeqDelegate(self)
+object ImmutableArray extends StrictOptimizedClassTagSeqFactory[ImmutableArray] {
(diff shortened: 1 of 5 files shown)


collections/src/main/scala/strawman/collection/immutable/ImmutableArray.scala
```diff
@@ -125,8 +125,8 @@ sealed abstract class ImmutableArray[+A]
   * @define coll immutable array
   * @define Coll `ImmutableArray`
   */
-object ImmutableArray extends StrictOptimizedClassTagSeqFactory[ImmutableArray] { self =>
-  val untagged: SeqFactory[ImmutableArray] = new ClassTagSeqFactory.AnySeqDelegate(self)
+object ImmutableArray extends StrictOptimizedClassTagSeqFactory[ImmutableArray] {
+  val untagged: SeqFactory[ImmutableArray] = new ClassTagSeqFactory.AnySeqDelegate(this)
```
szeiger: What's the reason for these changes? Just removing some unnecessary code, or is there a change in Dotty that necessitates this?

allanrenucci: We recently disallowed named self in lampepfl/dotty#4118. We might reconsider it though

odersky: But the fix refers to the wrong this! Btw we will revert #4118.

odersky:
Superseded by #519.

PR #518, opened 2018-03-20
Don’t expose Set{1..4} names via className / toString

szeiger:
They all print as “Set” now, like in the old collection library.


julienrf:
You need to import `scala.Predef.String` (or `java.lang.String`)

szeiger:
Happens every time when the change is so trivial that I don't bother building it standalone (but only as part of the standard library)...

PR #528, opened 2018-03-28
Don’t replace Stream with LazyList

julienrf:
Because their behaviour is not the same (`LazyList` has a lazy head) I think we should not rewrite `Stream` usage to `LazyList`. Note that `Stream` is deprecated in 2.13, though, that’s why I’m not totally sure what should be the right decision.

(this PR depends on #527)

### DIFF
--- a/scalafix/2.13/input/src/main/scala/fix/Collectionstrawman_v0_Stream.scala
+++ b/scalafix/2.13/input/src/main/scala/fix/Collectionstrawman_v0_Stream.scala
@@ -4,7 +4,8 @@ rule = "scala:fix.Collectionstrawman_v0"
 package fix
 
 object Collectionstrawman_v0_Stream {
-  Stream(1, 2, 3)
+  val s = Stream(1, 2, 3)
+  s.append(List(4, 5, 6))
   1 #:: 2 #:: 3 #:: Stream.Empty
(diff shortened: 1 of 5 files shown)


linqing:
scala> def h() = { println("compute head"); 1}
h: ()Int

scala> LazyList.iterate(h())(_ + 1).take(3).toList
compute head
compute head
compute head
res5: List[Int] = List(1, 2, 3)
It's bug or not?

szeiger:
Can we modularize these rules? It would be nice if you could perform this migration separate from other ones. In most cases you'll want to use LazyList and do not need any additional changes to make it work.

julienrf:
What we could do:

- provide each rule separately, so that users can have fine grained control on which ones to apply,
- provide a “pack” of rules (ie. a single rule that aggregates several rules) that migrate a 2.12 code base to a 2.13 code base,
- provide a “pack” of rules that migrates a 2.12 code base to a cross-compatible 2.12/2.13 code base (using the collections-compat library).

julienrf:
I’m closing this PR. We should still provide this rule and let users not apply it if they don’t want to. I’ve created #559 to keep track of that.

SethTisue:
I hadn't commented because I wasn't sure what I thought about this. But I'm now comfortable with this outcome. 👍

PR #558, opened 2018-04-17
Update the collections-contrib module to use 2.13.0-M4-pre-20d3c21.

julienrf:
Also, update README and CONTRIBUTING files. This repository now only
hosts the code of the collections-contrib module and the scalafix
rewrite rules.

I didn’t remove the `collections/` directory, nor the `benchmarks/` and `test/` directories, even though I’ve removed the corresponding sbt projects.

Thinking more about this I think we should:

- move `collections-contrib` to its own repository,
- move the scalafix rewrite rules to the `scala-collection-compat` repository,
- just update the README of this repository and lock it.


lrytz:
We should still talk about "collections-contrib". Do we release it, what's the release schedue? Stability guarantees? Should we include parts into the scala/scala repo, e.g., the classes that help defining extension methods?

julienrf:
I agree that we could move the extension methods framework to the scala/scala repository.

For `collections-contrib` itself, would it be possible for Lightbend to maintain it as separate module, like `scala-parallel-collections`?

README.md
```diff
@@ -61,9 +52,9 @@ The `collections-contrib` artifact provides additional operations on the collect
 ### Migrating from the standard collections to the strawman
 
 There is an [entry in the FAQ](https://github.com/scala/collection-strawman/wiki/FAQ#what-are-the-breaking-changes)
-that aims to list all breaking changes.
+that aims to list all the breaking changes.
 
-A tool is being developed to automatically migrate code that uses the standard
+A tool is currently developed to automatically migrate code that uses the standard
```
SethTisue: "is currently developed" is not clear English, I'm afraid — it's not a phrase a native speaker would use, and it leaves the reader uncertain of the intended meaning — does the tool exist or not?

if it's not ready yet, "is being developed" was right (and something like "is under development" would be more natural). if it is ready now, suggest "is now available"

julienrf: Thanks, I’m putting “is being developed” back!

SethTisue: (okay, note that "is currently being developed" would also be idiomatic and fine, if you like the additional emphasis on "currently")

PR #561, opened 2018-04-18
Be more explicit about unsound code in collections-contrib.

julienrf:
Subset of scala/collection-strawman#521 containing only changes applied
to the collections-contrib module. The other changes have already
been ported to scala/scala#6508.


julienrf:
Superseded by https://github.com/scala/scala-collection-contrib/pull/1

Issue #560, opened 2018-04-18
Reorganize repositories


julienrf:
I think we should:

- move the `collections-contrib` module to its own repository (maybe `scala-collection-contrib` or `scala-collection-extra`),
- move the scalafix rewrite rules to the `scala-collection-compat` repository,
- lock this (`collection-strawman`) repository.

PR #564, opened 2018-04-19
Remove code that has been moved in other repositories

julienrf:
After https://github.com/scala/scala-collection-compat/pull/15 is merged we can merge this one. The https://github.com/scala/scala-collection-contrib repo contains the contrib module.

Fix scala/collection-strawman#560


lrytz:
It might still be useful to be able to look up things in the codebase here. How about creating a branch that points to the current state, and add a link to it to the readme?

